PR #18, opened 2015-02-11
Update wifi.md

shaunmulligan:


milosh012:
@shaunmulligan just a note... when you are changing MD files (or adding new), you need to run `node index-builder.js`. That will update search indexer (you will have modified `lunr_index.json` file, that also needs to be committed), so we have up to date search results.

PR #28, opened 2015-03-21
Restructure

craig-mulligan:
* organised pages in folders
* split pages where seemed logical eg projects into snippets, seed-projects ect.
* made side bar menu expandable
* changed search index-builder.js to index new folder system
* left aligned main menu
* added introduction page


lorenzo-stoakes:
Design-wise looks good, we defo need to update the content to reflect the .img file and various grammar fixes etc. but I think this is outside the scope of this PR. I've left some comments on the code-side. 

I've run this locally and everything seems to be working for me.

Do you update the actual content anywhere, other than updating links? It's kinda hard to check with the big changes :) if not, then no need to look more in depth at that, if so then can take a look there.

It also seems like you have conflicts as github is moaning about that - I can help you fix them if you like, let me know!

I've made some comments on some stuff you need to fix, we'll take another look once you've got that done and let me know if you want help with fixing the conflicts :)

package.json
```diff
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/resin-io/docs",
   "dependencies": {
-    "express": "~4.11.0"
+    "express": "~4.11.0",
+    "async": "~2.2.0"
```
emirotin: how did it work for you? async is at 0.9.0 now...

emirotin:
I've resolved conflicts and addressed issues here https://github.com/resin-io/docs/compare/restructure-resolve

From one side we want to merge this asap and fix it further because RPi1 docs are outdated (they refer ZIP file while we do IMG now).

From the other we should be careful - we have multiple links to docs from our UI and Craig has changed almost every of them :)

PR #29, opened 2015-03-25
Details on SD card and reminder to share connection

pcarranzav:
In Getting Started with RPI2.

The Internet connection sharing thing could be expanded a bit more, but so far I couldn't find a single webpage  / wikiHow with a tutorial for all OS's.


shaunmulligan:
Yeah, its difficult...There are also a number of issues with connection sharing with OSX especially with the new retina macbooks when you need a thunderbolt ethernet adaptor, I have had trouble making it work reliably :cry:

PR #32, opened 2015-04-13
Feature/edison support 1

shaunmulligan:


emirotin:
The Edison type has (PREVIEW) in the screen

PR #43, opened 2015-04-28
re-add /data section 

craig-mulligan:
The CLI updates are automatically pulled from the CLI repo and lunr is re-indexed. I've just changed runtime.md


lorenzo-stoakes:
In that case, LGTM :)

PR #49, opened 2015-05-12
Add warning about GPIO4/5 reservations.

lorenzo-stoakes:
Currently GPIO4/GPIO5 are reserved, the former due to the one-wire protocol, the latter is uncertain (perhaps related?) 

For internal resin people:- https://www.flowdock.com/app/rulemotion/user_happiness/messages/1088665 is where the fix that removes the need for this is being discussed :)


lorenzo-stoakes:
@emirotin hmm I think I also messed this up by not making it clear this was for gh-pages, it's now been merged into the update-cli-docs branch :P I was comparing the 2 in order that the first that got merged would land first and the 2nd would land vs. gh-pages but I didn't want the diff to include stuff that update-cli-docs is doing.

I'll delete both branches,and resubmit update-cli first vs. gh-pages then add-reserved-gpio afterwards also vs. gh-pages.

emirotin:
I abominated everything, didn't even check the target there :(

PR #50, opened 2015-05-13
Update CLI documents and index.

lorenzo-stoakes:
This change is a result of running `bash index.sh`, which has added the await CLI functionality documentation.


emirotin:
lgtm

PR #51, opened 2015-05-13
Update GPIO page to reflect reserved GPIO4/GPIO5.

lorenzo-stoakes:
These are reserved due to the one-wire protocol (GPIO4 is explained by this,
GPIO5 is more of a mystery, but may be related to only being present on the B+
or rpi2.)

Additionally this commit references the rpi2 in the B+ pinout diagram.


emirotin:
lgtm

PR #55, opened 2015-06-05
CLI now has --yes option.

lorenzo-stoakes:
Ran `index.sh` to import latest CLI changes which turn out to include the --yes option :)


shaunmulligan:
LGTM

PR #64, opened 2015-07-10
Feature/dockerfile templates

shaunmulligan:


shaunmulligan:
Don't merge until device-type and base images naming align

craig-mulligan:
@shaun Perhaps could do with a link to your multi-device app at the end?

shaunmulligan:
@craig-mulligan this has been updated to reflect the RESIN_MACHINE_NAME variable rather than the device type

shaunmulligan:
i will try add my multi-arch ADC app later, it needs one change for it to work i think

PR #68, opened 2015-07-23
Fixed signup link

mikeboehm:


emirotin:
Thanks for the fix! But the link should actually be fixed at the very bottom of the file https://github.com/resin-io/docs/pull/68/files#diff-1f22d41f1c5b87721ea1e9250f97315cR223

We'll probably have to close your PR and reapply the fix in a new branch. @shaunmulligan?

shaunmulligan:
Yup, I agree. We need to maintain the links at the bottom of the .md files. I will patch it to today. Thanks again for the contribution @mikeboehm :+1:

mikeboehm:
:+1:

PR #71, opened 2015-08-25
fix git endpoints

craig-mulligan:
The format of git endpoint on production has changed? Updated the getting started guides to reflect this.


petrosagg:
SGMT

PR #74, opened 2015-09-09
Docs for new boards

nghiant2710:


nghiant2710:
@shaunmulligan, Please merge this when we have images for new boards on docker registry

PR #91, opened 2015-10-09
Document the need of Edison drivers on Windows

jviotti:
If drivers are not installed, I get:

    dfu-util: Device has DFU interface, but has no DFU functional descriptor


shaunmulligan:
LGTM, merge away :+1:

PR #92, opened 2015-10-13
Docs for Golang images

nghiant2710:


shaunmulligan:
@nghiant2710 I think we need to add a note in here about the ARM images and the fact that you will need to push to the native ARM builders.

this can now be done with: `git push resin master:master-arm`

PR #100, opened 2015-10-16
add golang image example to seed projects

nghiant2710:


pages/examples/seed-projects.md
```diff
@@ -92,6 +92,11 @@ in the `Dockerfile`.
 __Note:__ This project will not build on any of the ARM architecture board. Currently,
 this project will only work on the Intel Edison.
 
+[Resin Golang Image Example][golang-image-link]
+
+This project contains examples of how to use Resin Golang images on devices supported by Resin.io.
+This project will work on any devices, notice that it requires native ARM builders when building for ARM architecture devices.
+
```
shaunmulligan: you perhaps should explain how to use the arm servers..

nghiant2710: @shaunmulligan, details of how to use the arm servers is described my project README, so I don't think we should bring it here.

shaunmulligan: I don't think it hurts to have the info in both places, but its up to you.

nghiant2710: @shaunmulligan Updated! I'm happy to bring it here, if it can help our users  :)

PR #115, opened 2015-12-02
spelling fix

lekkas:


lekkas:
Hm maybe I should group these into a single commit to reduce the noise in history. Closing this PR.

PR #117, opened 2015-12-03
Jerky searchbar fix

lekkas:
Bug:
bug

After fix:
fix


static/scripts/main.js
```diff
@@ -6,23 +6,39 @@ window.onload = function () {
...
+var UnpinOffset = 400;
+var PrevScrollTop = 0;
 
 $(window).scroll(function() {
-  var NextScroll = $(this).scrollTop();
-  if (NextScroll < CurrentScroll && NextScroll > 67){
-     //upscroll
-     $('[data-md-sticky-header]').addClass('sticky');
+  function handleScrollUp() {
+    if (ScrollTop > SearchbarTop && ScrollTop <= UnpinOffset)
```
jviotti: Can you add curly braces around `if`/`else` bodies for safety and consistency?

static/scripts/main.js
```diff
@@ -6,23 +6,39 @@ window.onload = function () {
...
       hash.scrollIntoView()
     }
+    $('[data-md-sticky-header]').headroom({
+      offset: UnpinOffset,
+      tolerance: 0
+    });
 }
 
-var CurrentScroll = 0;
+var SearchbarTop = 61;
```
jviotti: What does this magic number stands for? The height of the header? Can it be computed dynamically?

static/scripts/main.js
```diff
@@ -6,23 +6,39 @@ window.onload = function () {
...
+var PrevScrollTop = 0;
 
 $(window).scroll(function() {
-  var NextScroll = $(this).scrollTop();
-  if (NextScroll < CurrentScroll && NextScroll > 67){
-     //upscroll
-     $('[data-md-sticky-header]').addClass('sticky');
+  function handleScrollUp() {
+    if (ScrollTop > SearchbarTop && ScrollTop <= UnpinOffset)
+      $('[data-md-sticky-header]').addClass('sticky');
```
jviotti: This element is being queried on each scroll event. Can we save a reference outside the scroll event handler and reuse here?

jviotti:
Looks really nice!

static/scripts/main.js
```diff
@@ -6,23 +6,39 @@ window.onload = function () {
...
     }
+    $('[data-md-sticky-header]').headroom({
+      offset: UnpinOffset,
+      tolerance: 0
+    });
 }
 
-var CurrentScroll = 0;
+var SearchbarTop = 61;
+var UnpinOffset = 400;
```
jviotti: This variable is used when applying `headroom` on the header at line 9, however the variable is declared at line 15, which could cause some confusion. Can we move this variable to a place before it's used?

static/scripts/main.js
emirotin: yeah, even single-statement blocks should be in braces

this code (I mean the entire source, not this PR specifically) actually needs linting, for example, variables should start from lower-case letter

emirotin:
Cool improvement, just code style nits

static/scripts/main.js
lekkas: It is the topmost vertical position of the searchbar. I agree we shouldn't be using a magic value for that, we can compute it dynamically. I'll look into that, I think $(".search-wrapper").offset().top will do.

static/scripts/main.js
lekkas: Fair point, will add curly braces in the next commit.

static/scripts/main.js
lekkas: :+1:

jviotti:
LGTM

PR #124, opened 2015-12-09
Add/fa qs

shaunmulligan:
Adds a couple of FAQs and Troubleshooting bits all stolen from lolo's scratch pad


shaunmulligan:
@lorenzo-stoakes just need your okay before i merge this badboy

PR #145, opened 2016-01-14
Document the new update strategies feature

pcarranzav:
(Should only be merged after the env var whitelisting in the API is deployed to production)


pcarranzav:
:tada:

PR #147, opened 2016-01-15
Adding README and start script

fedealconada:
- Adding README and start script

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,40 @@
-# resin.io docs
+Resin.io Documentation
+======================
 
-This repo contains our public-facing documentation.
+[![Build status](https://ci.appveyor.com/api/projects/status/qbsivehgnq0vyrrb?svg=true)](https://ci.appveyor.com/project/fedealconada/resin-docs)
+
+Join our online chat at [![Gitter chat](https://badges.gitter.im/resin-io/chat.png)](https://gitter.im/resin-io/chat)
+
+Documentation for the [Resin.io](https://resin.io/) platform.
+
+Role
+----
+
+The intention of this repo is to provide our public-facing documentation.
+
+
+Running locally
+---------------
+
+To run the docs web application locally you should do the following:
+
+```sh
+git clone https://github.com/resin-io/docs
+cd docs
+npm install && bower install
+npm start
+```
+And then open [http://localhost:3000](http://localhost:3000).
+
+Contribute
+----------
+
+- Issue Tracker: [github.com/resin-io/docs/issues](https://github.com/resin-io/docs/issues)
+- Source Code: [github.com/resin-io/docs](https://github.com/resin-io/docs)
+
+If you think something is not documented, or can be improved, don't hesitate in sending a PR!
+
+License
+-------
+
+The project is licensed under the Apache 2.0 license.
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "This repo contains our public-facing documentation.",
   "main": "app.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,3 +1,45 @@
-# resin.io docs
+Resin Documentation
+---------
 
-This repo contains our public-facing documentation.
+[![npm version](https://badge.fury.io/js/docs.svg)](http://badge.fury.io/js/docs)
```
jviotti: I don't think these badges make any sense in this repo.

README.md
```diff
@@ -1,3 +1,45 @@
...
+[![dependencies](https://david-dm.org/resin-io/docs.png)](https://david-dm.org/resin-io/docs.png)
+[![Build Status](https://travis-ci.org/resin-io/docs.svg?branch=master)](https://travis-ci.org/resin-io/docs)
+[![Build status](https://ci.appveyor.com/api/projects/status/qbsivehgnq0vyrrb?svg=true)](https://ci.appveyor.com/project/fedealconada/resin-docs)
+
+Join our online chat at [![Gitter chat](https://badges.gitter.im/resin-io/chat.png)](https://gitter.im/resin-io/chat)
+
+The official JavaScript [Resin.io](https://resin.io/) Documentation.
```
jviotti: Why JavaScript? Probably better to stick with the repo tagline: "Documentation for the resin.io platform.".

README.md
```diff
@@ -1,3 +1,45 @@
...
+Install the Resin Documentation by running:
+
+```sh
+git clone https://github.com/resin-io/docs
+cd docs
+npm install && bower install
+npm start
+```
+
+Support
```
jviotti: I don't think a "Support" section is worth here since this is not an app people will be cloning and using themselves. I'd probably extend the "Contribute" section to write up something like: "If you think something is not documented, or can be improved, don't hesitate in sending a PR!".

README.md
```diff
@@ -1,3 +1,45 @@
-# resin.io docs
+Resin Documentation
+---------
```
jviotti: The dashes should extend to the same width as the title. Also, it's a good practice to use equal signs instead of dashes for the main title:

```
Resin Documentation
===================
```

jviotti:
@shaunmulligan and @craig-mulligan might want to take a look at this as well.

README.md
```diff
@@ -1,3 +1,45 @@
-# resin.io docs
+Resin Documentation
```
jviotti: I think we should be referring as "Resin.io" instead of just "Resin", since that's how the brand is registered.

README.md
jviotti: Ugh, there are actually some projects that still use dashes. I'll make a note to update myself!

README.md
```diff
@@ -1,3 +1,45 @@
...
+Role
+----
+
+The intention of this repo is to provide our public-facing documentation.
+
+
+Installation
+------------
+
+Install the Resin Documentation by running:
```
jviotti: Probably better to rephrase as "To run the docs web application locally do..." or something like that. Maybe even worth to rename the "Installation" header to "Running locally"?

README.md
```diff
@@ -1,3 +1,42 @@
-# resin.io docs
+Resin.io Documentation
+======================
 
-This repo contains our public-facing documentation.
+[![dependencies](https://david-dm.org/resin-io/docs.png)](https://david-dm.org/resin-io/docs.png)
+[![Build Status](https://travis-ci.org/resin-io/docs.svg?branch=master)](https://travis-ci.org/resin-io/docs)
```
jviotti: These two "build status" badges should be removed, given that this repository doesn't run any tests in the CI servers.

shaunmulligan:
LGTM

PR #161, opened 2016-03-03
Update CLI, SDKs and re-index so new stuff is searchable

shaunmulligan:
This PR pulls in the new changes in the SDK and CLI docs.
It then re-indexes all the MD files in the repo, so that they are searchable on the docs.


lorenzo-stoakes:
This is bringing in a bunch of other changes, SDK docs and CLI troubleshooting, we should separate out or just rename the PR to be accurate

shaunmulligan:
True, I will just update the description... I forgot that all the cli and sdk stuff gets pulled in too.

lorenzo-stoakes:
LGTM

Issue #166, opened 2016-03-14
'Montserrat' font is not loaded


emirotin:
The recent change breaks the headers because the Montserrat webfont is not loaded.

taahirisaacs:
@emirotin Will amend thanks for the catch

emirotin:
I can fix it, no worries

taahirisaacs:
@emirotin thanks here's the link for reference:

<link href='https://fonts.googleapis.com/css?family=Montserrat:400,700' rel='stylesheet' type='text/css'>

PR #167, opened 2016-03-14
fix Montserrat font #166 + update CLI docs + update index + update deps

emirotin:
I've added the Montserrat font and also reduced the download size by removing font-weight 700 for Open Sans - see index.html. 
I made small changes to header styles (see main.css changes).

h1

h2 and h3

h4, h5

h5, h6

The rest are dependency updates + CLI docs pulled + index rebuilt.


taahirisaacs:
Looks good :+1: @emirotin

PR #173, opened 2016-03-16
Updated screenshots

taahirisaacs:
@craig-mulligan @shaunmulligan Updated doc images. 

Some images are updated, but img paths/file structure need to be corrected. Will amend in next commit.


craig-mulligan:
I see a couple files have the (1) at the end of the name eg: `img/OdroidC1/application_empty_OdroidC1 (1).png` I searched for that url referenced it the repo, but it didn't find anything. So I think you just have a few duplicates.

craig-mulligan:
Have you run the app locally to see that all images are loaded?

taahirisaacs:
@craig-mulligan thanks just picked that up.

 I've checked all images locally. Those odroid images are some of the images that aren't being pointed to currently.

PR #175, opened 2016-03-17
Update supervisor docs

shaunmulligan:
Update the supervisor-api docs...still need to figure out a good way for this to be automatic, like we have for the SDK and CLI, but this will require some restructuring, so for now have just copied over the new documentation


craig-mulligan:
@shaunmulligan I'm closing this as it now auto pulls from super repo when runnning index.sh

PR #179, opened 2016-03-22
Seo indexing

craig-mulligan:
So basic steps to get google to index a javascript site. 

1. Google notices that your page is rendered using Javascript. (sees the hashbang #! in your url)
2. Google requests your pages with a modified url. (replaces #! with ?_escaped_fragment_=)
3. You return the prerendered HTML to the crawler. 

So we could host our own pre-rendering HTML and do the custom routing, but I found a service (prerender.io) that has a phantom browser that generates the static HTML and caches it. You can have 250 unique urls(pages) for free.

Before merging we'd need create a resin account and add our token as an envar. 

@emirotin @shaunmulligan

### DIFF
--- a/static/scripts/main.js
+++ b/static/scripts/main.js
@@ -127,7 +127,12 @@ angular
 
       .otherwise('/pages/introduction/introduction.md');
   })
-
+  .config([
+     '$locationProvider',
+     function($locationProvider) {
+         $locationProvider.hashPrefix('!');
+     }
+  ])
   // services
   .service('PageRendererService', function($http) {
     this.getPageHtml = function(pageName) {
@@ -306,7 +311,7 @@ angular
       templateUrl: '/static/templates/directives/navigation.html',
       link: function(scope, el, attrs) {
         function addActiveClass() {
-          var activeEl = angular.element('.site-navigation ul a[href="/#/pages/'+ $routeParams.pageName +'"]').parent()
+          var activeEl = angular.element('.site-navigation ul a[href="/#!/pages/'+ $routeParams.pageName +'"]').parent()
 
           el.find('.expand').removeClass('expand');
           activeEl.addClass('active');
@@ -342,10 +347,3 @@ angular
       }
     };
   })
-
-angular.module('resinDocs').config([
-  '$locationProvider',
-  function($locationProvider) {
-      $locationProvider.hashPrefix('!');
-  }
-]);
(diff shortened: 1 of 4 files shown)


dfunckt:
If we get to actually render static pages, why not also serve them directly?

craig-mulligan:
@dfunckt we could but it would require a fair amount of work generate them and replace our current logic in `main.js`. This is meant as a quick fix as we are considering finding static site generator to replace our current system.

static/scripts/main.js
```diff
@@ -245,7 +249,7 @@ angular
         var url = location.href;
         url = url.replace(/#\w.+/, '')
```
emirotin: this seems to also needs updating to `#!`, no?

static/scripts/main.js
```diff
@@ -3,7 +3,7 @@ var UnpinOffset = 400;
 window.onload = function () {
     // fix first page load anchor issue
     var url = window.location.hash;
-    var divid = url.split('#');
+    var divid = url.split('#!');
```
emirotin: this looks wrong
this code splits the _has_ itself to find a hash-in-hash that identifies an element on the page (like heading) to scroll to it

emirotin:
First, a question: how would you set the env var? The site is hosted on github pages, right?

Second, as far as I know google should actually do normal crawling for JS pages and escaped_fragment shouldn't be required:
http://searchengineland.com/tested-googlebot-crawls-javascript-heres-learned-220157
http://www.infoq.com/news/2014/05/google-renders-css-js-hints

Third, this PR changes the URLs scheme, so it will make all the old URLs (in our UI, cross-linked in docs, in blog posts, in user's bookmarks, everywhere) broken, right?

craig-mulligan:
@emirotin I assumed it was hosted on heroku, because of the `Procfile`. It would break urls but if the crawler can do this without using _escaped_fragment_ URLs and rendering the html itself then I can't see why our docs could not be indexed?

emirotin:
Procfile makes sense, it's the gh-pages branch that makes me confused.
Maybe the key is `#!` in the URLs (check the most recent google link).
We should use webmaster tools and check how google sees our pages. Some of the linked articles should give insights into this process, I think.
Breaking URLs is super-bad :( If me manage to organize redirects (with angular-router's $otherwise clause) it's OK then, but first let's run the investigation.

PR #186, opened 2016-04-04
Some clarification on Raspberry Pi generations

marnikvde:
Although the text states "The Full Raspberry Pi family", I didn't find any info on RPI1 model A, and RP-zero. If these are not supported, it should probably be mentioned here.

### DIFF
--- a/pages/hardware/devices.md
+++ b/pages/hardware/devices.md
@@ -2,11 +2,11 @@
 
 ## Currently Supported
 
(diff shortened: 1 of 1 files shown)


shaunmulligan:
@marnikvde thanks for the PR, not sure if you are aware, but both the RPI1 model A and the RPI-zero work with the regular "Raspberry Pi (v1 or Zero)" image. Could you add that to your PR as well and then we will merge.

PR #187, opened 2016-04-05
pretty urls

emirotin:
So...

The main part of this PR is switching from `/#/pages/path.md` to `/pages/path`.

What else is done:
- code style improvements, performance improvements, old code cleanup
- rewriting old links in HREFs
- search results improvement — the index size is now slightly bigger, but it doesn't traverse the _navigation bar_ to get the document titles


craig-mulligan:
LGTM 👍  one thing that has come of this is that we should create a 404 page (in a seperate PR). Currently when you try to visit an old link it redirects to landing page which can be confusing.

PR #195, opened 2016-04-13
Add rpi3 to Nav

craig-mulligan:


emirotin:
add Zero as well?

craig-mulligan:
Or maybe we just put `Raspberry Pi` because we support all?

emirotin:
yeah. maybe `(all models)`

shaunmulligan:
yeah, i think (all models) will be good

shaunmulligan:
I think we can close this one now as it will get fixed in the switch to metalsmith

PR #213, opened 2016-04-27
Add rpi example projects

craig-mulligan:
@shaunmulligan @emirotin


emirotin:
LGTM. Just an (obvious) warning: the `$language_details` and `$device_details` variables are only available when partials are used within a dynamic page. So that nobody accidentally tries to reuse the same partial inside of a normal page.

craig-mulligan:
True, I'm going to change it to plain text just to be safe

PR #219, opened 2016-04-29
Add contrib info

shaunmulligan:
Adds info about how to contribute to the docs and a little bit about using partial and generating the dynamic pages.


README.md
```diff
@@ -26,14 +26,75 @@ npm start
...
 If you think something is not documented, or can be improved, don't hesitate in sending a PR!
 
+### Get Setup
+
+```sh
+git clone https://github.com/resin-io/docs.git
+cd docs
+npm install
+```
+Now to check that everything it working correctly, start the local server with:
```
alexandrosm: everything _is_ working correctly

README.md
```diff
@@ -26,14 +26,75 @@ npm start
...
+```
+However there is a handy watch functionality that will watch for changes in `/pages` and rebuild each time you save your changes. To use this, open a new tab in you terminal and run:
+```
+npm run watch-pages
+```
+Once you are happy with your changes, submit a pull request for you branch against `master`.
+
+### Using Partials
+
+When creating new content or altering current pages, its recommended that you try keep thing D.R.Y (Don’t Repeat Yourself). This is made easy by using “partials”. If you look in the `/shared` folder you will see a number of folders, these keep snippets or partials of the docs that can be reused all over the place. They also allow you to cleverly override a partial for a particular board type and language type. So in general you will see most of them just contain a `_default.md` which means that this partial will be used for all board and language types, but for partials like `getDeviceOnDash` we have device specific overrides.
```
alexandrosm: keep _things_ dry

README.md
```diff
@@ -26,14 +26,75 @@ npm start
...
+```
+and the engine will pull that partial into this location in the page.
+
+### Some other important bits
+
+Its important to note that `h1` and `h2` headings will automatically be included in the left hand side navigation bar, so make sure to use these appropriately.
+
+In the dynamically generated pages ( the ones with he device-type and language dropdown) it is possible to use templates to dynamically change the device name and language name in your text. Here is an example:
+```
+Getting started with {{ $device_details.name }} using {{ $language_details.name }}
```
alexandrosm: this is not a knock on the text, but I would be much happier if the _details thing was removed. looks like useless boilerplate.

README.md
```diff
@@ -26,14 +26,75 @@ npm start
...
+```
+Do add additional dynamic properties, see the files in `/config/dictionaries`.
```
alexandrosm: Do -> To?

alexandrosm:
besides the typos commented, LGTM

README.md
emirotin: `$device` is the device ID (the key from the `dictionaries` file). Then there's `$device_details` (the name is auto-built) and `$device_dictionary` (all possible devices). I think it's better to be explicit here and the original `$device` property can be useful (especially in helpers and in other parts of the code).

But I can also refactor it so that `$device` becomes `$device_id` and `$device_details` is `$device`.

alexandrosm: @emirotin how about $device.id, $device.name etc etc. think in terms of how the api looks, not how it was created.

alexandrosm: actually this is important enough to talk about in a bit more depth. let's catch up on monday or sth.

emirotin: If I just rename $device_details -> $device, $device.id, $device.name etc. will work out of the box.
Now I remember why I didn't go that way. There are 2 different contexts: Handlebars context (as here) and "meta-context" as with the partials search order field. In the meta context we say `$device+$language` and we expect it to expand to `raspnerrypi+python`, so there `$device` stands for its ID. It makes more sense to me that the value should be the same in both contexts. That's why `$DICTNAME` is reserved for the _axis_ value (which is an ID), and `$DICTNAME_details` is the entire object from the _DICT_. Let's talk on Mon.

alexandrosm: Actually I have a much better idea on how to do the partials search order, and it seems like it might solve this problem too, so even better.

alexandrosm: basically I was thinking that we could just specify the priority order of the dimensions, and the search order could be computed. Something like this:

```
_ = require("lodash");

var order = function(list){
  if(list.length == 0){ return; }

  results = []

  for(var i=list.length; i>0; i--){
    results.push((_.take(list, i).join("+")));
  }

  return results.concat(order(_.tail(list)));
}

console.log(order(["device","language","distro","os"]));
```

shaunmulligan: @emirotin is this contrib info still valid? i updated the PR, so if it is, lets merge it

emirotin: It is in the current version of the code. Let's merge, and then I'll apply changes in my PR that updates `doxx`

shaunmulligan:
@emirotin we can merge this now right?

PR #227, opened 2016-05-11
migrate to using Doxx as a stand-alone dependency

emirotin:
This is supposed to keep the entire functionality we have, moving the generic bits to the separate package https://github.com/resin-io-projects/doxx


emirotin:
It currently keeps the majority of the API intact to simplify the transition. I will later improve the doxx' API and migrate this repo to a newer version.

pages/tools/python-sdk.md
```diff
@@ -393,6 +393,24 @@ Get device name by device uuid.
```
emirotin: This file was refetched from the source.

PR #237, opened 2016-05-19
add generic i2c-python example

shaunmulligan:


shared/exampleProjects/beaglebone+python.md
```diff
@@ -0,0 +1,6 @@
+### I2C Accelerometer Example
```
emirotin: If you want to remove duplication between this one and the rpi-python you can move this paragraph into a normal partial and reuse in both places.

emirotin:
LGTM

PR #251, opened 2016-07-11
feat(docker): Add Dockerfile

### LINKED ISSUE #250: Dockerize Application
We should have a Dockerfile for running the docs using Docker.

brownjohnf:
@emirotin, As a step towards hosting this with Docker, I've created a Dockerfile. Can you take a look and see how it looks? Closes #250.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,8 @@ COPY . .
 
 # npm install needs to run after the COPY because of postinstall deps.
 RUN npm install --unsafe-perm --allow-root \
-	&& npm cache clean
+	&& npm cache clean \
+	&& node_modules/bower/bin/bower --allow-root cache clean
 
 ENTRYPOINT ["/usr/local/bin/npm"]
 CMD ["start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,14 @@
 FROM library/node:6.3
 
-WORKDIR /src
+EXPOSE 3000
 
-COPY package.json .
-
-RUN npm install
+WORKDIR /usr/src/app
 
 COPY . .
 
-RUN node_modules/bower/bin/bower --allow-root install \
-	&& tools/prepare.sh
-
-EXPOSE 3000
+# npm install needs to run after the COPY because of postinstall deps.
+RUN npm install --unsafe-perm --allow-root \
+	&& npm cache clean
 
 ENTRYPOINT ["/usr/local/bin/npm"]
 CMD ["start"]
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM library/node:6.3
+
+WORKDIR /src
+
+COPY package.json .
+
+RUN npm install
+
+COPY . .
+
+RUN node_modules/bower/bin/bower --allow-root install \
+	&& tools/prepare.sh
+
+EXPOSE 3000
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -22,6 +22,9 @@ To run the docs web application locally you should do the following:
 git clone https://github.com/resin-io/docs
 cd docs
 npm install
+npm install -g bower
```
emirotin: NIT, but I prefer using the bundled version (in which case the path should be adjusted, of course).

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN npm install
+
+COPY . .
+
+RUN node_modules/bower/bin/bower --allow-root install \
+	&& tools/prepare.sh
+
+EXPOSE 3000
+
+ENTRYPOINT ["/usr/local/bin/npm"]
```
emirotin: Is there a reason to use both entrypoint and the cmd?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM library/node:6.3
+
+WORKDIR /src
+
+COPY package.json .
```
emirotin: can do the same for bower to improve caching

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN npm install
+
+COPY . .
+
+RUN node_modules/bower/bin/bower --allow-root install \
```
emirotin: actually it should happen post npm-install thanks to the "postinstall" script in package.json

brownjohnf: Then we'd need to modify that in the package.json to use the local bin, right?

Dockerfile
brownjohnf: We could do it for the bower install, but if we leave it bundled with the tools/prepare.sh it won't work because prepare.sh requires more of the environment to be present.

Dockerfile
emirotin: nope, scripts configured in package.json by default have `node_modules/.bin` prepended to PATH when executed with `npm run `

Dockerfile
emirotin: Yes, if we care about caching we'll have to change it a bit. Also as I've mentioned currently npm will automatically run the postinstall script which will not work as expected. That script was added for heroku, and can be removed once we have properly working dockerfile and deployment procedure.

Dockerfile
brownjohnf: Oh good. I'm a JS n00b.

Dockerfile
brownjohnf: Setting the entrypoint to npm allows you to do things like `docker run resin/docs test` and have it be the equivalent of running `npm test` inside the container. That's sort of a contrived example, but you can do things like `start --production` to switch environments, etc.

emirotin: Oh, that's interesting.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+EXPOSE 3000
+
+WORKDIR /usr/src/app
+
+COPY . .
+
+# npm install needs to run after the COPY because of postinstall deps.
+RUN npm install --unsafe-perm --allow-root \
+	&& npm cache clean
```
emirotin: could also clean the bower cache

emirotin:
LGTM. Suggested a small improvement

PR #255, opened 2016-07-22
fix up table to proper markdown

imrehg:
The line ending `|` was not included in the table. On the web the
generated pages were okay, but eg. on Github the markdown table
did not show properly. Fix it up for future proofing.


emirotin:
lgtm

Issue #259, opened 2016-07-27
Improve docs shareability on social media


imrehg:
This would include, among other things:
- adding proper meta tags (Open Graph, for example)
- add preview image
- set page titles from the page data itself
- ...?

Testing: [Open Graph Object Debugger by Facebook](https://developers.facebook.com/tools/debug/og/object/)

imrehg:
Looks like might need to add some helper functions, unless these already exist (some might e.g. in Ghost, where we could check for good implementation:

* excerpt: get it from an excerpt tag, or by number of words, or letters, ignore first header if exists
* preview image: get it from tag, or find first image on page, or fall back to default image

PR #261, opened 2016-07-27
Minimal viable shareability updates v2

imrehg:
Along the way for #259, the minimal change that already improves things.

Incorporated comment on #260.


imrehg:
As of now it's two commit, one more is coming to be added this PR soon, adding a default preview image.

PR #285, opened 2016-08-10
updated linker

josephroberts:
Notified by johnboy14 in public chat that this has changed


imrehg:
I think it's the wrong place to change the settings. The updated variable is using partials, which are pulled from `config/dictionaries`. There it needs to be updated at the applicable devices.

imrehg:
LGTM

PR #307, opened 2016-08-25
Container Runtime Info

shaunmulligan:
@imrehg @craig-mulligan and @pcarranzav would love some review on this PR, added quiet a bit of stuff on resin container runtime and not sure if its 100% accurate or clear :P


pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
 ---
-title: Runtime Environment
+title: Container Runtime
 ---
+# Container Runtime
+
+On resin.io devices all your application code runs a [ Docker container][container-link]. This means that whatever you define as `CMD` in your `Dockerfile` will be PID 1 of the process tree in your container. It also means that this PID 1 process needs to know how to properly process UNIX signals, reap orphan zombie processes [[1]][ref-one] and if it crashes your whole container crashes, losing logs and debug info.
```
imrehg: Can the `ref-one` be an inline reference? I don't have a problem with `[1]` in general, though we don't normally use this citation style in general, so it stands out. (not problem, just comment)

config/navigation.txt
```diff
@@ -43,7 +43,6 @@ Learn
 
     Runtime Environment
       [/runtime/runtime]
```
imrehg: The URL like this looks a bit strange (the repeated words), could it work as `/runtime/container`, in line with the title? Though this would also need hunting down links to the page and updating those links, if any.
(all this is not problem, just comment)

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
+| `RESIN_SUPERVISOR_VERSION` 	|  The current version of the supervisor agent running on the device.	|
+| `RESIN_SUPERVISOR_API_KEY` 	|  Authentication key for the supervisor API. This makes sure requests to the supervisor are only coming from containers on the device. See the [Supervisor API reference][supervisor-api-link]	for detailed usage.|
+| `RESIN_SUPERVISOR_ADDRESS` 	|  The network address of the supervisor API. Default: `http://127.0.0.1:48484`	|
+| `RESIN_SUPERVISOR_HOST` 	  |  The IP address of the supervisor API.	Default: `127.0.0.1`|
+| `RESIN_SUPERVISOR_PORT` 	  |  The network port number for the supervisor API. Default: `48484`	|
+| `RESIN_API_KEY` 	          |  API key which can be used to authenticate requests to the resin.io backend. Can be used with resin SDK on the device. **WARNING** This API key gives the code full user permissions, so can be used to delete and update anything as you would on the Dashboard.  	|
+| `RESIN_DEVICE_RESTART` 	    |  This is a internal mechanism for restarting containers and can be ignored as its not very useful to application code.  Example: `1.13.0`	|
+
+Here's an example from a Raspberry Pi 3:
+```
```
imrehg: Add `Bash` style to help the highlighter's heuristics?

pages/runtime/runtime.md
```diff
@@ -36,17 +98,95 @@ var server = app.listen(80, function () {
...
+In some cases its useful to set up a service that starts up when your container starts. To do this with systemd, make sure you have the initsystem enabled in your container as mentioned [above](#init-system). You can then create a basic service file in your code repository called `my_service.service` and add something like this:
+```
+[Unit]
Description=My Super Sweet Service
+[Service]
Type=OneShot
ExecStart=/etc/init.d/my_super_sweet_service
+[Install]
WantedBy=basic.target
+```
```
imrehg: Code block above has some new-lines which are pasted as unicode chars.

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
-title: Runtime Environment
+title: Container Runtime
 ---
+# Container Runtime
+
+On resin.io devices all your application code runs a [ Docker container][container-link]. This means that whatever you define as `CMD` in your `Dockerfile` will be PID 1 of the process tree in your container. It also means that this PID 1 process needs to know how to properly process UNIX signals, reap orphan zombie processes [[1]][ref-one] and if it crashes your whole container crashes, losing logs and debug info.
+
+## Init System
+
+For these reasons we have built an [init system][init-system-link] into most of the resin base images listed here: [Resin Base Images Wiki][base-image-wiki-link]. The init system will handle signals, reap zombies and also properly handle [udev][udev-link] hardware events correctly.
```
imrehg: `udev-link` does not exists, so no link generated

shared/general/container-ssh.md
```diff
@@ -0,0 +1,25 @@
+To help you debug and develop your application in a resin.io container, we provided a browser based terminal or a command line tool called [resin ssh](/tools/cli/#ssh-60-uuid-62-). This gives you console access to your running [container][container-link] on the device and allows you to test out small snippets of code or check some system logs on your device.
+
+In order for you to start a terminal session in your device [container][container-link], you first need to ensure that your device is **online** and code is **pushed to it** and is running. If your container code crashes or ends quickly, it is not possible to attach a console to it. One option to keep your containers running is to enable the INITSYSTEM in your container. This can easily be done by creating a device environment variable called `INITSYSTEM` and setting its value to `on`.
```
imrehg: ... enable the `INITSYSTEM` in... (add ticks)

pages/runtime/runtime.md
```diff
@@ -36,17 +98,95 @@ var server = app.listen(80, function () {
 })
 ```
 
-## Persistent Storage		
+## Access to /dev
+
+In many projects you may need to control or have access to some external hardware via interfaces like GPIO, I2C or SPI. On resin.io your container application will automatically have access to `/dev` and these interfaces since the container is run in [**privileged** mode](https://docs.docker.com/engine/reference/commandline/run/#/full-container-capabilities-privileged). This means you should be able to use any hardware modules like you would in a vanilla linux environment.
+
+__Note:__ If you are not using one of the docker base images recommended in our [base images wiki][base-image-wiki-link], then its most likely you will need to handle the updating of `/dev` via [udev][udev-link] yourself. You can see an example of how our base images handle this [here](https://github.com/resin-io-library/base-images/blob/master/debian/armv7hf/jessie/entry.sh#L54).
```
imrehg: `udev-link` again

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
 
 ## Exposed Ports
 
 Resin.io devices expose all ports by default, meaning you can run applications
 which listen on any port without issue. There is no need to have the docker `EXPOSE` command in your `Dockerfile`.
 
-## Device URLS
+## Public Device URLS
 
 Resin.io currently exposes port 80 for web forwarding. To enable web forwarding on a specific device, navigate to the device's **actions** tab on the resin.io dashboard and select the `Enable a public url for this device` checkbox. For more information about device URLS you can head over to the [Device Management Page](/management/devices#enable-public-device-url)
```
imrehg: Actually port 8080 is exposed as well, but only on http connection.

* port 80 exposed at: `https://<UUID>.resindevice.io` and `http://<UUID>.resindevice.io`
* port 8080 exposed at: `http://UUID.resindevice.io:8080`

Thus can have two distinct applications listening. The UI does not give any indication to this, though, just mentioned in chat.

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
+## Init System
+
+For these reasons we have built an [init system][init-system-link] into most of the resin base images listed here: [Resin Base Images Wiki][base-image-wiki-link]. The init system will handle signals, reap zombies and also properly handle [udev][udev-link] hardware events correctly.
+
+There are two ways of enabling the init system in your application. You can either add the following environment variable in your Dockerfile:
+```Dockerfile
+# enable container init system.
+ENV INITSYSTEM=on
+```
+or you can add an environment variable from the Dashboard by navigating to the `Environment Variables` menu item on the left and adding the variable as show below:
```
craig-mulligan: as show below -> as shown below

pages/runtime/runtime.md
emirotin: I've already commented on that somewhere. It was mentioned before, then 8080 was deprecated and we've hidden any mentions of it. But it's still working. I don't know why we started speaking about it again — is it that we get many requests to support multiple ports?

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
+![Enable init system](/img/common/app/app_initsystem_envvar.png)
+
+Once you have enabled your init system you should see something like this in your device logs:
+![init system enabled in logs](/img/common/device/device_logs_initsystem_enabled.png)
+
+You shouldn't need to make any adjustments to your code or `CMD` it should just work out of the box. Note that if you are using our Debian or Fedora based images, then you should have [systemd][systemd-link] in your containers, where as if you use one of our Alpine images you will have [OpenRC][openrc-link] as your init system.
```
craig-mulligan: You shouldn't need to make any adjustments to your code or `CMD`. Why not say you **won't** need to make any adjustments? Are there cases when the init system messes with someones app?
 
where as if you use -> whereas if you use

pages/runtime/runtime.md
imrehg: @emirotin my bad then, didn't get the note that it was deprecated. Then it's good as it is.

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
+
+## SSH Access
+
+{{> "general/container-ssh"}}
+
+## The Container Environment
+
+When you start a terminal session, either via the web terminal or the CLI, you are dropped into your applications running container. Its important to note that your container needs to be running for you to actually SSH into it, this is where the [init system](#init-system) helps a lot. By default you are the `root` user and granted root privileges in the container.
```
craig-mulligan: Its important -> It's

pages/runtime/runtime.md
craig-mulligan: its -> it's

pages/runtime/runtime.md
```diff
@@ -1,21 +1,83 @@
...
 
-## Device URLS
+## Public Device URLS
 
 Resin.io currently exposes port 80 for web forwarding. To enable web forwarding on a specific device, navigate to the device's **actions** tab on the resin.io dashboard and select the `Enable a public url for this device` checkbox. For more information about device URLS you can head over to the [Device Management Page](/management/devices#enable-public-device-url)
 
 ![Enable device url](/img/screenshots/device-url-new.png)
 
-
 Running a server listening on port 80 with public device url enabled will allow you to serve content from the device to the world. Here is an example of an [express.js][expressjs-link] server which will serve to the devices url.
```
craig-mulligan: I'd change all instances of url -> URL

pages/runtime/runtime.md
```diff
@@ -36,17 +98,95 @@ var server = app.listen(80, function () {
...
+
+__Note:__ If you are not using one of the docker base images recommended in our [base images wiki][base-image-wiki-link], then its most likely you will need to handle the updating of `/dev` via [udev][udev-link] yourself. You can see an example of how our base images handle this [here](https://github.com/resin-io-library/base-images/blob/master/debian/armv7hf/jessie/entry.sh#L54).
+
+## Tips, Tricks and Troubleshooting
 
-If you want specific data or configurations to persist on the device through the update process, you will need to store them in `/data` . This is a special folder on the device file system which is essentially a [docker data `VOLUME`][docker-volume-link].
+### Writing to logs on the Dashboard
+
+In general anything written from the application to `stdout` and `stderr` should appear on the device's dashboard logs. Have a look at some of our [example projects][projects-github] on github to get an idea of how to do this.
```
craig-mulligan: Why do you say `in general`? isn't it always forwarded or are you accounting for the bugs in logs? 

I'd just say: Anything written from the application to `stdout` and `stderr` will appear on the device's dashboard logs.

craig-mulligan:
LGTM, in some cases I feel like you could add more confidence to the language. Eg. Replacing `should` with `will`. Great to have environment documented and all the tips + tricks. 👍

pages/runtime/runtime.md
shaunmulligan: yeah, thats a fair point, I'll make it inline

config/navigation.txt
shaunmulligan: yeah, its definitely not great. I have a plan to fix up all the URLs and restructure some of the menu, but it will be in a later PR, this one is more content focused for now

pages/runtime/runtime.md
shaunmulligan: @emirotin yeah we get quiet a few people that want to add some extra metrics or admin stuff on a different port and feel like they can't do it until we tell them about port 8080. We should allow arbitrary ports. For the time being I will leave port 8080 out of this doc. we can pick it up later.

pages/runtime/runtime.md
shaunmulligan: @craig-mulligan , i think i was thinking along the lines of stuff you would put in your start.sh. in the case you dont have systemd you would probably start the binary in the start.sh script.

shaunmulligan:
@craig-mulligan and @imrehg addressed most of the comments and responded where appropriate. Lets get this guy out there

imrehg:
@shaunmulligan thanks for the updates, look awesome, and comments make sense! :)

And sorry, I had to push out a change in the meantime for Bryan, which conflicts with this one in `config/navigation.txt`, I guess a rebase on master would do the trick to resolve?

emirotin:
Rebase and resolve is the way to go indeed.

pages/runtime/runtime.md
```diff
@@ -36,17 +98,101 @@ var server = app.listen(80, function () {
...
+```
+curl -X POST --header "Content-Type:application/json" \
+    "$RESIN_SUPERVISOR_ADDRESS/v1/reboot?apikey=$RESIN_SUPERVISOR_API_KEY"
+```
+[Read more about the supervisor API](/runtime/supervisor-api/#post-v1-reboot)
+
+__Note:__ `RESIN_SUPERVISOR_API_KEY` and `RESIN_SUPERVISOR_ADDRESS` should already be in your environment by default. You will also **need** `curl` installed in your container.
 
-If you want specific data or configurations to persist on the device through the update process, you will need to store them in `/data` . This is a special folder on the device file system which is essentially a [docker data `VOLUME`][docker-volume-link].
+<!-- TODO: explain how to reboot from systemd -->
```
shaunmulligan: to reboot from systemd:

DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host_run/dbus/system_bus_socket dbus-send --system --print-reply --dest=org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager.Reboot

shaunmulligan:
rebased and resolved. merging....

PR #308, opened 2016-08-26
Update base-images documentation:

nghiant2710:
- Refactor content structure.
- Add information for new boards
- Add information for fedora base images.


shaunmulligan:
@nghiant2710 LGTM, but i think we might need to move towards having resin base images as a top level item in the docs menu, then we can have a nice menuing of the architectures and the device types on the side and also have it dynamically show the appropriate base images for whatever device they user has selected.

nghiant2710:
@shaunmulligan yeah, I like that idea. I will merge this PR first then work on it later. Created an issue for it https://github.com/resin-io-library/base-images/issues/180

PR #310, opened 2016-08-30
rename network setup page

imrehg:
Connects to #309

### DIFF
--- a/config/redirects.txt
+++ b/config/redirects.txt
@@ -15,4 +15,4 @@
 ^\/configuration\/resin-base-images(\/|$) -> /runtime/resin-base-images
 
 # Renamed pages
-^\/deployment\/wifi\/?(.*)? -> /deployment/network/$1
+^\/deployment\/wifi\/?(#.+)?$ -> /deployment/network/$1
--- a/config/redirects.txt
+++ b/config/redirects.txt
@@ -15,4 +15,4 @@
 ^\/configuration\/resin-base-images(\/|$) -> /runtime/resin-base-images
 
 # Renamed pages
-^\/deployment\/wifi\/ -> /deployment/network/
+^\/deployment\/wifi(\/.*)? -> /deployment/network/$1
--- a/config/redirects.txt
+++ b/config/redirects.txt
@@ -13,3 +13,6 @@
 ^\/getting-started -> /raspberrypi/nodejs/getting-started
 
 ^\/configuration\/resin-base-images(\/|$) -> /runtime/resin-base-images
+
+# Renamed pages
+^\/deployment\/wifi\/ -> /deployment/network/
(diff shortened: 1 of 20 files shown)


config/redirects.txt
```diff
@@ -13,3 +13,6 @@
 ^\/getting-started -> /raspberrypi/nodejs/getting-started
 
 ^\/configuration\/resin-base-images(\/|$) -> /runtime/resin-base-images
+
+# Renamed pages
+^\/deployment\/wifi\/ -> /deployment/network/
```
emirotin: could be 

`^\/deployment\/wifi(\/.*)? -> /deployment/network/$1` to keep the rest of the URL

imrehg: fixed

emirotin:
I didn't test the regex part for the redirect. If it works for you everything LGTM

imrehg:
@emirotin after testing, updated the regex, so the slash is not in the matched group, otherwise end up with two slashes. How does this one look to you? (tested out working, hopefully in enough detail:)

config/redirects.txt
```diff
@@ -13,3 +13,6 @@
 ^\/getting-started -> /raspberrypi/nodejs/getting-started
 
 ^\/configuration\/resin-base-images(\/|$) -> /runtime/resin-base-images
+
+# Renamed pages
+^\/deployment\/wifi\/?(.*)? -> /deployment/network/$1
```
emirotin: Yeah, looks good. The problem I'd tried to solve was if there are more URLs like

`/deployment/wifi-something` that will be captured here as well now.

imrehg: Very good point! How's this?  `^\/deployment\/wifi([\/]?|[\/]?(#.*)?)$ -> /deployment/network/$2`
Tested here: http://regexr.com/3e4m8 Let me know if you have any idea to simplify it....
If okay, I'll add a new commit. Cheers!

emirotin: it works and seems to be equivalent to this: `^\/deployment\/wifi(\/?(#.*)?)$`

because:
* `a|ab` is equivalent to `ab?`
* `[\/]` is equivalent to `\/` (because you don't actually need the backward slash in the character class, and you don't need the class for a single character)

I would probably also replace `(#.*)` with `(#.+)` because we only care about non-empty fragments

emirotin: Ah, and you also don't need the outer group:

`^\/deployment\/wifi\/?(#.+)?$` (and then you can use `$1` in the replacement

http://regexr.com/3e4mh

imrehg: Yeah, that great, thanks a lot! :)

shaunmulligan:
@imrehg what is the new URL supposed to be? when i click on the "network setup" menu item it takes me to http://localhost:3000/deployment/network/ but it says "page not found"

PR #314, opened 2016-08-30
adding description of delta updates

imrehg:
Adding minimal amount of description to be able to share with partners.


imrehg:
Draft version, will merge it to preview so it can be shared (time sensitive). Will expand later if needed.

shaunmulligan:
yeah we will need to expand this, I had hoped to do a Updates + App Lifecycle section in runtime section. This would include deltas, update strategies, restart policies, signals sent to containers, etc.

PR #325, opened 2016-09-10
Improve runtime document: add environment file to systemd service

nghiant2710:
I think it will be better if we add the environment file in the example service file since many users asked about this


shaunmulligan:
@nghiant2710 that's a good addition, thanks. LGTM

PR #336, opened 2016-09-23
Fix docker-images-containers link

dgallegos:
Docker moved the explanation for Docker images and containers.
Old Link: https://docs.docker.com/engine/introduction/understanding-docker/#inside-docker
New Link: https://docs.docker.com/engine/understanding-docker/#/inside-docker


emirotin:
Thanks!

emirotin:
Note: it could take several days before the change is deployed.

PR #343, opened 2016-10-19
Distinguish Param from Payload

gernest:
This PR tries to fix the confusion that arise when Payload is Mixed with Param

From my understanding, we use Param to specify parameters that are passed with the URL.  This can be query params after the ? part of the URL or some placeholder in the case of something like `/device(<ID>)`.

On the other hand, what is written in the request body is Payload, which in the cURL examples are passed with the `-data` flag`.

### DIFF
--- a/pages/runtime/data-api.md
+++ b/pages/runtime/data-api.md
@@ -71,6 +71,9 @@ curl "{{ $links.apiBase }}application?\$filter=app_name%20eq%20'<name>'"
 Param | Type | Description
 ------|------|------------:
 app_name | `STRING` | application name
+
+Payload | Type | Description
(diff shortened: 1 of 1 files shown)


pages/runtime/data-api.md
```diff
@@ -71,6 +71,9 @@ curl "{{ $links.apiBase }}application?\$filter=app_name%20eq%20'<name>'"
 Param | Type | Description
 ------|------|------------:
 app_name | `STRING` | application name
```
imrehg: Looking at the API, `app_name` should be also Payload

gernest:
@imrehg fixed

PR #352, opened 2016-10-28
update information on using the RPi camera

imrehg:


pages/hardware/i2c-and-spi.md
```diff
@@ -76,16 +76,35 @@ An example of this is shown in our [Firebase Temperature Logger][firebaseTemp-li
 
 ### Raspberry Pi camera module
 
-In order to work with the Raspberry Pi camera module you will need to do the following:
+Depending on the version of your ResinOS, the system contains different version of the Raspberry Pi firmware, and you need to apply slightly different settings. In both cases you can either modify `config.txt` on the `resin-boot` partition of your SD card, or add the settings remotely by using `RESIN_HOST_CONFIG_variablename` settings in your [fleet or device configuration](/management/env-vars/).
```
emirotin: Should it be `resinOS` with lowercase `r`?

emirotin: I see. I've fixed that in couple of places in the dashboard some time ago, but now found couple more. And it looks like what's shown for the new versions is actually in the **DB** and comes from the supervisor.

shaunmulligan: I agree they we should be consistent and call it `resinOS`everywhere.

lurch:
Does https://github.com/resin-io-projects/resin-rpi-python-picamera/blob/master/start.sh need to be updated too?

PR #373, opened 2016-11-16
add info on disabling hdmi on BBB

shaunmulligan:
This PR documents how to disable HDMI on the Beaglebone Black so one can make use of all the GPIO pins on the device.

### DIFF
--- a/pages/hardware/i2c-and-spi.md
+++ b/pages/hardware/i2c-and-spi.md
@@ -216,7 +216,7 @@ fdtfile=am335x-boneblack-emmc-overlay.dtb
 ```
 
 You should now be able to use any of the pins that used to be occupied by the HDMI. To test this you can run the following from your webterminal:
-```
+```bash
 echo 74 > /sys/class/gpio/export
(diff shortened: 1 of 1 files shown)


pages/hardware/i2c-and-spi.md
```diff
@@ -206,6 +207,22 @@ OCPDIR=/sys/devices/platform/ocp/ocp*
 SLOTS=/sys/devices/platform/bone_capemgr/slots
 ```
 
+### Disable HDMI
+
+before provisioning, mount the `.img`, in the `flash-boot` partition you should see a file named `uEnv.txt_internal` open that up with your favourite text editor and add
```
lurch: Capital B on 'before'

pages/hardware/i2c-and-spi.md
```diff
@@ -206,6 +207,22 @@ OCPDIR=/sys/devices/platform/ocp/ocp*
...
+the following line to the bottome of the file:
```
lurch: 'bottome' -> 'bottom'

Issue #265, opened 2016-07-28
Document UART usage on RPI3


shaunmulligan:
There seems to be difficulties for users getting this right, we need proper documentation on how to use it.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/inboxes/topics/24925)

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_j8d)

imrehg:
This and a bunch of other RPi hardware guidance would be very timely. For example there's a lot of BLE question too, that is non-obvious. The best would be probably to add an RPi "cookbook" or "tips & tricks" section or page.

imrehg:
Add also `RESIN_HOST_CONFIG_dtoverlay` usage with `pi3-miniuart-bt-overlay`.

shaunmulligan:
to enable uart on RPI3:

RESIN_HOST_CONFIG_core_freq = 250

RESIN_HOST_CONFIG_dtoveraly=pi3-disable-bt

RESIN_HOST_CONFIG_enable_uart=1

PR #374, opened 2016-11-16
Rpi3 hardware

shaunmulligan:
This PR adds info on using UART0 on the rpi3 and is associated with this demo project: https://github.com/resin-io-playground/rpi3-uart

It closes issue #265


lurch:
Doesn't really matter, since they've both been merged, but it looks like this PR accidentally contains the commits from #373 too.

PR #409, opened 2017-01-25
Typo in "dive into code"

glikson:


shared/diveIntoCode/go.md
```diff
@@ -53,4 +53,4 @@ CMD ./resin-go-hello-world
 [resinDockerHub]:https://hub.docker.com/u/resin/
 [docker-ref]:https://docs.docker.com/engine/reference/builder/
 [systemd-link]:https://en.wikipedia.org/wiki/Systemd
-[listOfResinNames]:/devicetypes/
\ No newline at end of file
+[listOfResinNames]:/devicetypes/
```
emirotin: What is the change on this line? A new line at the end of the file?

glikson: No idea - I don't remember touching it. Maybe a vi/git 'side-effect'.

emirotin: Actually it is (which is good), but for some reason GH doesn't show it, but an email from it does show.

emirotin:
Thanks for the fix!

lurch:
Should the end of the line also say "remove the line from the `Dockerfile`." instead of "remove the line for the `Dockerfile`."?
Let me know if so and I'll submit a PR.

PR #420, opened 2017-02-06
redirect old cli link

craig-mulligan:
Reported here: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/d67HySDyrbn5HENvXSuslvtBeTu


config/redirects.txt
```diff
@@ -17,3 +17,4 @@
 # Renamed pages
 ^\/deployment\/wifi\/?(#.+)?$ -> /deployment/network/$1
 ^\/runtime/terminal -> /runtime/runtime#using-the-web-terminal
+^\/using/cli -> /tools/cli
```
emirotin: Does it work? I thought the forward slashes must be escaped :)

PR #423, opened 2017-02-08
add note about dbus mount point for > resinOS v2

craig-mulligan:
We should definitely add resinOS vs into the docs matrix. @shaunmulligan I'll create an issue if you agree?

### DIFF
--- a/pages/runtime/runtime.md
+++ b/pages/runtime/runtime.md
@@ -139,6 +139,8 @@ Alternatively, it is possible to reboot the device via the dbus interface as des
 
 ### Dbus communication with hostOS
 
+__Note:__ If you are using resinOS greater than `v2` the mount point has changed to `/host/run/dbus/system_bus_socket`. 
+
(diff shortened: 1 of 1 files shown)


craig-mulligan:
Second thought this maybe better to PR into the resinOS docs, and only add v2 specifics once it's released on production?

imrehg:
@craig-mulligan @shaunmulligan sorry, didn't see that it was here (wasn't linked from the issue), and made this update #431
Either one is good for me :)

shaunmulligan:
@craig-mulligan we should probably get people to use `/host/run/dbus` on 1.x now anyways. so it might be worth changing the doc to say use that and if you are on less than 1.13 (or whatever it is) then we should use the old path.

craig-mulligan:
@shaunmulligan Cool, I had a look through the [meta-resin changelog](https://github.com/resin-os/meta-resin/blob/master/CHANGELOG.md) but didn't see any mention of it?

shaunmulligan:
@craig-mulligan the supervisor adds all the bind-mounts to the user containers, so its most likely tied to a specific supervisor version.

craig-mulligan:
Lets go with Gregley PR #431 closing this one.

PR #424, opened 2017-02-09
Not actually a wiki

sonyagreen:
Docs is a more accurate word

### DIFF
--- a/config/navigation.txt
+++ b/config/navigation.txt
@@ -95,4 +95,4 @@ Reference
   CLI Reference[/tools/cli]
   Node.js SDK[/tools/sdk]
   Python SDK[/tools/python-sdk]
-  Base Images Wiki[/runtime/resin-base-images]
+  [/runtime/resin-base-images]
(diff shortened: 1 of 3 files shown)


sonyagreen:
Oh, I'm not sure if the name will get picked up where it's being referenced (sidebar, top nav) by the # or if that requires a separate change somewhere.

lurch:
From a quick scan through the [search results](https://github.com/resin-io/docs/search?utf8=%E2%9C%93&q=wiki&type=Code) it looks like [config/navigation.txt](https://github.com/resin-io/docs/blob/master/config/navigation.txt) and [pages/introduction.md](https://github.com/resin-io/docs/blob/master/pages/introduction.md) might need to be updated in concert?

imrehg:
@sonyagreen @lurch 
In this particular case I went ahead and fixed the mentioned issue (won't normally push onto other people's PR, but here it felt appropriate to fix up the uncovered errors, let me know if I'm wrong!

* `pages/introduction.md` updated
* in the `navigation.txt` one can just remove the page name (see change), and then it will be pulled from the actual markdown file. For most pages that should be the default, unless we want to prevent overflow in the navigation.

shaunmulligan:
LGTM

PR #430, opened 2017-02-13
[Issue #425] Dockerize resin docs

brownjohnf:


Dockerfile
```diff
@@ -1,15 +1,20 @@
...
-WORKDIR /usr/src/app
+RUN npm install
 
 COPY . .
 
-# npm install needs to run after the COPY because of postinstall deps.
-RUN npm install --unsafe-perm --allow-root \
-	&& npm cache clean \
-	&& node_modules/bower/bin/bower --allow-root cache clean
+RUN node_modules/bower/bin/bower --allow-root install \
```
craig-mulligan: Why not run bower before copying all the src?

craig-mulligan: oh I see bower installs into /static :/

brownjohnf: I'm open to more efficient approaches.

emirotin: I think it should work — bower would create a dir and then the rest will be merged when copying?
If not we can first `mkdir -p` it and ensure the source directory is `.dockerignore`d just in case (if building locally for example).

Also NIT: the path can be `node_modules/.bin/bower` —

Issue #492, opened 2017-04-28
remove bower 


craig-mulligan:
Currently we do multiple requests for various assets, we should remove bower and build assets into a single bundle.

PR #493, opened 2017-04-28
Build assets with webpack

craig-mulligan:
connects #492 

* remove any bower deps
* build all assets with webpack

Change-Type: minor


craig-mulligan:
ping @pimterry @emirotin if you get a chance today could you take a look?

Issue #488, opened 2017-04-25
Document auth token


craig-mulligan:
Valid for 7 days and can be refreshed hitting https://api.resin.io/whoami.

pimterry:
More or less duplicates #458, but a bit more generally, we should close one of these.

PR #494, opened 2017-05-01
Document /whoami endpoint

craig-mulligan:
Connects to #488

Change-Type: patch


craig-mulligan:
@Page- How do does the api versioning work, this endpoint isn't behind the `/v1` path, do we version per endpoint?

PR #495, opened 2017-05-01
Update gettingStarted-Parallella.md

cmr:
Fixes some of the details that seemed suspect to me.


shaunmulligan:
Thanks for the contribution @cmr ! this all looks great to me, so I'll merge 👍

PR #502, opened 2017-05-08
Make npm run in the docker build

craig-mulligan:
Connects 501

Change-Type: patch


craig-mulligan:
I assumed `npm run build` ran in the docker build, but it doesn't so now staging is broke because of #493.

PR #506, opened 2017-05-11
Best practices org account

afitzek:
Describe best practices for organizational accounts


afitzek:
@sonyagreen Thanks for your comment. I took your version and added a sentence for the section in the brackets. What do you think?

PR #509, opened 2017-05-15
Add dev vs. prod os distinction and better git remote add resin example

zwalchuk:
Include a warning in the Getting Started guide about not using dev images in production, as well as a link to the dev vs. prod details.

Update the example command for "git remote add resin" example to include generic endpoint, rather than charlie1 example endpoint.


zwalchuk:
@imrehg Good idea, I've updated for all the getting started pages and the FAQs. At some point here I'll get together a style guide for the docs and tutorials to keep these things consistent.

PR #545, opened 2017-07-24
Getting started updates

zwalchuk:
Converting the NUC, Nitrogen 6X, and Hummingboard getting started guides to the parametric format.

Adding getting started guide for the DART-6UL.


zwalchuk:
@shaunmulligan right now that one is using the generic instructions, which basically says you need to make sure the image is on the primary boot media. I can put specific instructions in there using the development kit if we have them

shaunmulligan:
@zwalchuk yeah I think that would be helpful. I also think it might make sense to split the jetson TX2 out from these other boards, because then we can merge them right aways, as the Jetson might still take a little while until its in production.

zwalchuk:
@shaunmulligan that makes sense, I was doing that with the CHIP so I'll do it with the Jetson as well. That will give us a little time to get those sections up to snuff on both

shaunmulligan:
sounds good 👍

Issue #535, opened 2017-07-11
Clarify that network connections are on the SD card partitions


imrehg:
In https://docs.resin.io/deployment/network/2.0.0/#wifi-setup it's not clear that `/resin-boot/` is not a folder (and not within the application container), but it's a partition of the SD card, and `system-connections` is a folder on that partition.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_208p)

Issue #513, opened 2017-05-23
Forcing wifi to be the primary route to the internet on resinOS 2.0


shaunmulligan:
To set wifi as the default route to the internet, we need to force `eth0` to have a lower route metric and set the `never-default` option. One should be able to just drop the following ethernet configuration file in `/boot/system-connections`

```
[connection]
id=my-ethernet
type=ethernet
interface-name=eth0
permissions=
secondaries=

[ethernet]
mac-address-blacklist=

[ipv4]
never-default=true
route-metric=2000
dns-search=

ignore-auto-routes=true
method=auto

[ipv6]
addr-gen-mode=stable-privacy
dns-search=
method=auto
```

Issue #485, opened 2017-04-18
Document how to connect to an unsecured wifi network


pimterry:
Fairly easy in the end: just remove the `[wifi-security]` section completely.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_1hnh)

Issue #479, opened 2017-04-11
Update network troubleshooting to reflect 2.0 changes


shaunmulligan:
https://docs.resin.io/troubleshooting/error/#unable-to-connect-to-the-internet

also mention that one should power the device down and mount the SD card.

PR #547, opened 2017-07-31
Updates to network-related documentation

zwalchuk:
A few updates to address issues #479, #485, #513, and #535.

### DIFF
--- a/pages/troubleshooting/error.md
+++ b/pages/troubleshooting/error.md
@@ -15,7 +15,7 @@ If the Raspberry pi is unable to connect to the resin.io servers, the `ACT` LED 
 
 This is either because it is not connected to the network or because the network ports which resin.io relies on are blocked in some way.
 
-* The first things to check in this case is that your device is correctly connected to ethernet or that you correctly entered the wifi credentials. To check that the wifi credentials are correct, you can check the `config.json` file on the `resin-boot` partition of the SD card (since Resin OS 1.2, or `resin-conf` for earlier versions).
+* The first things to check in this case is that your device is correctly connected to ethernet or that you correctly entered the wifi credentials. To check wifi credentials, power your device down, remove the SD card, and mount the SD card on your personal computer. If your device is running resinOS version 2.0 or greater, wifi credentials are listed in `system-connections/resin-wifi`, found in the `resin-boot` partition of the SD card. Otherwise, check the `config.json` file (in the `resin-boot` partition for versions 1.2 and greater, or `resin-conf` for earlier versions).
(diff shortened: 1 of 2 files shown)


pages/troubleshooting/error.md
```diff
@@ -15,7 +15,7 @@ If the Raspberry pi is unable to connect to the resin.io servers, the `ACT` LED
 
 This is either because it is not connected to the network or because the network ports which resin.io relies on are blocked in some way.
 
-* The first things to check in this case is that your device is correctly connected to ethernet or that you correctly entered the wifi credentials. To check that the wifi credentials are correct, you can check the `config.json` file on the `resin-boot` partition of the SD card (since Resin OS 1.2, or `resin-conf` for earlier versions).
+* The first things to check in this case is that your device is correctly connected to ethernet or that you correctly entered the wifi credentials. To check wifi credentials, power your device down, remove the SD card, and mount the SD card on your personal computer. If your device is running resinOS version 2.0 or greater, wifi credentials are listed in `system-connections/resin-wifi`, found in the `resin-boot` partition of the SD card. Otherwise, check the `config.json` file (in the `resin-boot` partition for versions 1.2 and greater, or `resin-conf` for earlier versions).
```
shaunmulligan: One note here is that any device that is a `flasher-type` its not possible to inspect the running device boot partition (because its inside the device), however they can probably check the SD card anyways, since that would have the "initial configuration"

shared/networking/2.0.0.md
```diff
@@ -45,7 +45,34 @@ psk=super_secret_wifi_password
...
 This file sets up a simple wifi connection to a network named `My Awesome Wifi Ssid` with a wifi password (`psk`) of `super_secret_wifi_password`. If you want to add multiple different wifi credentials, then simply make a copy of `resin-wifi` or `resin-sample` and change the `SSID` and `psk` key values. You can also add the `autoconnect-priority` integer values in each file to define the priority of each connection when the device has multiple wifi connections it can connect to.
 
-The connection file also defines what kind of security the network expects, if you want to setup a network with a more elaborate security requirements, checkout the [NetworkManager settings reference page][nm-setting-ref].
+To set wifi as the default route to the internet, you need to force `eth0` to have a low route metric and set the `never-default` option. The following ethernet configuration file can be put in the `/system-connections/` folder to accomplish this:
+
+```
+[connection]
+id=my-ethernet
+type=ethernet
+interface-name=eth0
```
shaunmulligan: I probably should be noted that not all ethernet ports will be named in this way. On newer x86 devices they have some weird naming. So it would be recommended that a user first check their ethernet interface name using something like `ifconfig` from the webterminal.

PR #568, opened 2017-08-22
Add a FAQ for discontinued devices

zwalchuk:
Relates to https://github.com/resin-io/resin-ui/pull/1156 and https://github.com/resin-io/hq/pull/785


pages/troubleshooting/faq.md
```diff
@@ -120,3 +121,6 @@ The resin.io agent needs to be able to access our cloud services in order for yo
 
 ##### How do you secure your own "cloud" to prevent malicious attack which may allow attacker to break-in our systems?
 Generally we try to follow good OPSEC practices for our systems. We support 2FA for user accounts and force all the connections to be over HTTPS.
+
+##### What does it mean when a device type is discontinued?
+Discontinued devices will no longer be actively supported by resin.io. This means we will no longer provide prebuilt versions of resinOS for these devices, and we will not be resolving any issues related to these boards. Please contact sales@resin.io with any questions regarding continued device support.
```
shaunmulligan: We should also add the at we wont allow the creation of new applications for these device types, but already existing applications and the devices associated with it will continue to function as normal, but will not have new OS or supervisor releases.

shaunmulligan: It might also be worth mentioning that it will still be possible to build versions of resinOS for this device type from the open source repos at github.com/resin-os

PR #590, opened 2017-09-15
Make a list with all Local Mode requirements

### LINKED ISSUE #589: User misses some of the requirements of using local mode
Front logo [Front conversations](https://app.frontapp.com/open/cnv_67f1ev)

I also missed some of them on first read.
Changing to a checklist might make it easier to the reader.

thgreasi:
Fixes #589

### DIFF
--- a/pages/development/local-mode.md
+++ b/pages/development/local-mode.md
@@ -8,9 +8,12 @@ Local mode is the new experimental development mode for resin.io. It allows you 
 
 <img src="/img/local-mode/local-mode-diagram.png" width="100%">
 
-Local mode will only work on devices running resinOS v2.0 or higher with supervisor v4.0 or higher. It will only work on `development` variants of the OS—if you try to use a `production` variant, you will not be able to use local mode. The production devices have ssh and the Docker socket locked down, both of which are needed by the local mode feature.
-
-To start using local mode on one of your development devices, simply click on the small *Actions* dropdown at the top right of the device page and select *Enable Local Mode*.
+In order to use local mode on a device:
+* The device has to be running resinOS v2.0 or higher with supervisor v4.0 or higher.
+* The device has to be running a `development` variant of the OS.  
+  If you try to use a `production` variant, you will not be able to use local mode. The production devices have ssh and the Docker socket locked down, both of which are needed by the local mode feature.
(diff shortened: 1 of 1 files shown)


pages/development/local-mode.md
```diff
@@ -8,9 +8,12 @@ Local mode is the new experimental development mode for resin.io. It allows you
...
 <img src="/img/local-mode/local-mode-diagram.png" width="100%">
 
-Local mode will only work on devices running resinOS v2.0 or higher with supervisor v4.0 or higher. It will only work on `development` variants of the OS—if you try to use a `production` variant, you will not be able to use local mode. The production devices have ssh and the Docker socket locked down, both of which are needed by the local mode feature.
-
-To start using local mode on one of your development devices, simply click on the small *Actions* dropdown at the top right of the device page and select *Enable Local Mode*.
+In order to use local mode on a device:
+* The device must be running resinOS v2.0 or higher with supervisor v4.0 or higher.
+* The device must be running a [development](https://docs.resin.io/understanding/understanding-devices/2.0.0/#dev-vs-prod-images) variant of the OS.  
+  If you try to use a production variant, you will not be able to use local mode. The production devices have SSH and the Docker socket locked down, both of which are needed for the local mode feature.
+* Local mode must be enabled through the dashboard.
```
craig-mulligan: NIT: I'd say `must` -> `can` because it's also accessible via the api.

thgreasi: @craig-mulligan  How about keeping the `must` for symmetry with the previous sentences and changing it to just `Local mode must be enabled`. The line right below, hints on a way to enable that through the dashboard.

zwalchuk: @craig-mulligan @thgreasi As far as I understand, you use local mode with the CLI, but it's still enabled via the dashboard. That could be incorrect though, if you know how to enable through the CLI please let me know because I couldn't find it in the CLI docs.

thgreasi: Technically you could use the API directly to enable local mode, or even set the respective env var through the CLI or the Dashboard. But since the officially documented way for now is to use the "Enable Local Mode" button from the Dashboard, this LGTM for now.

PR #605, opened 2017-10-16
Fix typo

lurch:


lurch:
Also, I'm not sure why there's a `artik710.md` in https://github.com/resin-io/docs/tree/master/shared/selectNetworkConfig ? (it also has the same typo)

PR #636, opened 2017-11-21
data-api: add example of getting device data by UUID

imrehg:
Important now that we are moving away from device IDs


imrehg:
Related to #637

PR #645, opened 2017-12-12
Update networking page, including proxy config with redsocks

zwalchuk:
A new section on the networking page with instructions for configuring redsocks for connecting to resin.io behind a proxy. A few other cleanups for the networking page are also included.


shared/networking/2.x.md
```diff
@@ -294,6 +297,16 @@ Additionally, if the network your device is connecting to works with whitelistin
 
 Additionally make an outgoing connection to `mixpanel.com`, but this is not a functional requirement for resin.io, but rather allows us to track some useful metrics.
 
+## Connecting Behind a Proxy
+
+We use **[redsocks][redsocks]** to connect your to device to resin.io from behind a proxy. To enable, a [redsocks.conf][redsocks-conf-example] file should be added at `/mnt/boot/system-proxy/`, specifying the needed proxy configuration and ensuring that:
+
+- `daemon = off` is set, so that **redsocks** doesn’t fork.
+- `local_port = 12345` is set, so that the iptables rules and **redsocks** port match.
+
```
imrehg: I wonder if a minimalist example would be helpful here, as that linked example is nice but super verbose: https://gist.github.com/imrehg/e720f7d5192a2248c0105639c7f9770b 
(but just asking, it would work for me as it is too)

imrehg:
Do you think that `navigation.txt` can be updated such that the two "Network Setup" pages show up different in the search? Right now both show with the same name, and it's weird to see (weird to choose). But if in the current infrastructure is not really feasible, then it's fine.

shared/networking/2.x.md
zwalchuk: Perfect, I'll add that in directly

PR #657, opened 2017-12-20
Update the alias name for RPI1 base images

nghiant2710:


lurch:
@nghiant2710 You need to add a `Change-type: ` footer to keep VB happy ;-)

pages/runtime/resin-base-images.md
```diff
@@ -84,7 +84,7 @@ These are base images for different architectures:
 
 | Image | Arch | Repository | Available Tag|
 |:-----------|:------------|:------------|:------------|:------------|
```
lurch: I think this needs to be changed to
```
|:-----------|:------------|:------------|:------------|
```
in order for GitHub to render this table correctly?

nghiant2710: @lurch it still renders 4 columns correctly even it should be 5. However, I've fixed this, thanks

Issue #652, opened 2017-12-18
Document Host SSH capability


zwalchuk:
Add here: https://docs.resin.io/runtime/runtime/#ssh-access

afitzek:
With the CLI release 6.12.0 we released direct host OS access for our customers.

To use it just add the `--host` or  `-s` option to the ssh command:

For example:

```
resin ssh 7cf02a6 -s
```

This will only work for Resin OS versions >= 2.7.5

The host OS can be used to check on the actual application container, supervisor logs, ...

afitzek:
This will only work when VPN is active and connected.

PR #660, opened 2017-12-20
Add host OS SSH info

zwalchuk:
Connects to #652 

Added a new section with information on using the CLI to access the host OS, including requirements and things to watch out for.

Change-type: patch


shared/general/host-ssh.md
```diff
@@ -0,0 +1,15 @@
...
+
+__Warning:__ Making changes to running services and network configurations carries the risk of losing access to your device. Before making changes to the host OS of a remote device, it is best to test locally. When in doubt, [reach out][forums] to us for guidance. 
+
+To use this option, add the `--host` or `-s` option to the `resin ssh` command:
+
+```shell
+$ resin ssh <device-uuid> -s
+```
+
+Host OS access is only available for devices running resinOS versions greater than 2.7.5. Like container SSH access, it requires the VPN to be active and connected.
```
afitzek: Just a small thing: It is available for resinOS version greater or equal to 2.7.5

shared/general/host-ssh.md
```diff
@@ -0,0 +1,15 @@
+### Accessing the host OS
+
+Starting with CLI release 6.12.0, it is possible to SSH into the host OS rather than the application container. This gives you access to logs and tools for services that operate outside the scope of your application, such as **NetworkManager**, **Docker**, and the supervisor.
```
agherzan: We should mention here VPN.

shared/general/host-ssh.md
```diff
@@ -0,0 +1,15 @@
...
+
+__Warning:__ Making changes to running services and network configurations carries the risk of losing access to your device. Before making changes to the host OS of a remote device, it is best to test locally. When in doubt, [reach out][forums] to us for guidance.
```
agherzan: We should mention that with host OS updates don't maintain changes over updates. And can even break the updater.

shared/general/host-ssh.md
alisondavis17: we should say "as well as the application container" since they can still SSH into the container as well, but now have the additional option of SSHing into the host

shared/general/host-ssh.md
alisondavis17: Very good point

alisondavis17:
Once the UI is ready, we should update these docs and add a screenshot, showing how users can SSH into the host from the web terminal (in addition to the CLI)

Issue #143, opened 2016-01-08
improve search functionality


shaunmulligan:
Search is a bit broken, perhaps try something like: https://community.algolia.com/docsearch/

zwalchuk:
fixed by #669

PR #669, opened 2017-12-22
Set up Algolia docsearch

pimterry:
Connects-To: #143
Change-Type: minor

This enables https://community.algolia.com/docsearch/ - algolia provide free autoconfigured searching for documentation sites.

They're already automatically indexing our site (I think daily?), and we just need to drop in this config, to get a much nicer search experience out of the box. Exact code to do that came direct from the email they sent me.

Comparison of the results:

screenshot from 2017-12-22 12-16-08

vs

screenshot from 2017-12-22 12-15-47

This is mostly a quick proof of concept. Limitations:
* I haven't removed the existing search code, which would break any links to search results, but we could look at doing that in future.
* Algolia are only searching our prod site, so this makes the search box on non-prod sites (local & staging) link to prod pages regardless. Not sure if that's a problem.
* I think there's a few places where we could restructure the docs to make this work more clearly (I think this is an issue with heading structure our end, e.g. lots of CLI results get listed under 'Table of Contents', rather than 'CLI documentation'), but I think we should ship & iterate on that later

Overall imo it's a clear improvement anyway, and a much nicer experience overall.
---- Autogenerated Waffleboard Connection: Connects to #143


pimterry:
I've also just found where our indexing configuration lives (by more successfully reading my emails). 

We can open PRs to change it here: https://github.com/algolia/docsearch-configs/blob/master/configs/resin_io.json

alisondavis17:
Very nice @pimterry and definitely looks an improvement

PR #685, opened 2018-01-22
Document tags

zwalchuk:
Add information on creating and editing tags, as well as managing tag columns.

Change-type: patch


pages/management/applications.md
```diff
@@ -66,6 +66,40 @@ When you create a view, you are saving a specific set of filters that you may wa
 
 You will be asked to choose a name for the saved view. This view can then be selected at any time by clicking the *Views* button above the filter box.
 
+## Device tags
```
alisondavis17: we could explain that tags are created and stored as key:value pairs

pages/management/applications.md
```diff
@@ -66,6 +66,40 @@ When you create a view, you are saving a specific set of filters that you may wa
...
+
+Tags provide an easy way to include more information about your devices than what is already provided in the dashboard. A device can be given multiple tags, and the tags can be used when you create filters.
```
alisondavis17: it may be worth adding an example of this to the section of the docs on filters, since the combination of tags + filters is so powerful

pages/management/applications.md
```diff
@@ -66,6 +66,40 @@ When you create a view, you are saving a specific set of filters that you may wa
...
+
+To add a tag from the device list, click the checkbox to the left of the devices you wish to tag, followed by the *Tags* button on the right side of the dashboard:
+
+<img src="/img/common/app/tags.png" width="80%">
+
+This opens the tag management window. From here you can define a name for your tag and a value for this specific device. For example, you could create a tag to keep track of device location:
+
+<img src="/img/common/app/add_tag.png" width="80%">
+
+Tags you add to your device will appear in a list in the management window. They can be edited or removed from here:
```
alisondavis17: Tags can also be managed (added / edited / removed) from each individual device page - we should add docs on that as well. See:

pages/management/applications.md
```diff
@@ -66,6 +66,40 @@ When you create a view, you are saving a specific set of filters that you may wa
...
+
+<img src="/img/common/app/tag_list.png" width="80%">
+
+When you are finished creating and editing tags, click *Apply* to keep the changes.
```
thgreasi: How about `persist` instead of `keep`?
It might worth mentioning the `Add Tag` button as a reminder.

pages/management/applications.md
thgreasi: and perhaps that the name/key is unique and that the value is optional.

Also, the [spec describes](https://github.com/resin-io/hq/pull/281/files#diff-855f6128af9496112f322bd42abfe5ddR29) that we should promote using the `reverse domain notation` for tag keys/names. Even though not mandatory or enforced with any way, it can be really useful since it leads to structured tag names. I wonder whether this fits in here. Maybe not.

zwalchuk: @thgreasi I think that will be most useful to have in the API documentation around tags, but I'm not sure. Could you provide an example where using that structure would be helpful?

thgreasi: You mean the `reverse domain notation`?
For example you can have several `stats` and `status` related tags and keep them organized:
* stats.boot.lastbootdate
* stats.boot.lastbootduration
* status.disk.sizeavailable
* status.myappspecificstate (eg recording video)

On second thought, maybe it's too much to have it in this page. I'm fine with this as is.

Issue #692, opened 2018-01-29
List the latest available self-service hostOS update device types


imrehg:
Here are the HQ issues for devices, and the ticks of which are in production

* [x] [IOT2000](https://github.com/resin-io/hq/issues/1188)
* [x] [TS4900](https://github.com/resin-io/hq/issues/1121)
* [x] [TX2 / SKX2](https://github.com/resin-io/hq/issues/1172)
* [x] [QEMU 32/64bit](https://github.com/resin-io/hq/issues/1175)

When all are in production, will add the info to the docs.

imrehg:
Deployed

PR #707, opened 2018-02-09
Updating resinOS: adding newly supported devices to the documentation

imrehg:
Relates to #692


imrehg:
Thanks @zwalchuk, I've squashed these two commits, and will merge when the tests are finished.

PR #710, opened 2018-02-09
Update delta info

### LINKED ISSUE #701: Update delta behaviour description
Update the delta update behaviour description in https://github.com/resin-io/docs/blob/master/pages/runtime/delta.md – for details, see https://app.frontapp.com/open/cnv_k3gwo1

See also: https://github.com/resin-io/docs/issues/615

Front logo [Front conversations](https://app.frontapp.com/open/top_7wfl)

zwalchuk:
Include information about the resumable nature of delta updates, including a description of the applicable config variables.

Change-type: patch

Fixes #701


pages/runtime/delta.md
```diff
@@ -26,11 +26,15 @@ To enable this behaviour on a per-device basis, set the above variable at **Devi
...
 
-You might still observe the application **downloading the full application image**. This happens if there was an issue during the delta download, and in that case the supervisor will fall back to the original Docker pull approach to recover.
+Delta updates are resumable, so if the connection drops or otherwise stalls, the update will resume from the last byte received. A user can configure the number of times a delta is retried before it bails out and signals failure to the supervisor. This is set by the `RESIN_SUPERVISOR_DELTA_RETRY_COUNT` configuration variable.
+
+In addition, the wait time before an update is considered stalled and the time between retries are configurable, using `RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT` and `RESIN_SUPERVISOR_DELTA_RETRY_INTERVAL`. These are both specified in milliseconds. The defaults for these options allow for a 20 minute time frame where no byte is received before they give up, after which the supervisor retries the delta from the beginning. Even then, the actual bandwidth consumed is usually less than pulling layers.
```
dfunckt: > where no byte is received before they give up, ...

I think this reads better as...

> where no byte is received before giving up, ...

or as...

> where no byte is received before deltas give up, ...

dfunckt: > the wait time before an update is considered stalled

In the paragraph above we already talk about the connection stalling, so I think it’s fine to refer to it again here, to preserve context:

> the wait time before a connection is considered stalled

dfunckt: > Even then, the actual bandwidth consumed is usually less than pulling layers.

It might make sense to drop this given the previous paragraph where we already describe why deltas are smaller.

> As described above, even then, the actual bandwidth consumed is usually significantly less than pulling the new image.

pages/runtime/delta.md
```diff
@@ -26,11 +26,15 @@ To enable this behaviour on a per-device basis, set the above variable at **Devi
...
+
+The only case where the deltas fall back to pulling the entire layer is if corruption is detected.
```
dfunckt: Technically, this is more correct, given an image is potentially many layers:

> The only case where the deltas fall back to pulling the entire new image is if corruption is detected.

dfunckt:
There’s a link to the blog at the ebd of this document that includes a video that no longer exists. We might want to either remove the link or fix the video.

https://resin.io/blog/the-next-resin/#deltaupdates

Issue #696, opened 2018-02-01
Add TX1 to Jetson getting started guide


zwalchuk:
Update the current TX2 starting guide to include support for the TX1

zwalchuk:
@lurch correct

PR #713, opened 2018-02-09
Add getting started for Jetson TX1

### LINKED ISSUE #686: Document that first time flashing TX2
For first time flashing the Jetson TX2, might need to flash the device first with the latest JetPack from Nvidia https://developer.nvidia.com/embedded/jetpack After it was flashed (the bootloader update is the important part for us), the resinOS flashing process will work.

Not sure if it is always needed, but probably yes, as likely the boards ship with old version from Nvidia.

cc @telphan @shaunmulligan

zwalchuk:
New getting started guide for the TX1. Also update the flashing instructions to include the step of preparing with the latest version of JetPack.

Change-type: patch

Fixes #686 and #696


zwalchuk:
@imrehg yes, I think you are correct. Won't release this yet. I was living in the future

zwalchuk:
@imrehg TX1 is released now, could I get another review?

zwalchuk:
Added in #821

PR #718, opened 2018-02-12
Fix the improve this doc button on dynamic pages

### LINKED ISSUE #563: "Improve this Doc" links don't work on parametric pages
The URL generated on the "Improve this Docs" links is incorrect for any of the parametric pages. The links created for those buttons preface the file with the DOCS_SOURCE_DIR ('pages'), but for the pages that are selected with the dropdown, 'pages' isn't used (for example https://docs.resin.io/deployment/network/2.0.0/)

zwalchuk:
Previously, the "Improve this doc" button failed for pages that are generated dynamically, as the page does not exist anywhere as a single markdown file that can be edited. Some logic has been added so that any dynamic pages just link to the shared folder, which contains any markdown files that would need to be edited
to fix the page.

To make this change work, the Network and resinOS page links have been updated to match that of the getting started guides, for example "2.x/understanding/understanding-devices/" instead of "understanding/understanding-devices/2.x". The redirects have been updated to handle links in the old format.

Fixes #563


config/doxx.coffee
```diff
@@ -13,7 +13,8 @@ module.exports = {
   templatesDir: config.templatesDir
   partialsDir: config.partialsDir
   metaExtra: (fileName) ->
-    improveDocsLink: "#{config.editPageLink}/#{config.docsSourceDir}/#{fileName}"
+    [..., newName] = fileName.split "/"
```
pimterry: Nit: the convention normally is brackets for single-line calls, omitted for multi-line calls, and single quotes for strings, with `"` only for strings using interpolation (`#{}`). Docs here: https://github.com/resin-io/process/blob/master/process/coffeescript-coding-standards.md. That means this would be `fileName.split('/')`

config/index.coffee
```diff
@@ -4,6 +4,11 @@ DOCS_DEST_DIR = 'contents'
 TEMPLATES_DIR = 'templates'
 PARTIALS_DIR = 'shared'
 
+DYNAMIC_DOCS = [ 'getting-started.md'
+                  'network.md'
+                  'understanding-devices.md'
+]
```
pimterry: Formatting's a bit weird here too. I'd normally write this as:

```coffeescript
DYNAMIC_DOCS = [
    'getting-started.md'
    ...
]
```

config/doxx.coffee
pimterry: What does `new` mean in this context? `lastPathSegment` might be clearer (or something like that)

config/redirects.txt
```diff
@@ -8,19 +8,21 @@
...
 
 # Default redirects
-^\/getting-started -> /raspberrypi3/nodejs/getting-started
-^\/deployment\/network\/(2\.0.*|$) -> /deployment/network/2.x
-^\/deployment\/network\/1.x.x -> /deployment/network/1.x
-^\/understanding/understanding-devices\/(2\.0.*|$) -> /understanding/understanding-devices/2.x
-^\/understanding/understanding-devices\/1.x.x -> /understanding/undderstanding-devices/1.x
+^\/getting-started($) -> /raspberrypi3/nodejs/getting-started/
+^\/deployment\/network\/(2\.0.*|$) -> /2.x/deployment/network
+^\/deployment\/network\/1.x.x -> /1.x/deployment/network
```
pimterry: Is this right? The URLs I see for this right now are `https://docs.resin.io/deployment/network/1.x/` and `https://docs.resin.io/deployment/network/2.x/`, and this doesn't match either. I suspect the same is true below too.

config/redirects.txt
```diff
@@ -8,19 +8,21 @@
...
+^\/understanding/understanding-devices\/(2\.0.*|$) -> /2.x/understanding/understanding-devices
+^\/understanding/understanding-devices\/1.x.x -> /1.x/understanding/understanding-devices
+^\/deployment\/network\/([1-2]\.x) -> /$1/deployment/network 
+^\/understanding\/understanding-devices\/([1-2]\.x) -> /$1/understanding/understanding-devices
```
pimterry: Is the main motivation for this that these URLs are easier to match as dynamic? I can see how that's true, but we could still handle this fairly easy in other ways (a list of dynamic URL regexes), and it feels like it makes a little more sense to me with the parameter at the end really.

zwalchuk: @pimterry yes, that was the main motivation. I had originally switched things the other way (so that the getting started links matched the network and resinOS links) as I agree this make more sense to have the parameter at the end. I was having trouble getting it to recognize the dynamic pages in a way that wasn't too hacky, but I have a couple of other ideas I'll try.

config/redirects.txt
zwalchuk: @pimterry that's an older redirect for an even earlier version of the pages that had the versions as 2.0.0 and 1.x.x.

config/redirects.txt
```diff
@@ -8,19 +8,21 @@
...
+^\/installing\/gettingStarted(\/|$) -> /getting-started/raspberrypi3/nodejs/
+^\/installing\/gettingStarted-BBB(\/|$) -> /getting-started/beaglebone/nodejs/
+^\/installing\/gettingStarted-Humming(\/|$) -> /getting-started/hummingboard/nodejs/
+^\/installing\/gettingStarted-NUC(\/|$) -> /getting-started/nuc/nodejs/
+^\/installing\/gettingStarted-Nitrogen6x(\/|$) -> /getting-started/nitrogen6x/nodejs/
 
 
 # Default redirects
-^\/getting-started -> /raspberrypi3/nodejs/getting-started
+^\/getting-started($) -> /getting-started/raspberrypi3/nodejs/
```
pimterry: Nit: I don't think you need the brackets here

PR #752, opened 2018-03-16
Update persistent storage info

zwalchuk:
Add more info on the new location of persistent storage as a result of using named volumes.

Change-type: patch


pages/learn/develop/multicontainer.md
```diff
@@ -66,7 +66,7 @@ restart: always
 
 Setting `network_mode` to `host` allows the container to share the same network namespace as the host OS. When this is set, any ports exposed on the container will be exposed locally on the device. This is necessary for features such as bluetooth.
 
-To store data in [persistent storage][persistent-storage], you'll want to make sure to use the `volumes` field to link a directory in your container to the `resin-data` volume. Your named volume should be specified at the top-level of the `docker-compose.yml`, with the container path defined in the service:
+To store data in [persistent storage][persistent-storage], use the `volumes` field of the service to link the `/data` directory in your container to the `resin-data` named volume. The `resin-data` named volume should also be specified at the top-level of the `docker-compose.yml`:
```
pcarranzav: It would be good to explain that arbitrary named volumes can be created. Also, note that tge volumes will be persisted as long as there’s a volume with that name in every release of the application, so if you change the name of a volume, the data will be lost.

shared/general/persistent-storage.md
```diff
@@ -1,10 +1,12 @@
...
+If you have data or configurations that you want to persist through the update process, you can store them in the `/data` folder. For applications with a single container, this folder is automatically linked to a directory on the host OS, and it is guaranteed to be maintained when new versions of the application or the host OS are applied (for multicontainer applications, you'll want to [configure this link][multicontainer] as appropriate). The `/data` folder is a good place to write system logs and other application data that should be available across updates and services.
 
-This folder is guaranteed to be maintained across updates and thus files contained in it can act as persistent storage.	This is a good place to write system logs, etc.
+Note that the `/data` folder is __not__ mounted when your project is building on our build servers, so you can't access it from your `Dockerfile`. The `/data` volume only exists when the container is running on the deployed devices.
```
pcarranzav: We can add, though, that any data added in /data during the build will be copied into the volume the first time it’s created on the device (which is the default behavior for docker volumes) - for devices on resinOS 2.12 and higher.

PR #754, opened 2018-03-19
Update ARTIK 5 and 10 to 520 and 1020 names

zwalchuk:
The ARTIK 5 and 10 are now called 520 and 1020, and docs have been updated to reflect this. Device type names are still artik5 and artik10, only the display name has been changed.

Change-type: patch


floion:
@zwalchuk just a heads-up that we will change the device name to artik520 too

zwalchuk:
@floion sounds good, I'll merge this for now as it will address the immediate issue, and I'll update the device name once all the base images are updated, as those are the main thing the device name is used for in the docs

PR #757, opened 2018-03-20
Update script for external pages

zwalchuk:
Minor fixes to the script for external pages.

Change-type: patch


pages/reference/cli.md
```diff
@@ -81,76 +113,76 @@ environment variable (in the same standard URL format).
...
+	- [deploy &#60;appName&#62; [image]](#deploy-appname--image)
 
 - Utilities
 
 	- [util available-drives](#util-available-drives)
 
 # Application
 
-## app create &#60;name&#62;
+## <a name="#app-create-name"></a>app create &#60;name&#62;
```
pimterry: Looks like this is built from the previous CLI docs, not the latest, is that right?

zwalchuk: @pimterry yes, I'll rebase this with the latest changes.

pimterry:
@zwalchuk looks like this now doesn't have any of the CLI docs changes at all - is that right?

zwalchuk:
@pimterry right again, I thought I had the new changes in master already but it was still the one from a month ago. Should be all correct now.

PR #764, opened 2018-03-22
Fix note on purging data under move device action

zwalchuk:
Currently, this only describes behavior for devices before resinOS v2.12.0.

Change-type: patch


pages/learn/manage/actions.md
```diff
@@ -75,7 +75,7 @@ Note that currently you are only able to move devices between applications of th
...
 
-__Note:__ When moving between applications, the data stored in `/data` will persist, but will not be accessible to the new application. However, upon switching back to the original application, the `/data` store will be available once more. It is recommended that if one does not intend on reverting back to the original application, then a purge of `/data` is needed.
+__Note:__ For devices running resinOS version 2.12.0 and above, data in [persistent storage][persistent-storage] (named volumes) is automatically purged when a device is moved to a new application. On older host OS versions, the `/data` folder in the new application will not contain any of the old application data, but it can still be accessed via the host OS and if the device is switched back to the original application. Unless you plan to revert back to the original application, be sure to [purge][purge-data] the `/data` folder.
```
Page-: I think it would make more sense as a warning since it involves deleting data that would previously be preserved

PR #772, opened 2018-03-30
Updates for device (environment) variables

zwalchuk:
The service variables page needs to be updated for environment variables, which apply to all services on a device.

This will updated further as the UI is refined, but this should accurately describe the way things currently work.

Change-type: patch


pages/learn/manage/serv-vars.md
```diff
@@ -1,48 +1,52 @@
...
 
-__Note:__ Adding or modifying a service variable will cause the service to restart on any devices where the new setting is applied.
+__Note:__ Adding or modifying an environment variable will restart all services on a device. Service variables only restart the affected service.
 
-Resin.io allows you to set service variables at both the fleet and device level.
+Environment variables are accessible to all services running on a device, whereas service variables are assigned to a specific service. Both environment and service variables can be applied at the fleet or the individual device level. 
 
-__Note:__ Service variables defined in the dashboard will not apply to devices in [local mode][local-mode]. You will need to define them in your `resin-sync.yml`.
+Values defined for individual devices always override those defined for the fleet, and values defined for services generally override those defined for a device. So for any given variable, the **device service variable** will have top priority, followed by the **device environment variable**, then the **fleet service variable**, and finally the **fleet environment variable**.
```
thgreasi: > and values defined for services generally override those defined for a device.

This might need some tuning to describe the that service vars are more specific for both the app&device association group. A programmer minded description I could think of would look like:
> and values defined for the service level override those defined for the respective application or device level.

I'm sure that this more confusing, but you get my point.

pages/learn/manage/serv-vars.md
```diff
@@ -1,48 +1,52 @@
...
 
-## Device service variables
+## Device environment and service variables
 
-Device service variables are available to the service running on that particular device. If both the fleet and the device have a service variable of the same name, the code on the device will use the value of the device service variable. In other words, device service variables redefine (or override) fleet service variables.
+Device environment and service variables are applied to only one device. Device service variables override device environment variables of the same name, and all device variables override fleet variables.
 
-Adding a device service variable is very similar to adding a fleet service variable. From the device summary page, select the *Device Service Variables* tab, click *Add variable*, select the appropriate service from the drop down, add a name and value, and click *Add*.
+Adding a device variable is very similar to adding a fleet variable. From the device summary page, select *Device Variables* for environment variables or *Device Service Variables* for service variables. Click *Add variable*, select the appropriate service if necessary, add a name and value, and click *Add*.
```
thgreasi: > add a name and value

should this be
> add a name and a value

?

pages/learn/manage/serv-vars.md
zwalchuk: @thgreasi yeah, it's a tricky statement to get correct, as fleet service variables will override fleet environment variables, but not device environment variables. I'll try another approach, because I agree it's not clear as is.

PR #780, opened 2018-04-07
Fix location of multicontainer project example

alexose:


zwalchuk:
@alexose thanks for pointing those out! I've included your changes in #777, which has just been deployed to production.

PR #783, opened 2018-04-09
Update TX2 flashing instructions

### LINKED ISSUE #686: Document that first time flashing TX2

zwalchuk:
Include the instructions on flashing with JetPack before resinOS.

Fixes #686

Change-type: patch


shared/getDeviceOnDash/jetson-tx2.md
```diff
@@ -1,4 +1,6 @@
-Make sure your **{{ $device.name }}** development board is powered off. Insert the SD card from the previous step and plug in the power cord. Press and hold the power button for one second to turn the device on. 
+Before you can flash the **{{ $device.name }}** with resinOS, you'll need to prepare it with the latest [JetPack](https://developer.nvidia.com/embedded/jetpack) using the instructions provided by [Nvidia](http://docs.nvidia.com/jetpack-l4t/3.1/index.html#developertools/mobile/jetpack/l4t/3.1/jetpack_l4t_install.htm).
```
floion: @telphan so is it latest that needs to be flashed or a certain version?

PR #784, opened 2018-04-13
Added getting started for Haskell

nazrhom:


zwalchuk:
@nazrhom The icons are controlled by height, so the Haskell one just sticks out a little further because it is longer than it is tall. It looks fine to me

PR #798, opened 2018-05-11
Improve and update API documentation

zwalchuk:
A new structure for the API docs, with two pages: one overview page including authentication info and a tutorial on constructing API calls, and a resources page which includes available fields and example calls for all resources in the data model . This page is generated dynamically from a json dictionary, making it trivial to add new resources/examples.

Change-type: patch


shared/apiGeneral/calls.md
```diff
@@ -0,0 +1,71 @@
...
+A final tip for constructing API calls: for some of the fields in the API response, a link to another resource is provided rather than the complete information about that resource. For example, if you make a call requesting information about a specific device, the `belongs_to__application` field will return a link to an application, but not all the information about that application. To get all the fields for the application resource, you can use the `$expand` method:
+
+```
+curl -X GET \
+"{{ $links.apiBase }}/v4/device?\$filter=uuid%20eq%20'<UUID>'&\$expand=belongs_to__application" \
+-H "Content-Type: application/json" \
+-H "Authorization: Bearer <AUTH_TOKEN>"
+```
+
+It's also possible to filter on a field that belongs to la inked resource. To find all devices belonging to an application by that application's name, you would construct your query like this:
```
alisondavis17: think you've got a small typo here ("la linked")

shared/apiGeneral/introduction.md
```diff
@@ -0,0 +1,9 @@
+## Introduction
+
+The API is the core of the resin.io platform. It provides a secure channel for communication between resin.io services and the database. The API's HTTP interface not only informs the dashboard and the CLI, it also gives you the power to directly access resources associated with your account. With the API, you can fetch and update information about your applications, devices, environment variables and more.
+
+This guide is split into two parts. On this page, you will find a basic tutorial to help you construct API calls. The [Resources][resources] page provides more details about the resources that can be queried using the API, including example calls and a list of available fields.
+
+__Warning__: When using the API to make changes, take great care in selecting the appropriate resources, as there are *no checks* to prevent you from accidentally making widespread, irreversible mistakes. Test filters with a `GET` call *before* you use them in a `PATCH` or `DELETE` request.
```
alisondavis17: Would we also want to suggest that people consider creating a backup? We'd had people ask us to restore from our own backups when their API `PATCH` etc went wrong (was used without a filter), which is a huge hassle :/

lekkas:
Some fields (e.g. `user` in /reference/api/resources/application/, `source` in /reference/api/resources/release/) have a different colour, does this denote that these fields are special in any way?

shaunmulligan:
@zwalchuk this all looks great. One thing I think that is missing is some explanation about the API versioning and how older calls to say `v3` etc will be translated in the backed to the equivalent newer API calls. I think for users looking for say `V2` api documentation, they might be a bit frustrated to not find it anywhere and think that the v4 docs don't apply

shared/apiGeneral/introduction.md
zwalchuk: I'm not sure if the API is something they could backup in any straightforward way. Are you thinking maybe copy the current state of a resource before they make a change to it?

Issue #805, opened 2018-05-23
Setting environment variables from CLI


martenjacobs:
I'd like to add a number of service- and environment variables to my application from the CLI. It seems the current version of the documentation (or even the cli itself?) does not work properly with multicontainer set-ups. Using the docs and the current version of the cli, I'm able to set service variables for only one of my microservices and I'm completely unable to set environment variables.

pimterry:
@martenjacobs yes, for multicontainer apps the CLI can't yet configure every kind of variable that you'll want. We're working on that, but for now for any features not yet fully supported you'll need to set those through the API directly.

See the [API overview](https://docs.resin.io/reference/api/overview/) for an introduction, and the [application config](https://docs.resin.io/reference/api/resources/application_config_variable/), [application environment](https://docs.resin.io/reference/api/resources/application_environment_variable/), and [service environment](https://docs.resin.io/reference/api/resources/service_environment_variable/) variable resources for details & examples.

PR #819, opened 2018-05-30
Added message for envs in multi-container set-up

martenjacobs:
Control of environment variables through the CLI has some issues in multi-container set-ups, see #805

### DIFF
--- a/pages/reference/cli.md
+++ b/pages/reference/cli.md
@@ -553,6 +553,8 @@ This command lists all custom environment variables.
 If you want to see all environment variables, including private
 ones used by resin, use the verbose option.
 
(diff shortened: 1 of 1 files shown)


pimterry:
Hi @martenjacobs! Thanks for the submission, this is a good idea :+1:

The CLI docs in this repo are actually autogenerated from the CLI codebase itself though, so this isn't the right place for this change. I've made a PR to the CLI itself with the equivalent update over here: https://github.com/resin-io/resin-cli/pull/887.

Once that's merged & released the docs here will pick up the changes next time they're updated, so I'm going to close this PR now, and finish this in the CLI PR. Thanks for contributing!

PR #823, opened 2018-06-04
Attempt to build at runtime to pull in env vars

brownjohnf:


brownjohnf:
As you can see, we basically need to change everywhere that uses configuration to read that config from ENV vars. I just changed a few places to demonstrate the concept; you'll need to update everywhere else accordingly. Given how many configs the docs have, I'd suggest trying to consolidate things like upper/lower/capitalized versions of the name to generate them programatically.

PR #831, opened 2018-06-19
Remove buildpack mentions from base image docs

zwalchuk:
Change-type: patch


pages/reference/base-images/resin-base-images.md
```diff
@@ -101,10 +98,6 @@ These are base images for different architectures:
 | `resin/i386-alpine` | i386 | [Docker Hub][i386-alpine-dockerhub-link], [github][i386-alpine-github-link] | latest, 3.7, 3.6, 3.5, edge |
 | `resin/armv7hf-fedora` | armv7hf | [Docker Hub][armv7hf-fedora-dockerhub-link], [github][armv7hf-fedora-github-link] | latest, 26, 25, 24 |
 
-![Hierarchy Diagram](/img/hierarchy-diagram.jpg)
```
jviotti: This one is a bit tricky. Maybe we should update this diagram to make the `<distro>-<stack>` images appear to inherit straight from `<device-name>-<distro>`? 

Alternatively, if we're not going to use this image, should we delete it from `static/img`?

zwalchuk: @jviotti It might make sense to update, @nghiant2710 do you have the original?

nghiant2710: @jviotti @zwalchuk, I think it's better to remove the image since updating it will not illustrate correctly how we build the images. We can add new one for balena base images later.

PR #834, opened 2018-06-24
local-mode: <short-uuid>.local push does not work

chibby0ne:
The usage of `<short-uuid>.local` in the push command does not work and results
in a error of the type:

```
Push failed. Error: Error while inspecting image local-app: Error: getaddrinfo ENOTFOUND
```

Instead using the IP address or just the `<short-uuid>` works as expected.

Change-type: patch

### DIFF
--- a/pages/learn/develop/local-mode.md
+++ b/pages/learn/develop/local-mode.md
@@ -54,11 +54,11 @@ Reporting scan results
 
 ## Push over a new project
 
-Now that we know where our device is on the network we can start pushing some code to it. To do this, we use the `resin local push` command. This command instructs the device to do a Docker build and then runs your container in the same configuration as the resin.io device supervisor supervisor would. You can either pass the command your device's IP address or `<short-uuid>.local` name. If you leave this out, you will be presented with a list of devices to choose from.
+Now that we know where our device is on the network we can start pushing some code to it. To do this, we use the `resin local push` command. This command instructs the device to do a Docker build and then runs your container in the same configuration as the resin.io device supervisor supervisor would. You can either pass the command your device's IP address or `<short-uuid>` name. If you leave this out, you will be presented with a list of devices to choose from.
 
 **Command**
 ```
-sudo resin local push f340127.local -s .
+sudo resin local push f340127 -s .
 ```
 **Output**
 ```


chibby0ne:
It would be great if this is merged ASAP, since having the first push command fail after following the instructions to the letter on the usage of the `resin-cli` is very discouraging. 
Although the text also mentions that it can handle the use of IP addresses, most people will copy and paste and edit the example command, and when they see the error they would most likely think the tool is broken or their installation of the tool is broken, and either way wasting time finding fixes or worst yet give up on an falsely assumed broken tool.

chibby0ne:
I'm not sure, but given that the `hostname` is just the short UUID when checked directly in the devices and using it without any .local works just fine, I would be inclined to think that `.local` is just a suffix appended in the resin-cli when listing the scanned devices and not something coming from the network.

pimterry:
@chibby0ne Can we have some more info about your system? `.local` isn't something that's appended by the CLI, it's a standard suffix used for mdns hostnames (https://en.wikipedia.org/wiki/.local), which is how the CLI discovers devices like these on the network. On my machine local commands do require a `.local` suffix, and don't work without, so we can't simply update the instructions like this.

We should resolve this though, because I'm sure you're not the only user in this situation. I'd like to reproduce this, and work out what's causing the difference. Can you let us know:

* Your OS & version
* Your Resin CLI version
* The Resin OS version of the device you're trying to connect to
* Whether there's anything unusual about your network
* Whether you can reproduce this on all networks

lurch:
Not directly-related to the problem here, but `.local` addresses are also used e.g. by https://www.raspberrypi.org/blog/google-coder/ and https://blog.gbaman.info/?p=791  (just in case those articles give some good debugging-pointers)

floion:
@pimterry this may be related and a possible fix: https://github.com/resin-os/meta-resin/issues/1083

pimterry:
@floion I think that issue should only affect DNS lookups on the device itself, right? These lookups are happening on @chibby0ne's development machine, so I _think_ shouldn't be affected.

It's possible there's some more complicated interaction there though of course, with that issue somehow affected published MDNS too... Is there a known affected version range for that we could use to check this?

pimterry:
@chibby0ne any updates on this? I'd love to look into it, but I can't without the details to reproduce it.

pimterry:
@chibby0ne I'm going to close this for now. I'm happy to look into it further, but generally the `<uuid>.local` format should be correct, so we'll need some more information from you to reproduce the issue and work out why it's failing in your case.

Feel free to ping here and we'll investigate further later on if you (or anybody else) can give us the details to reproduce this.

Issue #833, opened 2018-06-22
Fix get all devices for an application example


zwalchuk:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_ccwz)

vipulgupta2048:
Closing issue in favor of #868

PR #868, opened 2018-08-08
fix: Update name field in Device resource example of API

### LINKED ISSUE #865: Mistake in device renaming
Hey,
while moving from API v1 to v4, I noticed that the device renaming doesn't work as documented. Seems like the new variable name is `device_name` instead of `name` (at least it works this way ;) ).

So changing https://github.com/resin-io/docs/blob/51edbdf855c7715644dd52cdd3ea915c51d30379/config/dictionaries/resource.json#L188 to use `device_name` in the `PATCH` request should be used instead `name`.

zwalchuk:
The rename device example has the field listed as name, and it needs to
be device_name. The get all devices by application example also needs to be updated

Fixes #865 and #833

Change-type: patch
Signed-off-by: Zach Walchuk <zach@resin.io>


chrisys:
Rewrite these changes into a new PR

chrisys:
@shaunmulligan this went through OK, I guess because the source branch was actually in the repo

PR #876, opened 2018-08-11
Added note linking to issue w/ setting hostname via DBus

dt-rush:


dt-rush:
I didn't have `Change-Type` in my commit message, so I'll open a new PR with a proper commit.

PR #877, opened 2018-08-12
Added note linking to issue w/ setting hostname via DBus in Resin 2.0

dt-rush:
Change-Type: patch


pcarranzav:
I think it would probably be better to just remove the old instructions, and instead point to the supervisor API endpoint that allows changing the hostname: https://github.com/resin-io/resin-supervisor/blob/master/docs/API.md#patch-v1devicehost-config (this would also close the linked issue).

chrisys:
Close this and add the note recommended by Pablo about changing hostname via supervisor API

chrisys:
Superseded in #969

PR #879, opened 2018-08-13
fix: Revert Procfile

zwalchuk:
The Procfile instructions were recently updated to include npm build in
addition to npm start, this may be causing problems with the app
crashing on deploy.

Change-type: patch
Signed-off-by: Zach Walchuk <zach@resin.io>


zwalchuk:
retest

PR #892, opened 2018-09-10
advanced: Improve docs on modifying config.txt

### LINKED ISSUE #882: Document how to set multiple dtoverlay values
It is possible to use a comma separated list as so:

`RESIN_HOST_CONFIG_dtoverlay ` = `"balena-fin","rpi-sense"`

Front logo [Front conversations](https://app.frontapp.com/open/top_dmuj)

pcarranzav:
We fix the reference to config.txt having to be modified
locally after first boot  - it's actually the other way round,
since the supervisor will apply whatever the API says it 
should after first boot (and only what was there before first boot will be kept).

We also expand on how to set dtoverlay and dtparam values,
with the new way to set several values as an array.

Closes #882 
Change-type: patch
Signed-off-by: Pablo Carranza Velez <pablo@resin.io>


pages/reference/OS/advanced.md
```diff
@@ -88,6 +86,33 @@ To enable the miniUART an entry should be added to `config.txt` as follows:
 enable_uart=1
 ```
 
+### Setting device tree overlays (`dtoverlay`) and parameters (`dtparam`)
+
+The Raspberry Pi allows loading custom device tree overlays using the `dtoverlay` setting in config.txt. It also allows setting parameters for the default overlay with the `dtparam` setting. For these settings, the syntax is different from other keys because several entries can be added, and the bootloader will use all of them.
+
+To allow setting several values, devices running resinOS version 2.12.0 and above (supervisor 7.0.0 and above), will parse the values of `RESIN_HOST_CONFIG_dtoverlay` and `RESIN_HOST_CONFIG_dtparam` in a special way: the value of the configuration variable will be treated as the contents of a JSON array (without the enclosing braces `[]`), so a comma-separated list of quote-enclosed values will be split into several lines.
```
CameronDiver: The "treated as the contents of a JSON array (without the enclosing braces `[]`)" reads a little weird, maybe specific comma-seperated values first, and touch on it's parsed as a JSON array as an afterthought kind of thing?

Keen for somebody more versed in documentation writing to weigh in though :)

PR #893, opened 2018-09-11
Remove information about dockerignore files

CameronDiver:
As far as I'm aware this has never been possible while I've been at resin. Either way I've removed it here, but I will make an issue on the builder to tackle this (and we'll certainly need to once the support has been released on the CLI)

Builder issue: https://github.com/resin-io/resin-builder/issues/542

Change-type: patch
Signed-off-by: Cameron Diver <cameron@resin.io>


shaunmulligan:
@CameronDiver is this PR still valid? I know we had added some handling of .dockerignore in the CLI recently.

CameronDiver:
@shaunmulligan it's now supported with `balena push` yeah.

PR #899, opened 2018-09-13
Fix api docs to rename device to point to the right field

ZubairLK:
Fixes https://github.com/resin-io/docs/issues/898

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -186,7 +186,7 @@
         "method": "PATCH",
         "endpoint": "/v4/device(<ID>)",
         "filters": "",
-        "data": "{\n    \"name\": \"<NEW NAME>\"\n}"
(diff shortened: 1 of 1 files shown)


ZubairLK:
@Page- can you check the docs (https://docs.resin.io/reference/api/resources/device/) to rename a device. I think there is a mistake in them as the customer pointed out in https://app.frontapp.com/open/cnv_1089fab

I've made a PR with this commit via github.
But I'm not 100% sure if this is the right place for this update..

Page-:
@ZubairLK `device_name` is correct but only for v4, older versions do use name, cc @CameronDiver was this change intentional as part of mc?

CameronDiver:
@ZubairLK @Page- IIRC it was to do with env var changes, and `device has name` was colliding with one the new env vars.

chrisys:
Closed in favour of #868

PR #903, opened 2018-09-14
Remove stale information from network requirements page

lekkas:
Closes https://github.com/resin-io/docs/issues/814

Front logo [Front conversations](https://app.frontapp.com/open/top_dvp7)


lekkas:
@hedss @shaunmulligan do you think we should provide a list of the current resin.io domains here? Also, any objections to move the detailed networking requirements for NTP from https://docs.resin.io/reference/OS/time/#networking-requirements here ?

imrehg:
@lekkas may I add some more things here?
Basically need to add `*.docker.com` and `*.docker.io` (for the OS/supervisor updates, that we are pulling from docker hub). Will try to narrow it down (what subdomains docker pull uses, but there's a number and likely can change)  but I think that's the current good thing to add....

I think pubnub is still required for devices before supervisor 7.16.x and the ones that are not yet migrated (many device types still don't have that one released) But this change is good aspirational, should probably keep it! Mentioning for clarity.

imrehg:
and yeah, NTP is important, UDP port 123, right? Those one we don't have specific domains as we use a pool :/ But they can set their own, yey!

shared/network/2.x.md
```diff
@@ -329,11 +329,8 @@ In order for a {{ $names.company.lower }} device to get outside of the local net
 
 Each of these should work with outward only (and inward once outward connection established) firewall settings.
 
-Additionally, if the network your device is connecting to works with whitelisting, you should whitelist the following domains on Port `80` and `443`:
+Additionally, if the network your device is connecting to works with whitelisting, you should whitelist the following domains on `443`:
```
hedss: Misread this as public URL port.

I think this needs to say 'you should whitelist the following domains for the relevant ports above:`.

As long as the fields include `pool.ntp.org`, `resin.io`, `docker.com`, etc. then they're good (but with the right ports for the right domains).

chrisys:
Close and reincorporate changes into new PR

PR #915, opened 2018-10-01
Fix formatting of examples for resin cli preload command

drjasonharrison-vp-eio:


pimterry:
@drjasonharrison-vp-eio thanks for the PR! Good catch, and good fix, but this isn't the right repo for it I'm afraid. Those docs actually get generated in the CLI themselves, and later imported into this repo, so we need to change it at the source.

I'm going to close this PR, but I've made the corresponding changes in the CLI itself over here: https://github.com/resin-io/resin-cli/pull/976. Once that's merged, the docs here will update next time we refresh them.

PR #921, opened 2018-10-22
small tweak to rpi config.txt var list section

dt-rush:
* clarified that these are only *some* of the variables which can be configured
* added link to raspberry pi docs for config.txt


shaunmulligan:
need to fix commit linting on resinCI so we can merge this.

shaunmulligan:
@dt-rush do you want to fix up this PR or should I just include it in one of my PRs?

PR #924, opened 2018-10-22
Mixpanel: Dynamically get the token from api.resin.io/config

jviotti:
I had to export jQuery from the main script so I could access it in the
Mixpanel snippet in order to request `/config`.

We're also changing the host where we post Mixpanel events to, as we are
now going to have all events going through the API.

Change-type: patch


jviotti:
Ping @dimitrisnl @LucianBuzzo @thgreasi

PR #925, opened 2018-10-23
Modify the repo to support serving content on a subpath

brownjohnf:
The path defaults to /docs, but can be overridden by setting
`PATH_PREFIX` as an env var. To actually work, this will require a newer
version of resin-io/doxx. PR coming for that repo.

Change-type: major
Signed-off-by: Jack Brown <jack@balena.io>


brownjohnf:
The PR this is waiting on is https://github.com/resin-io/doxx/pull/20

brownjohnf:
Okay, this should be ready for review/merge.

PR #930, opened 2018-10-30
Fix <version> Markdown formatting in base-images doc

mcornella:
Text between angle brackets is interpreted as HTML tags, so we need to escape the angle
brackets with a backslash.


shaunmulligan:
need to fix commit linting on resinCI so we can merge this.

chrisys:
Rolled up into https://github.com/balena-io/docs/pull/962

PR #969, opened 2018-11-16
Various issue fixes, dbus comms, custom hardware

chrisys:


lurch:
This page updated `runtime.md` to say that "Changing the device hostname is no longer possible via this method" (i.e. hostname can't be changed via dbus).
However at https://www.balena.io/docs/learn/develop/runtime/#dbus-communication-with-host-os it says "In some cases its necessary to communicate with the host OS systemd to perform actions on the host, for example changing the hostname."

Perhaps this example-of-why-you'd-want-to-use-dbus should be changed to something else? :wink:

garethtdavies:
@lurch there is an open PR that removes that "for example" text https://github.com/balena-io/docs/pull/1272/

PR #975, opened 2018-11-19
rewording of what happens to orphaned devices

drjasonharrison-vp-eio:


chrisys:
@drjasonharrison-vp-eio thank you for taking the time to contribute to the docs! It's my view that the existing wording is preferable to your proposed change, but please feel free to reopen & elaborate.

PR #998, opened 2019-01-03
fix gsm table number

nelsonpina:
fix gsm settings table number


shaunmulligan:
This was addressed in https://github.com/balena-io/docs/pull/1048

Issue #952, opened 2018-11-14
Documentation mobile version has wrong links and images.


dansku:

dansku:
Mobile version of the page is missing /docs/ as url prefix.

dimitrisnl:
@dansku 	I only see that the nav bar image is broken. Are there else I'm missing?

dansku:
@dimitrisnl the nav bar is broken, and also all the menu links lack `/docs/` in the url.

dimitrisnl:
@dansku 	Yeah, I only mentioned to images. I'm aware of the navigation issue.
It's pretty strange really. Both the desktop and mobile navigation use the same partial template. I believe this issue lies with `doxx` not properly applying the `docs` prefix.

Have to investigate some more.

dimitrisnl:
I have identified the issue for the links. 
We parse all the anchor elements with Jquery and prefix them with `docs`.

Now the mobile navigation is not present in the DOM while we do the above operation but instead rendered afterward. Consequently, when the navigation links are rendered in the document, they are missing the docs prefix. Working on a fix.

PR #1002, opened 2019-01-11
mobile:  Fix mobile navigation, search and logo rendering

dimitrisnl:
Connects-to: #952
Change-type: patch

### DIFF
--- a/static/scripts/main.js
+++ b/static/scripts/main.js
@@ -1,4 +1,3 @@
-var Ractive = require('ractive')
 var Headroom = require('headroom.js')
 var hljs = require('highlight.js/lib/highlight')
 var languages = require('./langs.js')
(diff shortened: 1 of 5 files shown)


dimitrisnl:
Currently, we parse all the link and apply the `docs` prefix. 
The links in the mobile navigation were not present in the DOM at that time and would not be updated.

Same goes for the image which would be missing the `docs` prefix for its path.

Another side effect was never initializing Algolia.

I chose to remove the `Ractive` library that was handling it. It was only used there so removing this dependency was a good choice.

In my fix, I simply hide the dropdowns and manipulate the height of the elements to make them invisible.

lurch:
Was this supposed to connect to #952 ?

dimitrisnl:
@lurch 	Good catch will update it.

static/scripts/main.js
```diff
@@ -154,24 +153,9 @@ $('.site-navigation')
...
-
-Ractive({
-  el: '#ract-mobile-sub-menu',
-  template: '#tpl-mobile-sub-menu',
-  data: {
-    collapsed: true
-  }
+$('.navbar-toggle').click(function() {
+  $(this).find('span').toggleClass('active')
+  $(this).parent().siblings().toggleClass('collapsed')
```
thgreasi: NIT: Could be
```suggestion
  $(this).closest('navbar').find('navbar-collapse').toggleClass('collapsed')
```

PR #1007, opened 2019-01-11
Update 'Pro Tips for slimming down your build'

willswire:
Referencing https://docs.npmjs.com/cli/cache.html#details, --force must be used with NodeJS versions higher than ver 5. "Cache corruption will either trigger an error, or signal to pacote that the data must be refetched, which it will do automatically. For this reason, it should never be necessary to clear the cache for any reason other than reclaiming disk space, thus why clean now requires --force to run."


shaunmulligan:
Hi @willswire  thanks for the contribution, these changes look great. Can we ask you a small favour and rewrite the commit log to conform to the commit guideline mentioned here https://github.com/balena-io/docs#how-to-contribute which helps with versioning and release of the docs.

PR #1010, opened 2019-01-15
Fix typos on base-images.md

gelbal:
Add minor changes to keep the document consistent as well

Change-type: patch
Signed-off-by: Firat Gelbal firat@balena.io


gelbal:
Thanks @chrisys -- these are good suggestions. Merging all of it.

PR #1011, opened 2019-01-15
nav: Fix submenu tabs navigation

dimitrisnl:
Change-type: patch


dimitrisnl:
Reusing the same template in mobile, results in having elements with the same ids.
There must be a more elegant way using swig, but I can't think of something better right now.

PR #1025, opened 2019-02-06
Prevent algolia input from from being submitted

dimitrisnl:
The algolia input was under a form. Before algolia would load, this form
could be submitted, resulting in an undesired redirect.

Connects-to: #1024
Change-type: patch

### DIFF
--- a/server/index.coffee
+++ b/server/index.coffee
@@ -38,17 +38,6 @@ getLocals = (extra) ->
 
 doxx.loadLunrIndex()
 
-app.get "#{config.pathPrefix}/search-results", (req, res) ->
-  { searchTerm } = req.query
-  res.render 'search', getLocals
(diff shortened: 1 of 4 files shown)


server/index.coffee
```diff
@@ -38,17 +38,6 @@ getLocals = (extra) ->
 
 doxx.loadLunrIndex()
 
-app.get "#{config.pathPrefix}/search-results", (req, res) ->
```
dimitrisnl: We're not using this route.

If we have results, algolia will handle the redirection.

templates/_search-form.html
```diff
@@ -1,3 +1,3 @@
-<form action="/search-results">
```
dimitrisnl: Removing this form that causes the redirect

Algolia does not depend on it https://community.algolia.com/places/examples.html#simple-input

PR #1027, opened 2019-02-07
Fix project links to github

dansku:
Change-type: patch
Signed-off-by: Daniel Andade <daniel@balena.io>


shared/getting-started/exampleProjects/generic/rpi-python.md
```diff
@@ -1,7 +1,7 @@
 ### Basic GPIO Control in Python
 
 {{> icon class="octicon octicon-mark-github" }}
-[Repository]({{ $links.githubProjects }}/balena-rpi-gpio-sample-with-python)
+[Repository]({{ $links.githubMain }}-playground/balena-rpi-gpio-sample-with-python)
```
dimitrisnl: Maybe create a githubPlayground entry instead?

lurch:
Seems to be a [few pages](https://github.com/balena-io/docs/search?q=balena-io-playground&unscoped_q=balena-io-playground) that are still using the hardcoded URL instead of the `githubPlayground` variable - do they need fixing?
I wonder if it's worth adding a CI task or something to automatically scan for hardcoded URLs that you know should be "variable-ised" instead? See also #1252 
ping @xginn8 @garethtdavies

PR #1030, opened 2019-02-13
Fix warning about BALENA_API_KEY to reflect that it's a device API key

pcarranzav:
As per https://forums.balena.io/t/security-implications-of-balena-api-key-envvar/5315/3 - the warning about it being a user key was extremely outdated.


shaunmulligan:
@pcarranzav I think versionist didn't like your changes here

shaunmulligan:
@pcarranzav can you resolve the conflict here and we can get it merged?

PR #1037, opened 2019-03-04
Clarify balena envs in CLI docs

nazrhom:
Make it clear that the envs command only applies to environment or
config variables.

Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>

### DIFF
--- a/pages/reference/cli.md
+++ b/pages/reference/cli.md
@@ -597,9 +597,7 @@ This command lists all application/device environment variables.
 If you want to see config variables, used to configure
 balena features, use the --config option.
 
-At the moment the CLI does not support per-service variables,
-so the following commands will only show service-wide
-environment variables.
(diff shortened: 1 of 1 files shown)


pdcastro:
@nazrhom, these CLI reference docs are somehow updated / overwritten automatically out of CLI pull requests like the one I merged a few hours ago:
https://github.com/balena-io/balena-cli/pull/1124/files

Which by the way conflicts with this PR, so better not to merge it. :-) Maybe it's something like a cron job that runs once a day (?), because they are not updated right after a pull request, but after some time the changes are published. Or maybe someone runs something manually every now and then... What I know for sure is that the changes that I merge through regular CLI pull requests eventually find their way to this page: https://www.balena.io/docs/reference/cli/

The latest change I found on that page is the `--registry-secrets` option for the `balena build` and `balena deploy` commands. The CLI PRs were merged last week, and by Monday (when I looked) the page was updated. :-)

Issue #1050, opened 2019-03-19
Links in TOC on FAQ page do not work 


chrisys:
https://www.balena.io/docs/faq/troubleshooting/faq

This is due to the `-` at the end of a lot of the links, probably related to https://github.com/balena-io/balena-cli/pull/1111 and https://github.com/balena-io/docs/pull/1039

PR #1047, opened 2019-03-14
Joe Docs update

joehounsham:
-Adding new commercial faqs to faq section section
- Fixing anchor bug in Faqs
- Correcting readme to point to http://localhost:3000/docs
- Updated SIM7600E in favour of SIM7110E

PR #1052, opened 2019-03-22
Fix TOC links in FAQ

sradevski:
After doing it I noticed there is [this PR](https://github.com/balena-io/docs/pull/1047), but since that one will be reopened and there are few more link fixes that were not in that PR it would be better to have this merged, and then they can do the PR on top of it without the link fixes.

Connects-to: #1050
Change-type: patch


joehounsham:
@sradevski What were the changes you made here? They aren't visible in the change log

sradevski:
@joehounsham It was mostly removing the `-` suffix from all URLs, and few of the URLs were wrong. I think it is easier to just see the changes in the PR diff.

Issue #940, opened 2018-11-09
404 page on docs renders without any styling


cyplo:
NUC flashing instructions `dashboard->application->add device` link to getting started guide using this URL:
http://docs.resin.io/nuc/nodejs/getting-started/#adding-your-first-device , this gets redirected to
https://www.balena.io/docs/learn/getting-started/docs/intel-nuc/nodejs/#adding-your-first-device

The page loads as pure text, with no styles in both Firefox and Chromium under Linux

![screenshot from 2018-11-09 08-18-11](https://user-images.githubusercontent.com/217899/48251109-75a5e100-e3f8-11e8-86b9-e4cb2f91aba5.png)

lurch:
If you scroll all the way to the bottom you can see it says
```
Page not found
We think we don't have a page you're looking for.
```

chrisys:
@cyplo this appears to be fixed now, or at least I cannot reproduce it. Reopen if still an issue 👍

lurch:
@chrisys Looks like any non-existent page still displays as completely unstyled, e.g. https://www.balena.io/docs/learn/page-not-here

chrisys:
@lurch thanks for pointing that out, have reopened and renamed to keep all the discussion here.

sradevski:
I just had a quick look at it, and the issue seems to be that the not-found page looks for the style and js at `/dist/<styles.css|main.js>`, but the other pages fetch it from `/docs/dist/<styles.css|main.js>`. I am assuming the reason is doxx adds the prefix to all static pages, but the not-found pages is not handled by doxx, therefore the error.

PR #1053, opened 2019-03-22
Fix the 404 page from not displaying correctly

sradevski:
In order to fix the issue I made the 404 page also be served
by Doxx, fixing the issue with the wrong routing of the css and 
js files. This one was very annoying, I am so glad it got fixed.

Connects-to: #940
Change-type: patch

### DIFF
--- a/pages/404.md
+++ b/pages/404.md
@@ -1,0 +1,4 @@
+---
+title: Not found
+layout: not-found.html
+---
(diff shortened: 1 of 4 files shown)


dimitrisnl:
I'm trying to reproduce it navigating to `https://www.balena.io/docs/learn/welcome/introduction/2`

but I get an unstyled page 
Screen Shot 2019-03-22 at 8 20 51 PM

PR #1061, opened 2019-03-26
Adding documentation about transferring application ownership

pcgeek86:


pcgeek86:
@shaunmulligan I've tried to fix this a bunch of times, but can't figure out what the problem is. The commit message has been corrected with a `git rebase --interactive HEAD~2` multiple times, but it's still rejecting it.

Issue #1020, opened 2019-01-31
Known Tested Modems: replace SIM7100E with SIM7600E


pdcastro:
* The docs mention we support the SIM7100E modem: https://www.balena.io/docs/reference/OS/network/2.x/#known-tested-modems-
* meta-balena doesn't mention the SIM7100E but the SIM7600E instead:
https://github.com/balena-os/meta-balena/blob/master/README.md#modems

@floion and @shaunmulligan have confirmed that _"meta-balena is correct, the SIM7100E is deprecated in favour of the 7600."_

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_h0er)

shaunmulligan:
closed by https://github.com/balena-io/docs/pull/1072

PR #1072, opened 2019-04-04
networking: Update tested modems list

shaunmulligan:
Change-type: patch
Signed-off-by: Shaun Mulligan <shaun@balena.io>


shaunmulligan:
closes https://github.com/balena-io/docs/issues/1020

PR #1075, opened 2019-04-10
Update build-optimization.md

willswire:
component: add --force flag.  Must be used with NodeJS versions higher than ver 5. "Cache corruption will either trigger an error, or signal to pacote that the data must be refetched, which it will do automatically. For this reason, it should never be necessary to clear the cache for any reason other than reclaiming disk space, thus why clean now requires --force to run."

Change-type: patch
Signed-off-by: William Walker <wwalker@udel.edu>


nazrhom:
@resin-ci retest

nazrhom:
@resin-ci retest

PR #1083, opened 2019-04-18
docs: fix up bandwidth reduction formatting

xginn8:
Change-type: patch


xginn8:
my only real changes are in pages/reference/supervisor/bandwidth-reduction.md

imrehg:
@xginn8 probably can stage only the relevant parts, the other, automatically pulled in files should be part of a separate PR as docs management (ie. a vendoring change). That should make things simpler.

xginn8:
> @xginn8 probably can stage only the relevant parts, the other, automatically pulled in files should be part of a separate PR as docs management (ie. a vendoring change). That should make things simpler.

Updated, though I don't love this process. imo we should be comfortable committing build artifacts, or committing them automatically when a dependent component is updated. To me this process is a point of friction for anyone outside balena committing to improving the docs.

PR #1086, opened 2019-04-22
baseImages: Update base images docs

nghiant2710:
- Add new stack: dotnet
- Update latest node and golang version

Change-type: patch
Signed-off-by: Trong Nghia Nguyen <nghiant2710@gmail.com>


lurch:
Should the updated node and golang versions also be added to the "Multiple language stacks" section?

PR #1087, opened 2019-04-25
devices: Add Jetson Nano device type

chrisys:
@shaunmulligan first draft for your review. There's a bit of a question around dev kit vs production Jetson Nanos as the process will likely change (dev kit is SD card, production is 16GB MMC on board).


vielmetti:
Two comments:

- if you supply power through a barrel jack, you also need to set a jumper on the board. My home stash of parts didn't include any jumpers, but I did have a short alligator cable as a temporary hack.
- the device I just got has an SD card only (no MMC).

chrisys:
@vielmetti thanks for the note about the barrel jack; have added. What I was meaning regarding the MMC is that the dev kits shipping now are using an SD card but from what I can tell when the standalone units are released they will include a 16GB MMC instead which will require a slightly different getting started guide.

PR #1088, opened 2019-04-26
Update local-mode.md

willswire:
component: in the updated local mode, a device will send logs back to the balenaCloud dashboard.

Change-type: minor
Signed-off-by: William Walker <wwalker@udel.edu>


willswire:
I was wrong!

PR #1089, opened 2019-04-30
Update ref docs for the CLI, Node SDK, Python SDK and Supervisor

pdcastro:
* Make the standalone installer the recommended option.
* Move installation instructions to a new INSTALL.md in the CLI repo.
* Also patch updates for the Node SDK, Python SDK and the Supervisor.

Btw, FYI, there is an open CLI PR implementing review changes requested by @hedss: 
https://github.com/balena-io/balena-cli/pull/1200

Change-type: minor

### DIFF
--- a/pages/reference/cli.md
+++ b/pages/reference/cli.md
@@ -1,85 +1,75 @@
 # Balena CLI Documentation
 
-This tool allows you to interact with the balena api from the comfort of your command line.
+The balena CLI (Command-Line Interface) allows you to interact with the balenaCloud and the
+[balena API](https://www.balena.io/docs/reference/api/overview/) through a terminal window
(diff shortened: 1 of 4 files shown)


xginn8:
@pdcastro this PR is the result of just a rebuild, right? i.e. all the changes here are actually pulled from other repositories

pdcastro:
> _(at)pdcastro this PR is the result of just a rebuild, right? i.e. all the changes here are actually pulled from other repositories_

That's correct -- I merely ran `npx -c './tools/prepare.sh'`. Something I wanted confirmation, though, is whether the updates for the SDKs and the Supervisor are "ready" for public publishing, or whether there is a reason not to publish them just yet. @CameronDiver, @shaunmulligan ?

PR #1093, opened 2019-05-01
config: update 'resinOS 2.x' to 'balenaOS 2.x'

xginn8:
Change-type: patch

### DIFF
--- a/shared/getting-started/getDeviceOnDash/bananapi-m1-plus.md
+++ b/shared/getting-started/getDeviceOnDash/bananapi-m1-plus.md
@@ -1,10 +1,9 @@
 Insert the SD card into your **{{ $device.name }}** and, if necessary, connect the ethernet cable or the USB WiFi adapter. Now power up your **{{ $device.name }}** by inserting the micro USB cable.
 
-It will take a minute or two for the **{{ $device.name }}** to appear on your {{ $names.company.lower }} [dashboard][resinDash]. While you wait, the resinOS is expanding the partitions on your SD card to use all available space, installing a custom Linux environment, and establishing a secure connection with the resin.io servers.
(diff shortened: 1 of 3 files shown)


lurch:
BTW doing a quick search shows there's also a couple of other pages which still say "resinOS" which should probably be properly parameterized so that they say "balenaOS" instead?

https://github.com/balena-io/docs/blob/master/shared/getting-started/getDeviceOnDash/bananapi-m1-plus.md

https://github.com/balena-io/docs/blob/master/pages/reference/OS/advanced.md

xginn8:
> BTW doing a quick search shows there's also a couple of other pages which still say "resinOS" which should probably be properly parameterized so that they say "balenaOS" instead?
> 
> https://github.com/balena-io/docs/blob/master/shared/getting-started/getDeviceOnDash/bananapi-m1-plus.md
> 
> https://github.com/balena-io/docs/blob/master/pages/reference/OS/advanced.md

Good idea to update those as well! I have updated this PR, thanks @lurch!

shared/getting-started/getDeviceOnDash/bananapi-m1-plus.md
```diff
@@ -1,10 +1,10 @@
 Insert the SD card into your **{{ $device.name }}** and, if necessary, connect the ethernet cable or the USB WiFi adapter. Now power up your **{{ $device.name }}** by inserting the micro USB cable.
 
-It will take a minute or two for the **{{ $device.name }}** to appear on your {{ $names.company.lower }} [dashboard][resinDash]. While you wait, the resinOS is expanding the partitions on your SD card to use all available space, installing a custom Linux environment, and establishing a secure connection with the resin.io servers.
+It will take a minute or two for the **{{ $device.name }}** to appear on your {{ $names.company.lower }} [dashboard][balenaDash]. While you wait, the balenaOS is expanding the partitions on your SD card to use all available space, installing a custom Linux environment, and establishing a secure connection with the balenaCloud servers.
```
lurch: Rather than hardcoding "balena" would it be better to use the same parameterization as e.g. https://github.com/balena-io/docs/blob/master/shared/getting-started/getDeviceOnDash/orangepi-plus2.md ?

PR #1096, opened 2019-05-07
Node version set to 10

imrehg:
* in testing on the CI servers
* in Heroku

Change-type: minor
Signed-off-by: Gergely Imreh <gergely@balena.io>


imrehg:
This should have been draft, as the CI still need some changes, that it doesn't break when there are no tests. So currently pending, will retest/modify later.

PR #1097, opened 2019-05-07
Restrict tests to node 10 and 12

nazrhom:
Since there are no tests, we run npm build in place of the tests

Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>


imrehg:
Closed due to #1096 already in the works.

PR #1098, opened 2019-05-13
Fix the example Dockerfile for crossbuilding

eifinger:
The example for building an ARM image on an x86 machine was using the outdated resin base image and installed python instead of python-pip


xginn8:
Hi @eifinger and thank you for contributing!

This change looks great, though your commit message should include the Change-type as per our contribution guidelines here: https://github.com/balena-io/docs/blob/master/CONTRIBUTING.md - without this the CI will fail to build and will not allow a merge. Since we had not publicized this well at the time of your commit, I've taken over this PR to get it merged ASAP.

xginn8:
@balena-ci retest

PR #1108, opened 2019-06-06
Update external docs (CLI, Node SDK)

pdcastro:
Change-type: patch


chrisys:
@pdcastro it looks like this PR is redundant (no changes) - could you double check it all looks as expected?

PR #1112, opened 2019-06-17
 balenaCLI: local push/scan have been deprecated

xginn8:
Change-type: patch

### DIFF
--- a/shared/overview/2.x.md
+++ b/shared/overview/2.x.md
@@ -28,7 +28,8 @@ The Development images are recommended while getting started with {{ $names.os.l
 The dev images enable a number of useful features while developing, namely:
 
 * Passwordless SSH into {{ $names.os.lower }} on port `22222` as the root user. So one can do `ssh root@<DEVICE_IP> -p22222` and poke around to see how the system runs.
-* Docker socket exposed on via port `2377`, which allows `{{ $names.company.lower }} local push` to do remote Docker builds on the target device.
+* Docker socket exposed on via port `2375`, which allows `{{ $names.company.lower }} push / build / deploy`
+  to do remote Docker builds on the target device (see [Deploy to your Fleet][deploy-to-fleet]).
 * Getty console attached to tty1 and serial.
 * Capable of entering [local mode][local-mode-link] for rapid development of application containers locally.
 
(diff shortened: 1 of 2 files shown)


shared/overview/2.x.md
```diff
@@ -28,7 +28,7 @@ The Development images are recommended while getting started with {{ $names.os.l
 The dev images enable a number of useful features while developing, namely:
 
 * Passwordless SSH into {{ $names.os.lower }} on port `22222` as the root user. So one can do `ssh root@<DEVICE_IP> -p22222` and poke around to see how the system runs.
-* Docker socket exposed on via port `2377`, which allows `{{ $names.company.lower }} local push` to do remote Docker builds on the target device.
+* Docker socket exposed on via port `2377`, which allows `{{ $names.company.lower }} push` to do remote Docker builds on the target device if in [local mode][local-mode-link].
```
pdcastro: * Isn't it port 2375 instead of 2377? I'll be disappointed with my memory if it betrays me like this, but it's telling me that I used port 2375 in the past. So much so, actually, that at one point I even added it to the help output of the `build` command:
```
$ balena help build
   --dockerPort, -p <dockerPort>              Docker daemon TCP port number (hint: 2375 for balena devices)
```
Also:
```
$ ssh -p 22222 root@192.168.1.118
root@b1878c6:~# netstat -ant
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
...
tcp        0      0 :::2375                 :::*                    LISTEN
```
Tested the above for a NUC in local mode and an RPi 3 **not** in local mode, both dev images. Both expose 2375, neither expose 2377.
* This section, "Dev vs. Prod images", seems to be more concerned about the fact that the Docker port is exposed, than the requirements of the push command (the device being in local mode). The `balena build` and `balena deploy` commands can also take advantage of the expose docker port, and do not require the device to be in local mode. So perhaps:
```
* Docker socket exposed on via port `2375`, which allows `{{ $names.company.lower }} push / build / deploy`
  to do remote Docker builds on the target device (see [Deploy to your Fleet](https://www.balena.io/docs/learn/deploy/deployment/)).
```
* Is this `.md` autogenerated from some source elsewhere? I don't know the answer, just a thought. :-)

xginn8: Changes added, thanks for looking! interestingly, `2377` has been around since it's inception: 7d0b399479f814bde82930a866c010992af4d12e (maybe that was changed elsewhere since then?)

I do not believe this `.md` is generated, but I believe many others are. I'm never 100% sure how these docs hang together with all the parameterization and other-doc-dependencies, but I usually just PR my changes and wait to be told if it's somehow incorrect.

PR #1126, opened 2019-07-02
Fix dashboard_domain references

mikesimos:
Change-type: patch
Signed-off-by: Michael Angelos Simos <michalis@balena.io>


mikesimos:
@balena-ci retest

PR #1133, opened 2019-07-10
Fix fixup_x extension

rpidanny:
Change-type: patch
Signed-off-by: Abhishek Maharjan <abhishek.rpidanny@gmail.com>


chrisys:
@balena-ci retest

PR #1135, opened 2019-07-15
Add documentation about supervisor volume cleanup operation

CameronDiver:
Change-type: patch
Signed-off-by: Cameron Diver <cameron@balena.io>

### DIFF
--- a/shared/general/persistent-storage.md
+++ b/shared/general/persistent-storage.md
@@ -1,4 +1,3 @@
-
 If you have data or configurations that you would like to persist through application and host OS updates, you have the option to keep them in persistent storage. Persistent storage is a good place to write system logs and other application data that should remain untouched even as your code changes.
 
 #### Before {{ $names.os.lower }} v2.12.0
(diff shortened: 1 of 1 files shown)


shared/general/persistent-storage.md
```diff
@@ -19,6 +18,23 @@ When using named volumes, note that:
 - If a device is moved to a new application, the old `/data` folder will be automatically purged.
 - During the build process, data added to a container directory that is configured to link to a named volume will be copied to the volume the first time it's created on the device.
 
+#### Using a Supervisor with a version >= v10.0.0
+
+Since v10.0.0 of balena-supervisor, volumes are no longer
+removed when an application no longer has a reference to
```
shaunmulligan: This section `, volumes are no longer
removed when an application no longer has a reference to...` seems clunking and I struggle to understand it, is there a better way to rephrase it.

shared/general/persistent-storage.md
```diff
@@ -19,6 +18,23 @@ When using named volumes, note that:
 - If a device is moved to a new application, the old `/data` folder will be automatically purged.
 - During the build process, data added to a container directory that is configured to link to a named volume will be copied to the volume the first time it's created on the device.
 
+#### Using a Supervisor with a version >= v10.0.0
+
+Since v10.0.0 of balena-supervisor, volumes are no longer
```
chrisys:
```suggestion
Since balena-supervisor v10.0.0, volumes are no longer automatically removed from disk when references to them are removed from an application's `docker-compose` file. This means that it's no longer possible for data to be lost due to the accidental rename of a volume. 

If you change volume names regularly, your device will now continue to retain all previous volumes from an application including their contents. To avoid this the supervisor API now provides an [endpoint to cleanup unreferenced volumes](https://www.balena.io/docs/reference/supervisor/supervisor-api/#cleanup-volumes-with-no-references). Additionally, it is possible to perform this action from the dashboard via the `Purge Data` action, found on the `Actions` tab for a device.

**Note:** volumes will continue to be removed automatically when moving a device between applications.
```

shared/general/persistent-storage.md
chrisys: I thought it was a bit clunky too, I put a suggestion on

PR #1145, opened 2019-07-26
Update AWS-IoT Documentation

dansku:
Change-type: minor
Signed-off-by: Daniel Andrade <daniel@balena.io>


dansku:
I just realized the build changed a few sdk files, should that be the case as well? Otherwise, I can remove those changes.

xginn8:
@dansku for one reason or another, the CLI/SDK/supervisor docs are checked in here, but deployed directly from their respective masters. Imho including the changes is fine here since there's no effective change other than keeping our docs repo in sync.

dansku:
@xginn8 great, then it should be fine ;)

Issue #1122, opened 2019-06-27
Clarify whitelisted domains subtlety in networking requirements


xginn8:
The documentation currently states that only three domains must be whitelisted, but by default NTP & DNS default to domains outside that whitelist (`*.resinio.pool.ntp.org` & `8.8.8.8`). I _think_ the intent is for these domains to fall under this category: "outward only (and inward once outward connection established)", though that is not entirely clear from context.

pdcastro:
> [pdcastro] This issue has attached support thread https://jel.ly.fish/#/support-thread~fe9f29fc-9fda-4589-856a-eee45e505074

It seems that `*.s3.amazonaws.com` also needs to be whitelisted, as found in the support thread linked above.

PR #1168, opened 2019-08-24
Add resinio.pool.ntp.org to whitelist requirements and fetch external docs

pdcastro:


shared/network/2.x.md
```diff
@@ -367,6 +367,7 @@ Additionally, you should whitelist the following domains for the relevant ports
 * `*.{{ $names.dashboard_domain }}`
 * `*.docker.com`
 * `*.docker.io`
+* `*.resinio.pool.ntp.org`
```
pdcastro: The line above is the main "reviewable" change in this PR (and it has its own commit).
The other commit is the result of fetching external docs.

xginn8:
This change also relates to https://github.com/balena-io/docs/issues/1122

pdcastro:
Before adding `*.resinio.pool.ntp.org` to the documented list of domain names that require whitelisting, maybe I could also invite comments from @majorz or @imrehg. Whether or not domain names require whitelisting would depend on the exact local network configuration... I guess that whitelisting is more common in combination with HTTP proxies (matching the domain names against the HTTP headers), rather than, say, low-level UDP port 123 used by NTP. But conceptually, a firewall could be configured to block / allow IP address ranges and those IP addresses could be specified by means of domain names, so specifying the whitelisting of `*.resinio.pool.ntp.org` might also mean allowing "communication" with the IP addresses that the domain names resolve to. It feels like it doesn't hurt to "disclose" `*.resinio.pool.ntp.org` in the list.

PR #1175, opened 2019-09-18
docs: update device type names for Raspberry Pi 4

peterblazejewicz:
Change-type: patch

Thanks!


chrisys:
Hey @peterblazejewicz ! Thanks so much for the contribution. There's a small mistake in that the machine name for the Pi 4 is actually `raspberrypi4-64`. I was in the process of adding this and the Pi 3 one anyway, so this is superseded by https://github.com/balena-io/docs/pull/1179

peterblazejewicz:
@chrisys thanks for update! Appreciate you're working on the PIs!

PR #1178, opened 2019-09-30
diagnostics: update check definitions to 4.8.0

xginn8:
Depends-on: https://github.com/balena-io/resin-proxy/pull/410
Change-type: patch

### DIFF
--- a/pages/reference/diagnostics.md
+++ b/pages/reference/diagnostics.md
@@ -4,16 +4,19 @@ As part of the diagnostics suite, you will find a group of checks that can be co
 description of each check and what each means or how to triage.
 
 A `check` in this context is defined as a function that returns a result (good/bad status plus some descriptive text), whereas a
-command is simply a data collection tool without any filtering or logic built in). Checks are intended to be used by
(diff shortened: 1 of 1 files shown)


imrehg:
I think there is a heads up, not to merge anything without pinging Mike, can link to it later, but mentioning in case. I *think* this is still in effect.

lurch:
There's a "closing bracket" at `or logic built in)` which doesn't seem to have a matching "opening bracket" ?  :eyes:

pages/reference/diagnostics.md
```diff
@@ -7,13 +7,16 @@ A `check` in this context is defined as a function that returns a result (good/b
 command is simply a data collection tool without any filtering or logic built in). Checks are intended to be used by
```
imrehg: Is this a good place to put the bracket or better to dispose of the bracket in this section entirely?
```suggestion
command is simply a data collection tool (without any filtering or logic built in). Checks are intended to be used by
```
based on @lurch's comment?

xginn8: Nice catch @lurch, I've opened a PR in the repo where we source this doc, and will pull that in once it's been approved.

PR #1179, opened 2019-10-01
Add 64-bit Pi 3 and Pi 4 machine names

chrisys:
Change-type: patch
Signed-off-by: Chris Crocker-White <chriscw@balena.io>


lurch:
Just out of curiosity - does this mean that Balena doesn't support 32-bit OS on the Pi 4?

chrisys:
@lurch we don't have a 32-bit balenaOS for the Pi 4, but you can of course run a mix of 32-bit and 64-bit containers on the 64-bit OS.

PR #1183, opened 2019-10-14
Pass anonymous tracking ID to dashboard URLs

roman-mazur:
Change-type: patch


roman-mazur:
Yes, we'll need to handle utm parameters separately. They are not yet
tracked automatically except Google Analytics.

PR #1184, opened 2019-10-15
Correct Python snippet

archiehenderson:
The current snippet has `lock` on the end of the path, which results in a lockfile called `updates.lock.lock`, which does not have any effect on updates.


xginn8:
thanks for the contribution @archiehenderson! can you just add a footer to your commit with the following for the CI to pass:

```
Change-type: patch
```

If you can amend the commit to add that and push it back up, the tests should succeed. Otherwise, looks great!

xginn8:
@balena-ci retest

PR #1192, opened 2019-10-24
[WIP] Next steps partial

garethtdavies:


shared/getting-started/nextSteps/_default.md
```diff
@@ -0,0 +1,12 @@
+- Try out [local mode][local-mode], the most efficient way to rapidly develop and test your {{ $names.company.lower }} application.
```
sradevski: I think `local mode` is a concept users might not be familiar with to begin with, so a more concrete explanation of it might be more appropriate.

shared/getting-started/nextSteps/_default.md
```diff
@@ -0,0 +1,12 @@
...
+- Build an application that uses [multiple containers][multicontainer].
```
sradevski: I think we should focus more on the benefits of multicontainers here, as it might not be clear to everyone.

garethtdavies:
Closed, as included in #1195

PR #1198, opened 2019-10-28
Update CLI npm installation note in Getting Started guide

pdcastro:
Change-type: patch

### DIFF
--- a/shared/getting-started/balenaCliAuth/_default.md
+++ b/shared/getting-started/balenaCliAuth/_default.md
@@ -2,7 +2,8 @@ The recommended way to deploy code is to install the [balenaCLI][cli]. The easie
 
 <img src="/img/common/cli/download-installer.png" width="60%">
 
-__Note:__ You may also install the balenaCLI via npm on a system running NodeJS. In a terminal on your development machine run `npm --global install balena-cli`.
+__Note:__ You may also install the balenaCLI via npm on a system running NodeJS, as explained in
+[NPM Installation][npminstall].
 
 To use the [balenaCLI][cli], you need to login to your {{ $names.company.lower }} account. Login via the terminal using the `balena login` command:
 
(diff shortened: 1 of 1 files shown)


shared/getting-started/balenaCliAuth/_default.md
```diff
@@ -2,7 +2,8 @@ The recommended way to deploy code is to install the [balenaCLI][cli]. The easie
 
 <img src="/img/common/cli/download-installer.png" width="60%">
 
-__Note:__ You may also install the balenaCLI via npm on a system running NodeJS. In a terminal on your development machine run `npm --global install balena-cli`.
+__Note:__ You may also install the balenaCLI via npm on a system running NodeJS, as explained in
+[NPM Installation][npminstall].
```
pdcastro: From my own experience and recent feedback from first-time users in the Hackaton:
* `npm --global install balena-cli` is almost guaranteed to fail on Windows without first installing the dependencies mentioned in the installation instructions, including for example downloading and installing the WDK from Microsoft (plus git, Python, Node, etc.).
* It is also "likely" to fail on Ubuntu and other Linux distros if the system Node is used instead of nvm, as mentioned in the installation instructions.
* `npm --global install balena-cli` duplicates and mismatches the command line given in the installation instructions document, which is `npm install balena-cli --global --production --unsafe-perm`.

The previous getting started wording (before this PR's changes) implied that running "npm install" is likely to succeed, and some Hackaton Windows users did just that, only to get dependency errors.

shared/getting-started/balenaCliAuth/_default.md
```diff
@@ -39,6 +40,7 @@ ID    APP NAME   DEVICE TYPE     ONLINE DEVICES DEVICE COUNT
 
 __Note__ See all the commands available with balenaCLI by running `balena help` 
 
-[cli]:https://github.com/balena-io/balena-cli
+[cli]:https://www.balena.io/docs/reference/cli/
```
pdcastro: The balena.io CLI page (as opposed to the github.com CLI page) is composed of automatically extracted, selected sections from the CLI README.md file and also adds the CLI command reference, plus having links to the github repo (e.g. links INSTALL.md, FAQ, etc.) IMHO, it's a nicer page to use, both in usefulness and visual appearance, as a link for balenaCLI.

garethtdavies: Thanks. I need to pass through this guide again to update a number of links to use variables i.e. https://github.com/balena-io/docs/blob/master/config/links.coffee but I'll do this post-merge as this is a better update.

shared/getting-started/balenaCliAuth/_default.md
garethtdavies: Thanks. Some of this copy was referenced from the CLI masterclass so I'll ensure any instructions match there.

PR #1218, opened 2019-11-06
Adding container contract documentation

garethtdavies:
Change-type: minor

### DIFF
--- a/pages/learn/develop/container-contracts.md
+++ b/pages/learn/develop/container-contracts.md
@@ -1,0 +1,112 @@
+---
+title: Container contracts
+excerpt: A guide to using container contracts to enforce device compatibility
+---
+
+# Container contracts
+
+**Note** Container contracts are available for devices running Supervisor versions >= 10.6.17. All prior versions will not enforce any contracts.
+
+Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the device is running a specific version of [Supervisor][supervisor] or has a specific version of the [NVIDIA Tegra Linux Driver Package][l4t] (L4T).
+
+Container contracts are enforced on the device via the device Supervisor. Each service can define the contract requirements that it enforces, and if a contract's requirements are not met, the release is not deployed to the device, unless it contains services labeled as [optional](#optional-containers). You only need to define a contract for services that you wish to enforce a contract upon. Currently, you may define contract requirements based on the versions of Supervisor and L4T.
+
+## Contract specification
+
+Container contracts are defined as a `contract.yml` file and must be placed in the root of the build context for a service. For example, the following [multi-container application][multicontainer] defines two services named `first-service` and `second-service`, each with a corresponding `contract.yml` file.
+
+```shell
+.
+├── docker-compose.yml
+├── first-service
+│ ├── contract.yml
+│ └── Dockerfile.template
+└── second-service
+  ├── contract.yml
+  └── Dockerfile.template
+```
+
+`contract.yml` is a YAML file containing `type`, `slug`, `name`, and `requires` keys, which define the requirements of the contract. The following example requires the device to be running a version of L4T equal to 28.2 for the contract to pass. For L4T exact versions (either 28.2 or 32.2) must be specified.
+
+```yaml
+type: "sw.container"
+slug: "enforce-l4t-version"
+name: "Enforce L4T Version"
+requires:
+  - type: "sw.l4t"
+    version: "28.2"
+```
+
+- `type` should be set to `sw.container` for container contracts.
+- `slug` should be a unique identifier to identify the contract.
+- `name` should be a descriptive name of the contract.
+- `requires` should be a list of requirements to be enforced by the contract defined by `type` and `version` keys. Currently, `type` may be either `sw.supervisor` for the device Supervisor version or `sw.l4t` for the L4T version.
+
+**Note** If `type`, `slug`, or `name` are omitted, the build will fail.
+
+Multiple requirements may be added to a single contract. The following example requires that the Supervisor version is greater than or equal to 10.6.17, and the L4T version is equal to 32.2.
+
+```yaml
+type: "sw.container"
+slug: "enforce-supervisor-and-l4t"
+name: "Enforce supervisor and l4t requirements"
+requires:
+  - type: "sw.supervisor"
+    version: ">=10.6.17"
+  - type: "sw.l4t"
+    version: "32.2"
+```
+
+**Note** Any `requires` type other than `sw.supervisor` or `sw.l4t` will result in the contract being invalid.
+
+When deploying a release, if the contract requirements are not met, the release will fail, and all services on the device will remain on their current release. This default behavior can be overridden by including an [optional label](#optional-containers) in the application docker-compose.yml file. This optional label specifies which services with unmet requirements will be ignored, with all other services on the device updating to the new release. If there are any existing running services with unmet requirements, they will not be destroyed and continue running the prior release.
+
+Should a contract fail, the dashboard logs will contain detail about the failing contract:
+
+```shell
+Could not move to new release: Some releases were rejected due to having unmet requirements:
+Contracts: Services with unmet requirements: first-service
+```
+
+**Note** For application fleets that contain a mixture of Supervisor and L4T versions, contracts may be used to ensure only compatible devices run specific services, as releases will only be deployed to those meeting the contract requirements.
+
+## Optional containers
+
+By default, when a container contract fails, none of the services are deployed to the device. However, in a multi-container application, it is possible to ignore those services that fail the contract requirements with the other services being deployed as normal. To do so, we make use of the `io.balena.features.optional: 1` [Supervisor label][labels] to indicate which services should be considered optional.
+
+In the `docker-compose.yml` file of the application, add the `io.balena.features.optional: 1` to the labels list for each service you wish to mark as optional. In the following example, even if the `first-service` container contract fails, the `second-service` service will still be deployed (assuming it doesn’t have any failing contracts of its own).
+
+```Dockerfile
+version: '2'
+services:
+  first-service:
+    build: ./first-service
+    labels:
+      - io.balena.features.optional: 1
+  second-service:
+    build: ./second-service
+```
+
+**Note** When using optional containers, if there are any existing running services of the failed contract, they will be killed and not continue running the old version.
+
+## Board support
+
+Container contracts are supported on all devices where the Supervisor version is >= 10.6.17. If a device in the application does not support L4T, and the contract specifies an L4T requirement, the device supervisor will reject the release unless the contract is marked as optional. L4T support is available on the following boards:
+
+- Aetina N510 TX2
+- Auvidea J120 TX2
+- CTI Orbitty TX2
+- CTI Spacely TX2
+- Nvidia blackboard TX2
+- Nvidia D3 TX2
+- Nvidia Jetson Nano
+- Nvidia Jetson TX1
+- Nvidia Jetson TX2
+- Nvidia Jetson Xavier
+- SKX2
+
+[os]: /reference/OS/overview/2.x/
+[supervisor]: /reference/supervisor/supervisor-api/
+[multicontainer]: /learn/develop/multicontainer/
+[labels]: /learn/develop/multicontainer/#labels
+[l4t]: https://developer.nvidia.com/embedded/linux-tegra
--- a/config/navigation.txt
+++ b/config/navigation.txt
@@ -40,6 +40,7 @@ Learn
     [/learn/develop/local-mode]
     [/learn/develop/dockerfile]
     [/learn/develop/multicontainer]
+    [/learn/develop/container-contracts]
     [/learn/develop/runtime]
     [/learn/develop/hardware]
       [/learn/develop/hardware/gpio]


pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Container Contracts
+excerpt: A guide to using container contracts to enforce device compatibility
+---
+
+# Container contracts
+
+__Note__ Container contracts are available for devices running Supervisor versions >= 10.6.0. All prior versions will not enforce any contracts.
+
+Container contracts are used to ensure the compatibility of a device to run an application service. For example, container contracts may be used to ensure that the host operating system is running a specific version of balenaOS or Supervisor such that if an application is dependent on a specific version, it only gets deployed to those devices that can meet these requirements to fulfill the contract. Currently, you may define contract requirements based on the versions of [balenaOS](/reference/OS/overview/2.x/) and [Supervisor](/reference/supervisor/supervisor-api/).
```
CameronDiver:
```suggestion
Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the host operating system is running a specific version of balenaOS or balena-supervisor such that if an application is dependent on a specific version, it only gets deployed to those devices that can meet these requirements to fulfill the contract. Currently, you may define contract requirements based on the versions of [balenaOS](/reference/OS/overview/2.x/) and [balena-supervisor](/reference/supervisor/supervisor-api/).
```

I'm not sure on these, and if you think the original is better I don't really mind :)

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+Should a contract fail, the logs will contain detail about the failing contract:
+
+```shell
+Could not move to new release: Some releases were rejected due to having unmet requirements:
+Contracts: Services with unmet requirements: first-service
+```
+
+__Note__ For application fleets that contain a mixture of hostOS and Supervisor versions, contracts may be used to ensure only compatible devices run specific services as releases will only be deployed to those meeting the contract requirements.
+
+## Optional Contracts
```
CameronDiver:
```suggestion
## Optional Containers
```

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+excerpt: A guide to using container contracts to enforce device compatibility
+---
+
+# Container contracts
+
+__Note__ Container contracts are available for devices running Supervisor versions >= 10.6.0. All prior versions will not enforce any contracts.
+
+Container contracts are used to ensure the compatibility of a device to run an application service. For example, container contracts may be used to ensure that the host operating system is running a specific version of balenaOS or Supervisor such that if an application is dependent on a specific version, it only gets deployed to those devices that can meet these requirements to fulfill the contract. Currently, you may define contract requirements based on the versions of [balenaOS](/reference/OS/overview/2.x/) and [Supervisor](/reference/supervisor/supervisor-api/).
+
+Each service container can define the contracts that it enforces. If a container's contract requirements are not met, the release is not deployed to the device. You only need to define a contract for service containers that you wish to enforce a contract.
```
CameronDiver: `upon` at the end?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Contract specification
+
+Container contracts are defined as a `contract.yml` file and must be placed in the application structure alongside the corresponding service Dockerfile. For example, the following [multi-container application](/learn/develop/multicontainer/) defines two services named `first-service` and `second-service`, each with a corresponding `contract.yml` file.
```
CameronDiver: This is a little strange, as the Dockerfile is not necesarilly required, for example Dockerfile.template, and also package.json are supported. I'd say maybe "the root of the build context for a service"

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+  - type: 'sw.os'
+    version: '>=2.40.0'
+```
+
+* `type` should be set to `sw.container` for container contracts.
+* `slug` should be a unique identifier to identify the contract.
+* `name` should be a descriptive name of the contract.
+* `requires` should be a list of requirements to be enforced by the contract defined by `type` and `version` keys. Currently, `type` may be either `sw.os` for the hostOS version or `sw.supervisor` for the device Supervisor version. 
+
+__Note__ If `name` or `type` are not supplied, the build will fail. If `slug` is not provided, the build will complete, but the contract will fail validation on deployment. If `requires` is not provided, the contract will have no effect and the release will be deployed.
```
CameronDiver: I'm going to fix this so maybe change it to "if `name`, `type` or `slug` are ommitted, the build will fail".

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+
+Multiple contract requirements may be added to a single contract. The following example requires that the hostOS is greater than or equal to 2.44.0, and the Supervisor version is greater than 10.
```
CameronDiver: Maybe "Multiple requirements"  (rather than "Multiple contract requirements") ?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+slug: 'enforce-supervisor-and-os'
+name: 'Enforce hostOS and supervisor'
+requires:
+  - type: 'sw.supervisor'
+    version: '>10'
+  - type: 'sw.os'
+    version: '>=2.44.0'  
+```
+
+__Note__ Types in the contract requirements are not validated, so passing values for `type` other than `sw.os` or `sw.supervisor` will result in the contract passing.
```
CameronDiver: I think this is actually the opposite result, they're not checking, but there's no way that the contract engine could fulfill the contract, so it's denied.

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+
+When deploying a release, if the contract requirements are not met, the release will fail, and all devices in the application will remain on their current release. This default behavior can be overridden by including an [optional label](#optional-contracts) in the application docker-compose.yml file. If there are any existing running services, they will not be destroyed and continue running the previous release.
```
CameronDiver: This reads a little weird, it's almost as if you're saying that adding the label will make the entire release ignore contracts, where as it's based on a single service.

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+
+Should a contract fail, the logs will contain detail about the failing contract:
```
CameronDiver: "dashboard logs" maybe?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,90 @@
...
+```shell
+Could not move to new release: Some releases were rejected due to having unmet requirements:
+Contracts: Services with unmet requirements: first-service
+```
+
+__Note__ For application fleets that contain a mixture of hostOS and Supervisor versions, contracts may be used to ensure only compatible devices run specific services as releases will only be deployed to those meeting the contract requirements.
+
+## Optional Contracts
+
+By default, when a service contract fails, none of the services are deployed to the device. However, in a multi-container application, it is possible to ignore those services that fail the contract requirements with the other services being deployed as normal. To do so, we make use of [Supervisor labels](/learn/develop/multicontainer/#labels) to indicate which services should be considered optional.
```
CameronDiver: We need to add the label to this section too

pages/learn/develop/container-contracts.md
garethtdavies: I struggled with the language here. The reason I added this was I could imagine bugs with typos and the contract not behaving as expected. As in, you may expect the contract to validate that types are being checked and so passing an invalid type isn't allowed whereas actually it's just going to be ignored. I'll try and clean up the language though.

pages/learn/develop/container-contracts.md
garethtdavies: Perfect, yeh I added here while I noted it but was a little prompt for this to be improved 👍

pages/learn/develop/container-contracts.md
garethtdavies: Thanks, that's just my lack of understanding on that one 😄

pages/learn/develop/container-contracts.md
garethtdavies: Something like the below, leaving the larger explanation to the dedicated section?

```suggestion
By default, when a service contract fails, none of the services are deployed to the device. However, in a multi-container application, it is possible to ignore those services that fail the contract requirements with the other services being deployed as normal. To do so, we make use of the `io.balena.features.optional=1` [Supervisor label](/learn/develop/multicontainer/#labels) to indicate which services should be considered optional.
```

pages/learn/develop/container-contracts.md
garethtdavies: @CameronDiver actually I'm confused on this. If I pass an invalid type say `sw.oss` the contract passes. I'm struggling with wording as not sure it 'passes' in as much as it isn't enforced. Is this the intended behaviour as I am not sure 'denied' explains it any better.

```
__Note__ Types in the contract requirements are not validated, so passing values for `type` other than `sw.os` or `sw.supervisor` will result in the contract not being enforced.
```

CameronDiver: Yeah it's how you describe, because we're not validating types (maybe we should be in the future?) the engine will just accept any type and try to reconcile it. Because there won't be an item in the engine universe which will conform to, say `sw.oss`, the engine will deny the release.

Basically, think about it like the engine not reporting anything different when there's no contracts which fulfill a type, and there's contracts which fulfill the type, but not the version requirement. Does this help?

garethtdavies: Right, that's how I understand except I think the engine doesn't "deny the release" it "passes" and is deployed. I think that's the source of the confusion. Are you saying that if I set type to `sw.oss` the contract should fail and not be deployed?

CameronDiver: Oh, this is weird, I haven't actually tested this specific case so maybe I'm wrong. If that is indeed the case, we'll need to do validation on the input types. I'll test this and get back to you.

garethtdavies: I'll also confirm this with testing as I'm not so sure anymore, perhaps it wasn't deployed but there was nothing in the logs. Either way, I'll double-check.

garethtdavies: Resolved via https://www.flowdock.com/app/rulemotion/i-docs/threads/7nQx7ciXHAw9MYpAJZxg6nEanK6 and #1218 314b6c5bc3d670e43bad95a0f93ad09e496ac9f9

pages/learn/develop/container-contracts.md
```diff
@@ -1,19 +1,19 @@
...
 
-Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the host operating system is running a specific version of balenaOS or balena-supervisor such that if an application is dependent on a specific version, it only gets deployed to those devices that can meet these requirements to fulfill the contract. Currently, you may define contract requirements based on the versions of [balenaOS](/reference/OS/overview/2.x/) and [balena-supervisor](/reference/supervisor/supervisor-api/).
+Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the host operating system is running a specific version of [{{ $names.os.lower }}][os] or [{{ $names.company.lower }} Supervisor][supervisor] such that if an application is dependent on a specific version, it only gets deployed to those devices that can meet these requirements to fulfill the contract. Currently, you may define contract requirements based on the versions of {{ $names.os.lower }} and Supervisor.
```
shaunmulligan: I assume on devices that don't meet the requirements will still download the images? or does it not get deployed at all. I thought the contract only enforced if it was run or not, but all of them would have it?

pages/learn/develop/container-contracts.md
```diff
@@ -44,7 +44,7 @@ requires:
 
 __Note__ If `name`, `type`, or `slug` are omitted, the build with fail.
 
-Multiple requirements may be added to a single contract. The following example requires that the hostOS is greater than or equal to 2.44.0, and the Supervisor version is greater than 10.
+Multiple requirements may be added to a single contract. The following example requires that the {{ $names.os.lower }} is greater than or equal to 2.44.0, and the Supervisor version is greater than 10.
```
shaunmulligan: I think we will have a problem with ESR os versions here.

pages/learn/develop/container-contracts.md
```diff
@@ -59,7 +59,7 @@ requires:
...
+When deploying a release, if the contract requirements are not met, the release will fail, and all devices in the application will remain on their current release. This default behavior can be overridden by including an [optional label](#optional-containers) in the application docker-compose.yml file. This label specifies which services with unmet requirements will be ignored, with the release being deployed to all other devices in the application. If there are any existing running containers for services with unmet requirements, they will not be destroyed and continue running the previous release.
 
 Should a contract fail, the dashboard logs will contain detail about the failing contract:
```
shaunmulligan: @CameronDiver will this fail over and over, or just the once?

shaunmulligan: just thinking of people on cellular

pages/learn/develop/container-contracts.md
```diff
@@ -59,7 +59,7 @@ requires:
...
+When deploying a release, if the contract requirements are not met, the release will fail, and all devices in the application will remain on their current release. This default behavior can be overridden by including an [optional label](#optional-containers) in the application docker-compose.yml file. This label specifies which services with unmet requirements will be ignored, with the release being deployed to all other devices in the application. If there are any existing running containers for services with unmet requirements, they will not be destroyed and continue running the previous release.
```
shaunmulligan: Its unclear where the contract resolution is done, is this on the builder or on the device. If on the builder, does the builder query all the OS versions. I assume it must be the device that decides these things? since if the device has a target release with a failed optional container, when it does a hostOS upgrade to the requisite version. I assume it would then start running the option container @CameronDiver ?

pages/learn/develop/container-contracts.md
CameronDiver: It will be rejected once every target poll, so no difference in terms of bandwidth

pages/learn/develop/container-contracts.md
CameronDiver: Yeah, it all happens on-device, and as the hup occurs the new valid containers will run.

pages/learn/develop/container-contracts.md
CameronDiver: It won't be downloaded in the contract fails - this is something that we did discuss and I'd like to implement, but I'd rather it was a future work thing.

pages/learn/develop/container-contracts.md
nazrhom: I think the issue here is the type `sw.os` is too generic. We often said ESR should be considered a different OS, as such it should have a different contract. In general we might have several os contracts (ubuntu, raspbian, ecc) in the system. The contract should express that it really wants balenaOS, so something like.
```
requires:
  - type: 'sw.os'
    version: '>=2.44.0'  
    slug: 'balenaOS'
```

This still sounds tricky to get right with ESR though, as we might want to represent the idea that compatibility with certain balenaOS versions imply compatibility with certain ESR versions. I think the correct way to do this is to further decompose the OS contract in a set of capabilities which are independent of the OS slug

pages/learn/develop/container-contracts.md
garethtdavies: Just to be clear, I think this is fully detailed later. Needs more clarification here?

pages/learn/develop/container-contracts.md
garethtdavies: Making sure I understand this (will give this a test later). If I have a release where the contract fails, say the OS is below a certain threshold and the release is not deployed to any devices (so not using optional containers). When I then do the OS upgrade on the device so the contract passes this release is then immediately deployed to all devices? Or does this only work in the optional container case as Shaun details above?

Also, what is hup? host upgrade?

garethtdavies: I confirmed this behaviour by updating the hostOS locally from a failing to passing version and indeed the optional container named `contract` started up upon reboot.

```
07.11.19 21:12:44 (+0000) Not running containers because of contract violations: contract
07.11.19 21:47:06 (+0000) Supervisor starting
07.11.19 21:47:08 (+0000) Killing service 'frontend sha256:d7c9e8e01da39ce3c3442a727f3aea3531e6562b687ca084a9088060c412f5bf'
07.11.19 21:47:08 (+0000) Killing service 'proxy sha256:44974585ebe63e4010fbbfbe161357c857b2bb4c965a6107c6da243af3a50cb2'
07.11.19 21:47:08 (+0000) Killing service 'data sha256:9c9688ec1e49f5c0c4c939c1bd329c6be01cc2b48b14f3f428161445c68e6253'
.......
07.11.19 21:47:59 (+0000) Downloading delta for image 'registry2.balena-staging.com/v2/66b98834140f6bfbb36920c432457142@sha256:0525c2f751b8d41f665e4eeacd97460f6e2658eb44353c71c39b5d37c2555565'
07.11.19 21:53:49 (+0000) Downloaded image 'registry2.balena-staging.com/v2/66b98834140f6bfbb36920c432457142@sha256:0525c2f751b8d41f665e4eeacd97460f6e2658eb44353c71c39b5d37c2555565'
07.11.19 21:53:49 (+0000) Installing service 'contract sha256:835937c4045a877fad29310369262e97dcfacbec06447fa4219d0ba817c68499'
07.11.19 21:53:52 (+0000) Installed service 'contract sha256:835937c4045a877fad29310369262e97dcfacbec06447fa4219d0ba817c68499'
07.11.19 21:53:52 (+0000) Starting service 'contract sha256:835937c4045a877fad29310369262e97dcfacbec06447fa4219d0ba817c68499'
07.11.19 21:53:56 (+0000) Started service 'contract sha256:835937c4045a877fad29310369262e97dcfacbec06447fa4219d0ba817c68499'
```

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+ - type: 'sw.l4t'
+ version: '28.2'
+```
+
+* `type` should be set to `sw.container` for container contracts.
+* `slug` should be a unique identifier to identify the contract.
+* `name` should be a descriptive name of the contract.
+* `requires` should be a list of requirements to be enforced by the contract defined by `type` and `version` keys. Currently, `type` may be either `sw.supervisor` for the device Supervisor version or `sw.l4t` for the L4T version.
+
+__Note__ If `name`, `type`, or `slug` are omitted, the build with fail.
```
xginn8:
```suggestion
__Note__ If `name`, `type`, or `slug` are omitted, the build will fail.
```

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+excerpt: A guide to using container contracts to enforce device compatibility
+---
+
+# Container contracts
+
+__Note__ Container contracts are available for devices running Supervisor versions >= 10.6.13. All prior versions will not enforce any contracts.
+
+Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the device is running a specific version of [{{ $names.company.lower }} Supervisor][supervisor] or has a specific version of the [NVIDIA Tegra Linux Driver Package][l4t] (L4T).
+
+Each service container can define the contracts that it enforces. If a container's contract requirements are not met, the release is not deployed to the device, unless it is marked as [optional](#optional-containers). You only need to define a contract for service containers that you wish to enforce a contract upon. Currently, you may define contract requirements based on the versions of Supervisor and L4T.
```
xginn8:
```suggestion
Each service container can define the contracts that it requires. If a container's contract requirements are not met, the release is not deployed to the device, unless it is marked as [optional](#optional-containers). You only need to define a contract for service containers that you wish to enforce a contract upon. Currently, you may define contract requirements based on the versions of Supervisor and L4T.
```

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+```shell
+Could not move to new release: Some releases were rejected due to having unmet requirements:
+Contracts: Services with unmet requirements: first-service
+```
+
+__Note__ For application fleets that contain a mixture of Supervisor and L4T versions, contracts may be used to ensure only compatible devices run specific services as releases will only be deployed to those meeting the contract requirements.
+
+## Optional containers
+
+By default, when a service contract fails, none of the services are deployed to the device. However, in a multi-container application, it is possible to ignore those services that fail the contract requirements with the other services being deployed as normal. To do so, we make use of the `io.balena.features.optional=1` [Supervisor label][labels] to indicate which services should be considered optional.
```
xginn8:
```suggestion
By default, when a service contract fails, none of the services are deployed to the device. However, in a multi-container application, it is possible to ignore those services that fail the contract requirements with the other services being deployed as normal. To do so, we make use of the `io.balena.features.optional: 1` [Supervisor label][labels] to indicate which services should be considered optional.
```

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+
+In the `docker-compose.yml` file of the application, add the `io.balena.features.optional=1` to the labels list for each service you wish to mark as optional. In the following example, even if the `first-service` service contract fails, the `second-service` service will still be deployed (assuming it doesn’t have any failing contracts of its own).
```
xginn8:
```suggestion
In the `docker-compose.yml` file of the application, add the `io.balena.features.optional: 1` to the labels list for each service you wish to mark as optional. In the following example, even if the `first-service` service contract fails, the `second-service` service will still be deployed (assuming it doesn’t have any failing contracts of its own).
```

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+
+```Dockerfile
+version: '2'
+services:
+ first-service:
+ build: ./first-service
+ labels:
+ - io.balena.features.optional=1
```
xginn8:
```suggestion
 - io.balena.features.optional: 1
```

pages/learn/develop/container-contracts.md
garethtdavies: I think both styles work but will update for consistency with labels in the docs 👍

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+
+Multiple requirements may be added to a single contract. The following example requires that the Supervisor version is greater than or equal to 10.6.13, and the L4T version is equal to 32.2.
+
+```yaml
+type: 'sw.container'
+slug: 'enforce-supervisor-and-l4t'
+name: 'Enforce supervisor and l4t requirements'
+requires:
+ - type: 'sw.supervisor'
+ version: '>=10.6.17'
```
CameronDiver: I believe version needs to be at the same indentation as the `type:` string, otherwise this yaml would fail to be parsed

garethtdavies: Thanks for the catch. Corrected indentation.

garethtdavies:
@balena-ci retest

garethtdavies:
@CameronDiver Given that 2.46 is being released with a compatible supervisor is this PR good to go now? It's been a while since this was written so also checking nothing else changed in interim :)

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+excerpt: A guide to using container contracts to enforce device compatibility
+---
+
+# Container contracts
+
+**Note** Container contracts are available for devices running Supervisor versions >= 10.6.17. All prior versions will not enforce any contracts.
+
+Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the device is running a specific version of [{{ $names.company.lower }} Supervisor][supervisor] or has a specific version of the [NVIDIA Tegra Linux Driver Package][l4t] (L4T).
+
+Each service container can define the contracts that it requires. If a container's contract requirements are not met, the release is not deployed to the device, unless it is marked as [optional](#optional-containers). You only need to define a contract for service containers that you wish to enforce a contract upon. Currently, you may define contract requirements based on the versions of Supervisor and L4T.
```
lurch: Is "service container" a common phrase in the Balena world?

"Currently, you may define contract requirements based on the versions of Supervisor and L4T."

Presumably that list will be expanded in future? Is it worth mentioning that?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
+---
+title: Container contracts
+excerpt: A guide to using container contracts to enforce device compatibility
+---
+
+# Container contracts
+
+**Note** Container contracts are available for devices running Supervisor versions >= 10.6.17. All prior versions will not enforce any contracts.
+
+Container contracts are used to ensure the compatibility of a device to run an individual service of an application. For example, container contracts may be used to ensure that the device is running a specific version of [{{ $names.company.lower }} Supervisor][supervisor] or has a specific version of the [NVIDIA Tegra Linux Driver Package][l4t] (L4T).
```
lurch: For somebody outside of the Balena world, "compatibility of a device" implies hardware compatibility to me, but I guess you're actually talking about software components compatibility?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+  - type: "sw.l4t"
+    version: "28.2"
+```
+
+- `type` should be set to `sw.container` for container contracts.
+- `slug` should be a unique identifier to identify the contract.
+- `name` should be a descriptive name of the contract.
+- `requires` should be a list of requirements to be enforced by the contract defined by `type` and `version` keys. Currently, `type` may be either `sw.supervisor` for the device Supervisor version or `sw.l4t` for the L4T version.
+
+**Note** If `name`, `type`, or `slug` are omitted, the build will fail.
```
lurch: Minor nitpick, but I wonder if it'd be worth always referring to "`type`, `slug`, `name`" in the same order?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+
+Multiple requirements may be added to a single contract. The following example requires that the Supervisor version is greater than or equal to 10.6.13, and the L4T version is equal to 32.2.
```
lurch: Says 10.6.13 here, but then 10.6.17 in the example below? :eyes:

pages/learn/develop/container-contracts.md
lurch: "When deploying a release, if the contract requirements are not met, the release will fail, and all devices in the application will remain on their current release."

Based on @CameronDiver 's comment about it being verified on-device, I guess it's actually "all devices that don't fulfil the contract will remain on their current release", rather than "all devices in the application" ? :man_shrugging: 

Also, it's not clear what "This label specifies which services with unmet requirements will be ignored, with the release being deployed to all other devices in the application." means.

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+
+**Note** Any type other than `sw.supervisor` or `sw.l4t` will result in the contract being invalid.
+
+When deploying a release, if the contract requirements are not met, the release will fail, and all devices in the application will remain on their current release. This default behavior can be overridden by including an [optional label](#optional-containers) in the application docker-compose.yml file. This label specifies which services with unmet requirements will be ignored, with the release being deployed to all other devices in the application. If there are any existing running containers for services with unmet requirements, they will not be destroyed and continue running the previous release.
+
+Should a contract fail, the dashboard logs will contain detail about the failing contract:
+
+```shell
+Could not move to new release: Some releases were rejected due to having unmet requirements:
+Contracts: Services with unmet requirements: first-service
```
lurch: I wonder if it'd also be worth mentioning the contract slug in these error logs?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+**Note** For application fleets that contain a mixture of Supervisor and L4T versions, contracts may be used to ensure only compatible devices run specific services as releases will only be deployed to those meeting the contract requirements.
```
lurch: NIT: Maybe add a comma between "services" and "as"?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+services:
+  first-service:
+    build: ./first-service
+    labels:
+      - io.balena.features.optional: 1
+  second-service:
+    build: ./second-service
+```
+
+**Note** When using optional containers, if there are any existing running services of the failed contract, they will be killed and not continue running the old version.
```
lurch: Does the optional service get killed, and the optional container left at the old version; or does the optional service get killed, with the optional container being updated but not started?

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Board support
+
+Container contracts are supported on all devices where the Supervisor version is >= 10.6.17. If a device in the application does not support L4T the device supervisor will reject the release unless the contract is marked as optional. L4T support is available on the following boards:
```
lurch: "If a device in the application does not support L4T the device supervisor will reject the release unless the contract is marked as optional."

Presumably this is only if the contract itself specifies an L4T requirement? The sentence as-is could be misread that the supervisor will reject all contracts on all devices that don't support L4T :wink:

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+├── docker-compose.yml
+├── first-service
+│ ├── contract.yml
+│ └── Dockerfile.template
+└── second-service
+  ├── contract.yml
+  └── Dockerfile.template
+```
+
+`contract.yml` is a YAML file containing `type`, `slug`, `name`, and `requires` keys, which define the requirements of the contract. The following example requires the device to be running a version of L4T equal to 28.2 for the contract to pass. For L4T exact versions (either 28.2 or 32.2) must be specified.
```
lurch: Could a use-case for optional containers be having one with a contract that depends on L4T 28.2 and one with a contract depending on L4T 32.2, and so therefore you'd always have one of the optional containers running on the device regardless of which version of L4T is installed? I know nothing about L4T... :laughing:

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+slug: "enforce-supervisor-and-l4t"
+name: "Enforce supervisor and l4t requirements"
+requires:
+  - type: "sw.supervisor"
+    version: ">=10.6.17"
+  - type: "sw.l4t"
+    version: "32.2"
+```
+
+**Note** Any type other than `sw.supervisor` or `sw.l4t` will result in the contract being invalid.
```
lurch: I wonder if there's an easy way of rewording this to indicate that it only applies to the requires-`type` and not the top-level-`type`? :man_shrugging:

pages/learn/develop/container-contracts.md
garethtdavies: I don't think so. L4T support is baked into the hostOS (https://github.com/balena-os/meta-tegra) so it's at the device level and not the container.

pages/learn/develop/container-contracts.md
garethtdavies: Service is synonymous to container so the service (container) is killed and so there is nothing to run. You noted earlier the term "service container" that I introduced which I have since tidied to make this more clear.

pages/learn/develop/container-contracts.md
garethtdavies: I suppose it's a bit of both given L4T can only run on certain devices but point duly noted. I've left as is for now.

pages/learn/develop/container-contracts.md
garethtdavies: As noted below I've made this more consistent using services. I think anything planned for the future can be left out of docs for now and would likely be covered in a blog or similar.

pages/learn/develop/container-contracts.md
lurch: Yeah, I guess I didn't explain that very well.
I meant if you had deviceA and deviceB with L4T 28.2 installed (in the hostOS) and deviceC and deviceD with L4T 32.2 installed, could you have optional container1 with a contract requiring L4T 28.2 and optional container2 with a contract requiring L4T 32.2; and then you could be sure that all devices would be using one-or-the-other of the optional containers?
(perhaps that's still not clear as I'm not very familiar with the current Balena terminology)

pages/learn/develop/container-contracts.md
```diff
@@ -57,9 +57,9 @@ requires:
...
 
-When deploying a release, if the contract requirements are not met, the release will fail, and all devices in the application will remain on their current release. This default behavior can be overridden by including an [optional label](#optional-containers) in the application docker-compose.yml file. This label specifies which services with unmet requirements will be ignored, with the release being deployed to all other devices in the application. If there are any existing running containers for services with unmet requirements, they will not be destroyed and continue running the previous release.
+When deploying a release, if the contract requirements are not met, the release will fail, and all services on the device will remain on thier current release. This default behavior can be overridden by including an [optional label](#optional-containers) in the application docker-compose.yml file. This optional label specifies which services with unmet requirements will be ignored, with all other services on the device updating to the new release. If there are any existing running services with unmet requirements, they will not be destroyed and continue running the prior release.
```
lurch: thier -> their

pages/learn/develop/container-contracts.md
```diff
@@ -0,0 +1,112 @@
...
+slug: "enforce-supervisor-and-l4t"
+name: "Enforce supervisor and l4t requirements"
+requires:
+  - type: "sw.supervisor"
+    version: ">=10.6.17"
+  - type: "sw.l4t"
+    version: "32.2"
+```
+
+**Note** Any `requires` type other than `sw.supervisor` or `sw.l4t` will result in the contract being invalid.
```
lurch: Will the contract also be invalid if you try to specify e.g.
```yaml
requires:
  - type: "sw.l4t"
    version: ">=32.2"
```
(since you said earlier that "For L4T exact versions (either 28.2 or 32.2) must be specified.")

Does an "invalid" contract get ignored, or does it get classed as a failing contract (i.e. as if it had unmet requirements) ?

pages/learn/develop/container-contracts.md
lurch: Yeah, reads much nicer now IMHO.

pages/learn/develop/container-contracts.md
garethtdavies: That example leads to `Not running containers because of contract violations: contract`

garethtdavies:
@balena-ci retest

PR #1211, opened 2019-10-29
Allow improve this doc for external repos

garethtdavies:
Change-type: patch
Signed-off-by: Gareth Davies gareth@balena.io

PR #1224, opened 2019-11-13
diagnostics: pull in latest docs automatically

xginn8:
Connects-to: https://github.com/balena-io/device-diagnostics/issues/110
Change-type: minor


xginn8:
> Looks great. I tested locally and am seeing the correct latest `DIAGNOSE_VERSION=4.9.2` vs `DIAGNOSE_VERSION=4.8.0` which is currently live on the docs.

@garethtdavies the only tricky part here is that these docs go "live" in the repo _before_ necessarily being deployed to balenaCloud. I still think it's better to autoupdate them, but maybe this needs more careful consideration to tightly couple the two.

garethtdavies:
Right, we could also version the `fetch-external.sh` script rather than pulling from master though I'm not aware of any way we could (automatically) update that based off balenaCloud deployments.

We also have the flip side where the docs are not updated even on an upstream change (unless there is a docs release), though that seems easier to resolve if we can find a suitable trigger.

garethtdavies:
@xginn8 also, see steps 2-3 here which I missed in my review 😕 https://github.com/balena-io/docs#adding-external-documentation which are required for the edit link to populate correctly. This was implemented for those other external repos via this PR https://github.com/balena-io/docs/pull/1211/files

Given we have similar issues currently, I'd be in favor of either merging as is (with the above steps added) or manually adding the docs versions to the `fetch-external.sh` script. I don't have any sense of how much current master could diverge from currently deployed balenaCloud, but if meaningful perhaps fixing to a version is the best approach which at least resolves the single source of truth issue.

xginn8:
> @xginn8 also, see steps 2-3 here which I missed in my review confused https://github.com/balena-io/docs#adding-external-documentation which are required for the edit link to populate correctly. This was implemented for those other external repos via this PR https://github.com/balena-io/docs/pull/1211/files
> 
> Given we have similar issues currently, I'd be in favor of either merging as is (with the above steps added) or manually adding the docs versions to the `fetch-external.sh` script. I don't have any sense of how much current master could diverge from currently deployed balenaCloud, but if meaningful perhaps fixing to a version is the best approach which at least resolves the single source of truth issue.

Thanks for pointing those out, sorry I missed them the first time around! I will merge as-is, I don't think it's a show-stopper but just wanted to point it out.

PR #1229, opened 2019-11-17
Remove outdated section about systemd

markspanbroek:
Systemd is no longer supported in the base images, see: https://www.balena.io/docs/reference/base-images/base-images/#major-changes

### DIFF
--- a/pages/learn/develop/dockerfile.md
+++ b/pages/learn/develop/dockerfile.md
@@ -115,52 +115,6 @@ As an example, let's say you have two Dockerfiles available, `Dockerfile.raspber
 
 Note that this feature works with the following commands: `git push`, `balena push`, `balena build`, and `balena deploy`.
 
-## Init system
-
-### Enable the init system
-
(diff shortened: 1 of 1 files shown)


garethtdavies:
Thanks for the contribution @markspanbroek 

@shaunmulligan, while this section is clearly outdated, do we want to keep any of it with a note it is only applicable to certain versions? Even if we keep some of it this likely isn't the best place for it and suggest moving from here.

shaunmulligan:
@garethtdavies yeah it might be good to keep this stuff around. We could just mention that this functionality is only supported on the discontinued `resin/` set of base images

garethtdavies:
There is also this for the legacy resin images https://www.balena.io/docs/reference/base-images/legacy-base-images/#init-system so I'm in favor of simply stripping this to avoid confusion.

garethtdavies:
@balena-ci retest

garethtdavies:
@balena-ci retest

Issue #1124, opened 2019-06-28
Document /v5/release_tag API


UncleSamSwiss:
The (newish) release tag API is missing from the resource documentation:
https://www.balena.io/docs/reference/api/resources/release/

See the following issue for the changes:
https://github.com/balena-io/balena-sdk/issues/622

garethtdavies:
This is resolved by this pending PR: https://github.com/balena-io/docs/pull/1232

PR #1232, opened 2019-11-19
docs: Bumping API version to v5

garethtdavies:
Closes: #1124 

Change-type: patch

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -9,7 +9,7 @@
       "depends_on__application",
       "actor",
       "app_name",
-      "git_repository",
+      "slug",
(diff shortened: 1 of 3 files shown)


garethtdavies:
Judging from this changelog v5 is the latest https://github.com/balena-io/balena-api/blob/master/CHANGELOG.md#v910---2018-10-29

While I tested a number of the endpoints in the API docs I'm unsure if anything else changed v4->v5 that needs updating.

garethtdavies:
Looks like there is an undocumented `release_tag` resource too.

garethtdavies:
@balena-ci retest

PR #1233, opened 2019-11-20
Update deployment.md

dtischler:
Remove link to non-existent anchor.


garethtdavies:
Thanks @dtischler. The commit needs a footer in order to pass CI, you can see more in the contribution guidelines here: https://github.com/balena-io/docs/blob/master/CONTRIBUTING.md but essentially something like:

```
docs: Removing broken anchor

Change-type: patch
Signed-off-by: Joe Soap <joe.soap@example.com>
```

garethtdavies:
I've opened https://github.com/balena-io/docs/issues/1234 as while SSH docs don't need to be in the Getting Started guide they should exist somewhere.

PR #1239, opened 2019-11-27
docs: Refreshing SSH documentation

### LINKED ISSUE #1147: balena cli documentation for balena ssh needs to be updated
https://www.balena.io/docs/learn/manage/ssh-access/#using-balena-ssh-from-the-cli is not up to date. balena ssh does not support the -s option any more and instead defaults to connecting to the host os. To connect to a container is service name has to be specified.

### LINKED ISSUE #1234: Add SSH key setup to manage section
This was stripped from the Getting Started Guide but should be present in the Manage section of the docs.

### LINKED ISSUE #1238: Mention SSH key is required for `balena ssh`
I'll link a support ticket where it was not clear that the use of the `balena ssh` CLI command required that an SSH key be set up on the user's account. It doesn't mention that it's required here: https://www.balena.io/docs/learn/manage/ssh-access/#using-balena-ssh-from-the-cli

However it is mentioned here: https://github.com/balena-io/balena-cli/blob/master/INSTALL.md#configuring-ssh-keys

### LINKED ISSUE #1263: Balena SSH CLI Docs
Looking at this page:  https://www.balena.io/docs/learn/manage/ssh-access/#using-balena-ssh-from-the-cli

The default behavior of the balena ssh command has been modified slightly, and documentation needs to be updated to reflect this.  There is no need to add the '--host' or '-s' flags anymore, as the default action will take you into the HostOS.  All that is needed to SSH to the host now is thus "balena ssh <device-uuid>".

In order to SSH into a container, that would be "balena ssh <device-uuid> <container-name>"

garethtdavies:
Closes #1263 
Closes #1238 
Closes #1234 
Closes #1147 

Change-type: patch


garethtdavies:
I've refreshed these docs to add the required SSH key info.

I’ve no idea if this section (Install an SSH server in a container) is still relevant https://github.com/balena-io/docs/pull/1239/files#diff-e7009d4cdf42562438e902abe2e20aabR49 perhaps @shaunmulligan can comment. There’s a link to a gitter discussion so I imagine a lot of this is very dated.

garethtdavies:
There is also this issue: https://github.com/balena-io/docs/issues/1116 specifically:

`Talk through adding sshKeys field to config.json for ssh in production devices.`

I'm documenting `config.json` elsewhere in a separate PR but to confirm if I add an SSH key via the control panel this also works for a production image?

pdcastro:
That page is only for `balena ssh`, right? I wonder if any of the following forum comments would be relevant here:
* https://forums.balena.io/t/rsync-over-balena-ssh-tunnel/6228/4
* https://forums.balena.io/t/howto-ssh-into-host-device/5312

garethtdavies:
Thanks for those links @pdcastro. I've pushed an update based on this and some additional testing.

I've updated this PR with standalone SSH commands, and I've also included details of tunneling the connection if the port is blocked (from your forum post).

I think the rsync/scp material makes for a great article in a "how to section" (that doesn't yet exist :) ) that references these docs.

I'm not clear on the second link as this seems more related to openBalena? Everything seems to be possible just via a combination of balenaCLI, standalone SSH, balena tunnel, and updating config.json so I'm not sure what else in this article would be required?

Anyway, I've updated this PR and would love more feedback as it already touches a _lot_ of open issues so clearly a point of some confusion. There is an intentionally broken link in this doc to updating config.json as this page doesn't exist, but there is an open PR for that here https://github.com/balena-io/docs/pull/1249.

Finally, I've removed the "Install an SSH server in a container" section, as I think this is redundant as everything seems possible to do with all the approaches already detailed in this PR. If I've misunderstood this, then someone, please correct me!

shared/general/container-ssh.md
```diff
@@ -1,46 +1,83 @@
...
-`<device-uuid>` is the unique identifier for the device you want to access, which can be found on the dashboard.
+`<device-uuid>` is the unique identifier for the device you want to access, which can be found via the dashboard or in the output of the `{{ $names.company.short }} devices` CLI command. By default, SSH access is routed into the host OS shell. However, you can SSH into a service by specifying its name as part of the command:
+
+```shell
+$ {{ $names.company.short }} ssh <device-uuid> main
+```
+
+This also works in multicontainer applications, simply pass the name of the appropriate service as defined in docker-compose.yml you'd like to access the shell for.
+
+For devices running development images on your local network, you can create a direct SSH connection by specifying the hostname using the `<device-uuid>.local` syntax or local IP address of the device. Development images have passwordless root SSH access enabled by default, and so they should never be directly exposed to the public internet. Using {{ $names.company.short }} ssh in this way does not use the balena VPN and instead makes a direct SSH connection to the device. For example:
```
pdcastro: > For devices running development images on your local network, you can create a direct SSH connection

I found this a bit confusing, because a "direct SSH connection" (meaning a connection that does not rely on the balenaCloud VPN) can be used with both development images and production images, and it is also not limited to local IP addresses -- a globally routable IP address would work too, although uncommon. I suggest (just a suggestion) reordering and slightly rewording this paragraph and the following two, as follows:

---
> ...
> When an application name or device UUID is used as above, `{{ $names.company.short }} ssh` uses the {{ $names.company.short }} VPN to create a secure tunnel to the device and then forward SSH traffic between the device and your development machine. For this method, you must have a SSH key configured on your development machine and [added to the {{ $names.cloud.lower }} dashboard][add-ssh-key].
>
> If an IP address or a `.local` hostname is used (instead of an application name or device UUID), `{{ $names.company.short }} ssh` establishes a direct connection that does not rely on the {{ $names.company.short }} VPN:
>
```shell
$ balena ssh 192.168.1.23
$ balena ssh <device-uuid>.local
```
>
> This should work without further configuration in the case of devices running a [development balenaOS image](https://www.balena.io/docs/reference/OS/overview/2.x/#dev-vs-prod-images), which allows passwordless root SSH access (and for this reason, should never be directly exposed to the public internet). In the case of a production balenaOS image, an SSH key needs to be present in the `config.json` file of the device. More details about configuring SSH keys in `config.json` may be found [here][config-json-ssh].

Above I have also replaced "development image" with "development balenaOS image", and made it a link.

shared/general/container-ssh.md
```diff
@@ -1,46 +1,83 @@
...
 
 __Note:__ Host OS SSH access is available for devices running {{ $names.os.lower }} version 2.7.5 and above.
 
 ## Using the dashboard web terminal
 
 To use this feature, navigate to your application and select the device you want to access. You will see a *Terminal* window below the *Logs* window:
 
-<img src="/img/common/device/terminal.png" width="60%">
-
-In order to start a terminal session for a service, you first need to ensure that your device is online and that the service container is running. If the container code crashes or ends quickly, it is not possible to attach a console to it. Check out our [Dockerfile][dockerfile-init] and [Docker Compose][compose-init] guides for more information on keeping your containers running.
```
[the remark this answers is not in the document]
garethtdavies: Right, this was an over-aggressive purge of outdated init links! Reworded this and avoided another note :) I actually had this issue testing this as many of our hello world projects simply print "hello world" and exit rather than looping/sleeping.

shared/general/container-ssh.md
garethtdavies: Thank you, this is excellent feedback and I have updated this.

garethtdavies:
@balena-ci retest

PR #1245, opened 2019-12-03
docs: Deltas enabled by default

garethtdavies:
Change-type: patch

### DIFF
--- a/pages/learn/deploy/delta.md
+++ b/pages/learn/deploy/delta.md
@@ -6,32 +6,36 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
 
 # Delta updates
 
-When a new version of your application is pushed to {{ $names.company.lower }} servers, your devices are notified, and will initiate an update of their running containers. The regular device update uses the Docker pull mechanism. It requests all the [layers](https://docs.docker.com/engine/userguide/storagedriver/imagesandcontainers/#/images-and-layers) of the new container image that are not present on the device (i.e. not shared with the previous image). Then from these layers the new image is assembled on the device, and replaced the previous version of the application. This process potentially moves a lot of data, uses a lot of space on the device to hold both the old and the new images, and the device can be in a sensitive state while Docker is updating (e.g. in case there's an unexpected power outage during that time).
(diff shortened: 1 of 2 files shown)


garethtdavies:
@shaunmulligan a very quick update to deltas by default. Is there anything additional you would like me to cover?

pages/learn/deploy/delta.md
```diff
@@ -6,29 +6,33 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
 
 Once the delta (difference between the old and new image) is calculated, the device downloads and applies this delta onto the old application image, in place. When the process is finished, the new container image (new version of the application) is started.
 
 These binary deltas save on the amount of data needed to be downloaded, reduce the storage space requirements on the device to perform an application update, and shorten the time when Docker is updating.
 
 ## Enabling delta updates
 
-Delta update behaviour is enabled with the `RESIN_SUPERVISOR_DELTA` configuration variable:
+For newly created applications, the delta update behavior is enabled by default. For existing applications, the delta update behavior is enabled with the `RESIN_SUPERVISOR_DELTA` configuration variable:
```
shaunmulligan: They might need to also add `RESIN_SUPERVISOR_DELTA_VERSION` == 3, to enable the V3 version of the deltas. @dfunckt can confirm

dfunckt: Users can't choose delta versions anymore @shaunmulligan -- that config var is now inaccessible by users.

pages/learn/deploy/delta.md
```diff
@@ -6,29 +6,33 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
 
 Once the delta (difference between the old and new image) is calculated, the device downloads and applies this delta onto the old application image, in place. When the process is finished, the new container image (new version of the application) is started.
 
 These binary deltas save on the amount of data needed to be downloaded, reduce the storage space requirements on the device to perform an application update, and shorten the time when Docker is updating.
 
 ## Enabling delta updates
```
dfunckt: I think it makes sense to flip the perspective of this document to assume the default behaviour (i.e. deltas are enabled by default for new apps) and describe how to *disable* them instead. I also think that note that's been added to the top of the document fits better here.

dfunckt:
I think we also need to update https://www.balena.io/docs/learn/welcome/primer/#getting-to-the-devices where it reads "To reduce download time and bandwidth consumption, you can also enable delta updates."

pages/learn/deploy/delta.md
garethtdavies: Agreed and I've moved the note here. I've updated the title to reflect this covers enabling/disabling.  

I've left if as enabling as the focus as for me the documentation says it's enabled by default but then here's how to enable it if you don't have it... Then once everyone is at the same state i.e. it's enabled, there are instructions for disabling (which in this instance are essentially the same).  I tried this the other way and don't think it flows well, particularly as I imagine a large % do not have this enabled already.

dfunckt: Sounds good, agreed.

garethtdavies:
This is already in production right @dfunckt so this can go live?

dfunckt:
@garethtdavies yep

garethtdavies:
@balena-ci retest

Issue #859, opened 2018-08-01
Document valid config.json fields


lekkas:
Currently config.json configuration options are scattered in our docs, we should probably look into having a single page / manual for it

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_e8m3)

pcarranzav:
There's some documentation in https://github.com/resin-os/meta-resin#configjson - but indeed, having this centralized in the docs would be great.

pdcastro:
Like `pcarranzav` mentioned, frequently enough I send customers (as a support agent) a URL to the `meta-balena` [Readme](https://github.com/balena-os/meta-balena#configjson) as documentation for fields in the `config.json` file, but customer feedback is that they would never have thought of looking in that place (most users may not even understand the relevance of the `meta-balena` repo).

I suggest going a bit beyond documenting the "valid fields", but also explaining for example:

* That on a device, the right file to look at is `/mnt/boot/config.json` rather than `/resin-boot/config.json`. (See also this [OS masterclass](https://gist.github.com/ZubairLK/c0f8918e983e94edc74e2e956143644d#configjson).)
  * And that the `jq` tool can be used on a host OS prompt to see the contents in a human-readable way: `jq . < /mnt/boot/config.json`.
* The recommended way of editing `config.json` -- e.g. via the web dashboard rather than manually, or using a script / tool (cc: @ZubairLK).
* Perhaps briefly mention that it's a completely different file to `config.txt`, [documented elsewhere](https://www.balena.io/docs/reference/OS/advanced/#configtxt).
* Documentation of each field, with examples of the syntax, and an example of a full file.
  * For example, the `meta-balena` Readme includes a partial example for the [sshKeys](https://github.com/balena-os/meta-balena#sshkeys) field, but it could be expanded to mention that the key could be found in the balena web dashboard, and to include an example of a key (e.g. to show that it starts with `ssh-rsa`):
```
"os": {
    "sshKeys": [
      "ssh-rsa AAAAB3Nza...M2JB paulo@macbook-pro",
      "ssh-rsa AAAAB3Nza...nFTQ paulo@zenbook"
    ]
  }
```

samuelbacaner:
@pdcastro I'm glad I found this ticket so I could find that Readme. Can you elaborate briefly on the recommended way of editing `config.json`? Where in the web dashboard can you edit `config.json`?

[...]

PR #1249, opened 2019-12-04
docs: Adding config.json documentation

garethtdavies:
Change-type: patch
Connects-to: #859

### DIFF
--- a/pages/reference/OS/configuration.md
+++ b/pages/reference/OS/configuration.md
@@ -1,0 +1,76 @@
+---
+title: Configuration
+excerpt: Instructions for editing and customizing config.json for modifying the behavior of {{ $names.os.lower }}
+---
+
+# Configuring {{ $names.os.lower }}
+
+The behavior of {{ $names.os.lower }} can be configured by editing the `config.json` file. This file is located in the [boot partition][boot-partition] and accepts a [range of fields](#valid-fields) to modify the behavior of the host OS.
+
+**Note:** The `config.json` file is different from the [`config.txt`][config-txt] file, also located in the boot partition, which is used by the Raspberry Pi to set device configuration options.
+
+Before the device is [provisioned][provisioned], you may edit `config.json` by mounting a flashed SD card (with the partition label `resin-boot`) and editing the file directly. The boot partition is mounted on the device at `/mnt/boot`, and so on the device, the file is located at `/mnt/boot/config.json`. For example, to output a formatted version of `config.json` on a device, use the following commands:
+
+```shell
+{{ $names.company.lower }} ssh <uuid>
+cat /mnt/boot/config.json | jq '.'
+```
+
+**Warning:** Editing config.json on a provisioned device should be done very carefully as any mistakes in the syntax of this file can leave a device inaccessible.
+
+After provisioning, editing `config.json` as described above is not reliable or advisable because the [supervisor][supervisor] may overwrite certain fields, such as `persistentLogging`, with values read from the {{ $names.cloud.lower }} API. To safely modify the values of `config.json` on a provisioned device use one of the following methods:
+
+- Update the device [hostname](#hostname) via the [supervisor API][hostname].
+- Modify the [persistent logging](#persistentlogging) setting via the device dashboard [configuration variables][configuration].
+- Apply `config.json` updates remotely via the {{ $names.cli }} using the [configizer project][configizer].
+
+Alternatively, you can always reprovision a device with an updated `config.json` file.
+
+**Note:** To manually edit `config.json` on a provisioned device, see the [{{ $names.os.lower }} masterclass instructions][masterclass-os].
+
+## Sample config.json
+
+The following example provides all customizable configuration options available in `config.json`. Full details about each option may be found in the [valid fields](#valid-fields) section.
+
+```json
+{
+  "hostname": "my-custom-hostname",
+  "persistentLogging": true,
+  "country": "GB",
+  "ntpServers": "ntp-wwv.nist.gov resinio.pool.ntp.org",
+  "dnsServers": "208.67.222.222 8.8.8.8",
+  "os": {
+    "network": {
+      "connectivity": {
+        "uri": "https://api.balena-cloud.com/connectivity-check",
+        "interval": "300",
+        "response": "optional value in the response"
+      },
+      "wifi": {
+        "randomMacAddressScan": false
+      }
+    },
+    "udevRules": {
+      "56": "ENV{ID_FS_LABEL_ENC}==\"resin-root*\", IMPORT{program}=\"resin_update_state_probe $devnode\", SYMLINK+=\"disk/by-state/$env{RESIN_UPDATE_STATE}\"",
+      "64": "ACTION!=\"add|change\", GOTO=\"modeswitch_rules_end\"\nKERNEL==\"ttyACM*\", ATTRS{idVendor}==\"1546\", ATTRS{idProduct}==\"1146\", TAG+=\"systemd\", ENV{SYSTEMD_WANTS}=\"u-blox-switch@'%E{DEVNAME}'.service\"\nLBEL=\"modeswitch_rules_end\"\n"
+    },
+    "sshKeys": [
+      "ssh-rsa AAAAB3Nza...M2JB balena@macbook-pro",
+      "ssh-rsa AAAAB3Nza...nFTQ balena@zenbook"
+    ]
+  }
+}
+```
+
+{{> "general/config-json-fields"}}
+
+[boot-partition]: /reference/OS/overview/2.x/#stateless-and-read-only-rootfs
+[config-txt]: /reference/OS/advanced/#configtxt
+[country-codes]: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
+[hostname]: /reference/supervisor/supervisor-api/#patch-v1devicehost-config
+[configuration]: /learn/manage/configuration/
+[nm-connectivity]: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
+[configizer]:{{ $links.githubPlayground }}/configizer
+[masterclass-os]:{{ $links.githubMain }}/balenaos-masterclass#12-advanced-editing-configjson-by-hand
+[provisioned]:/learn/welcome/primer/#device-provisioning
+[supervisor]:/reference/supervisor/supervisor-api/
--- a/shared/general/config-json-fields.md
+++ b/shared/general/config-json-fields.md
@@ -1,0 +1,118 @@
+## Valid fields
+
+The following fields can be used to modify the behavior of the host OS.
+
+### hostname
+
+(string) The configured hostname of the device, otherwise the device UUID is used.
+
+### persistentLogging
+
+(boolean) Enable or disable persistent logging on the device - defaults to false.
+
+### country
+
+(string) [Two-letter country code][country-codes] for the country in which the device is operating. This is used for setting the WiFi regulatory domain, and you should check the WiFi device driver for a list of supported country codes.
+
+### ntpServers
+
+(string) A space-separated list of NTP servers to use for time synchronization. Defaults to `resinio.pool.ntp.org` servers:
+
+- `0.resinio.pool.ntp.org`
+- `1.resinio.pool.ntp.org`
+- `2.resinio.pool.ntp.org`
+- `3.resinio.pool.ntp.org`
+
+### dnsServers
+
+(string) A space-separated list of preferred DNS servers to use for name resolution.
+
+- When `dnsServers` is not defined, or empty, Google's DNS server (8.8.8.8) is added to the list of DNS servers obtained via DHCP or statically configured in a NetworkManager connection profile.
+- When `dnsServers` is "null" (a string), Google's DNS server (8.8.8.8) will NOT be added as described above.
+- When `dnsServers` is defined and not "null", the listed servers will be added to the list of servers obtained via DCHP or statically configured via a NetworkManager connection profile.
+
+### os
+
+An object containing settings that customize the host OS at runtime.
+
+#### network
+
+##### wifi
+
+An object that defines the configuration related to Wi-Fi.
+
+- "randomMacAddressScan" (boolean) Configures MAC address randomization of a Wi-Fi device during scanning
+
+The following example disables MAC address randomization of Wi-Fi device during scanning:
+
+```json
+"os": {
+ "network" : {
+  "wifi": {
+    "randomMacAddressScan": false
+  }
+ }
+}
+```
+
+##### connectivity
+
+An object that defines configuration related to networking connectivity checks. This feature builds on NetworkManager's connectivity check, which is further documented in the connectivity section [here][nm-connectivity].
+
+- "uri" (string) Value of the url to query for connectivity checks. Defaults to `$API_ENDPOINT/connectivity-check`.
+- "interval" (string) Interval between connectivity checks in seconds. Defaults to 3600.
+- "response" (string). If set controls what body content is checked for when requesting the URI. If it is an empty value, the HTTP server is expected to answer with status code 204 or send no data.
+
+The following example configures the connectivity check by passing the {{ $names.cloud.lower }} connectivity endpoint with a 5-minute interval.
+
+```json
+"os": {
+ "network" : {
+  "connectivity": {
+    "uri" : "https://api.balena-cloud.com/connectivity-check",
+    "interval" : "300"
+  }
+ }
+}
+```
+
+#### udevRules
+
+An object containing one or more custom udev rules as `key:value` pairs.
+
+To turn a rule into a format that can be easily added to `config.json`, use the following command:
+
+```shell
+cat rulefilename | jq -sR .
+```
+
+For example:
+
+```shell
+root@resin:/etc/udev/rules.d# cat 64.rules | jq -sR .
+"ACTION!=\"add|change\", GOTO=\"modeswitch_rules_end\"\nKERNEL==\"ttyACM*\", ATTRS{idVendor}==\"1546\", ATTRS{idProduct}==\"1146\", TAG+=\"systemd\", ENV{SYSTEMD_WANTS}=\"u-blox-switch@'%E{DEVNAME}'.service\"\nLBEL=\"modeswitch_rules_end\"\n"
+```
+
+The following example contains two custom udev rules that will create `/etc/udev/rules.d/56.rules` and `/etc/udev/rules.d/64.rules`. The first time rules are added/modified, udevd will reload the rules and re-trigger.
+
+```json
+"os": {
+ "udevRules": {
+  "56": "ENV{ID_FS_LABEL_ENC}==\"resin-root*\", IMPORT{program}=\"resin_update_state_probe $devnode\", SYMLINK+=\"disk/by-state/$env{RESIN_UPDATE_STATE}\"",
+  "64" : "ACTION!=\"add|change\", GOTO=\"modeswitch_rules_end\"\nKERNEL==\"ttyACM*\", ATTRS{idVendor}==\"1546\", ATTRS{idProduct}==\"1146\", TAG+=\"systemd\", ENV{SYSTEMD_WANTS}=\"u-blox-switch@'%E{DEVNAME}'.service\"\nLBEL=\"modeswitch_rules_end\"\n"
+ }
+}
+```
+
+#### sshKeys
+
+(Array) An array of strings containing a list of public SSH keys that will be used by the SSH server for authentication.
+
+```json
+"os": {
+ "sshKeys": [
+  "ssh-rsa AAAAB3Nza...M2JB balena@macbook-pro",
+  "ssh-rsa AAAAB3Nza...nFTQ balena@zenbook"
+ ]
+}
+```
--- a/config/navigation.txt
+++ b/config/navigation.txt
@@ -98,6 +98,7 @@ Reference
       [/reference/OS/updates/update-process]
       [/reference/OS/updates/migrate-to-2.0]
     Network[/reference/OS/network/$os_version]
+    [/reference/OS/configuration]
     [/reference/OS/time]
     [/reference/OS/advanced]
     [/reference/OS/meta-balena]


pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+### ntpServers
+
+(string) A space-separated list of NTP servers to use for time synchronization. Defaults to `resinio.pool.ntp.org` servers.
+
+### dnsServers
+
+(string) A space-separated list of preferred DNS servers to use for name resolution.
+
+* When `dnsServers` is not defined, or empty, the default value of Google's DNS (8.8.8.8) is used.
+* When `dnsServers` is `"null"` (a string), the system will not alter the dnsServers to the default server as detailed above.
```
lurch: Does this mean there'd be no DNS resolution at all?

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+    },
+    "sshKeys": [
+      "ssh-rsa AAAAB3Nza...M2JB balena@macbook-pro",
+      "ssh-rsa AAAAB3Nza...nFTQ balena@zenbook"
+    ]
+  }
+}
+```
+
+## Valid fields
```
pdcastro: This section looks a lot like: https://github.com/balena-os/meta-balena#configjson
So there's a source of truth / sync issue: when someone merges a meta-balena PR that modifies the README, how does it get (automatically) propagated to the docs repo? We have an existing setup with scripts that copy md files from other repos - cli, supervisor, Node SDK... Should we setup something like that? Maybe the config.json docs could be moved out of the meta-balena README and into its own separate md file - also in the meta-balena repo?

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+
+* Using the supervisor API, you can update the device [hostname](/reference/supervisor/supervisor-api/#patch-v1devicehost-config).
+* Using the device dashboard [configuration variables](/learn/manage/configuration/), you may modify the persistent logging setting.
+* Using the [configizer project](https://github.com/balena-io-playground/configizer) you may apply updates remotely via the {{ $names.cli }}. Note that this has not been thoroughly tested and should be treated with caution. Always test any changes first on a development device.
+
+Alternatively, you can always reprovision a device with an updated `config.json` file.
+
+**Note:** Editing config.json on a provisioned device is not recommended, as mistakes to this file can leave a device inaccessible. If you need to manually edit this file, see the [balenaOS masterclass instructions](https://github.com/balena-io-projects/balenaos-masterclass#12-advanced-editing-configjson-by-hand).
```
pdcastro: This paragraph partially duplicates another paragraph above _("Special care should be taken...")_ Might be nicer to merge the 2 paragraphs, and perhaps emphasise a whole warning sentence in bold (if the reason to  duplicate the information was to make sure it is noticed...).

pages/reference/OS/configuration.md
pdcastro: No, DNS servers obtained via DHCP (and/or [mentioned in a NetworkManager connection profile](https://www.balena.io/docs/reference/OS/network/2.x/#setting-a-static-ip)?) are also used. Also, I think (hope!) that the following paragraph in the [Network Requirements](https://www.balena.io/docs/reference/OS/network/2.x/#network-requirements) docs is correct, and could be added / copied (?) to this section (or have it only here, and change the network requirements to link here...)

> Google's Public DNS server at 8.8.8.8 is used by default and in addition to DNS servers obtained via DHCP (balenaOS may issue queries to multiple DNS servers simultaneously, for the quickest response to be used). If additional DNS servers are configured via DHCP or other means, it is OK for the local network to block requests to 8.8.8.8. To avoid any requests being made to 8.8.8.8 by balenaOS, modify the dnsServers setting in config.json.

So maybe actually the three bullet points would be more informative if rewritten as follows, though this needs testing to confirm, or source code to be read, or expert advice from perhaps @majorz or @alexgg (Zubair is away for a while...)
* When `dnsServers` is not defined, or empty, Google's DNS server (8.8.8.8) is added to the list of DNS servers obtained via DHCP or statically configured in a NetworkManager connection profile.
* When `dnsServers` is `"null"` (a string), Google's DNS server (8.8.8.8) will NOT be added as described above.
* When `dnsServers` is defined and not `"null"`, the listed servers will _replace_ any servers obtained via DCHP or statically configured via a NetworkManager connection profile.

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+
+(boolean) Enable or disable persistent logging on the device. Defaults to false.
+
+### country
+
+(string) The country in which the device is operating. Used for setting the WiFi regulatory domain.
+
+### ntpServers
+
+(string) A space-separated list of NTP servers to use for time synchronization. Defaults to `resinio.pool.ntp.org` servers.
```
pdcastro: The [Network Requirements](https://www.balena.io/docs/reference/OS/network/2.x/#network-requirements) docs have a full list of NTP servers:

> NTP / UDP packets (port 123) are exchanged with:
> 0.resinio.pool.ntp.org
> 1.resinio.pool.ntp.org
> 2.resinio.pool.ntp.org
> 3.resinio.pool.ntp.org

But the Network Requirements docs fail to mention that the servers could be modified via `config.json`. So perhaps we could list the 4 servers above here, and modify the Network Requirements docs to point here for a list of servers.

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+
+(string) The configured hostname of the device, otherwise the device UUID is used.
+
+### persistentLogging
+
+(boolean) Enable or disable persistent logging on the device. Defaults to false.
+
+### country
+
+(string) The country in which the device is operating. Used for setting the WiFi regulatory domain.
```
pdcastro:
```
(string) Two-letter country code for the region the device is operating in.
Used for setting the WiFi regulatory domain.
```
I suspect not all country codes are supported and it's tempting to add something like _"Check the WiFi device driver for a list of supported country codes"_ and to have "Two-letter country code" be a link to [ISO 3166-1 alpha-2 on Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2), but I'm not sure it's correct... :-)

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+##### connectivity
+
+An object that defines configuration related to networking connectivity checks.
+
+* "uri" (string) Value of the url to query for connectivity checks. Defaults to `$API_ENDPOINT/connectivity-check`.
+* "interval" (string) Interval between connectivity checks in seconds. Defaults to 3600.
+* "response" (string). If set controls what body content is checked for when requesting the URI. If it is an empty value, the HTTP server is expected to answer with status code 204 or send no data.
+
```
pdcastro: Maybe we can add here something like:
> This feature builds on NetworkManager's connectivity check, which is further documented in the connectivity section at: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html

pages/reference/OS/configuration.md
garethtdavies: Yes, this is largely taken verbatim with a few fixes and that's a very good idea. I was thinking along those same lines by importing them as a partial. Let's complete this document in this repo for now (otherwise trying to coordinate reviews in two repos would be painful) and then once it is complete I'll move this section, to a new page in the meta-balena repo.

pages/reference/OS/configuration.md
garethtdavies: Agreed. I'm also not sure whether we should be recommending this or not - the masterclass also has its own fair share of warnings if we do link to it :)

pages/reference/OS/configuration.md
lurch: In case it has useful / relevant information: https://github.com/raspberrypi/documentation/blob/master/configuration/wireless/wireless-cli.md

pages/reference/OS/configuration.md
lurch: Is it worth mentioning (or linking to other docs) what happens if the connectivity check fails?

pages/reference/OS/configuration.md
alexgg: Hi @pdcastro, how I understand the DNS resolution in Balena is that it uses dnsmasq as a local cache and forwarder.
/etc/resolv.conf is symlinked to resolv.dnsmasq
and dnsmasq is started with:
`-r /etc/resolv.dnsmasq -z --servers-file=/run/dnsmasq.servers`
`/etc/resolv.dnsmasq` is a symlink to `/var/run/resolvconf/interface/NetworkManager`, which is the file that NetworkManager uses to store the DHCP or statically configured DNS server. This is the first place where dnsmasq lookes for upstream servers.
And `/run/dnsmasq.servers` is the file that is configured from config.json with either the user provided servers or 8.8.8.8.
So dnsServers only affects the contents of `/run/dnsmasq.servers`, and the NM DHCP or static configuration will always be there.
So I would rephase the above to:

- When `dnsServers` is not defined, or empty, Google's DNS server (8.8.8.8) is added to the list of DNS servers obtained via DHCP or statically configured in a NetworkManager connection profile.
- When `dnsServers` is "null" (a string), Google's DNS server (8.8.8.8) will NOT be added as described above.
- When `dnsServers` is defined and not "null", the listed servers will be added to the list of servers obtained via DCHP or statically configured via a NetworkManager connection profile.

pdcastro: Thank you for digging this @alexgg! 👍 We are finally getting more clarity on these points. It surprises me that `--all-servers` isn't the default behaviour, as I believe this is the "support knowledge" that we share with users. @majorz, are you similarly surprised? :-) 
In this case, we need to update the Network Requirements page as well, as it is incorrect.

pages/reference/OS/configuration.md
pdcastro: > I'm also not sure whether we should be recommending this or not 

**My two cents,** I think we should, yes, because:
* The masterclasses are public and were written with the intent of sharing with end users. We share them in support interactions as well.
* In forums or paid support, users often ask about editing `config.json` and we share this kind of advice anyway, though perhaps inconsistently between support agents. Having the information in a public documentation page support effort and improves advice consistency. And if users try it and it goes wrong in some step, we will know where to focus the improvement effort.

I wonder if @shaunmulligan thinks the same way... :-)

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,179 @@
...
+# Configuring {{ $names.os.lower }}
+
+The behavior of {{ $names.os.lower }} can be configured by editing the `config.json` file that resides in the boot partition of the device. This file accepts a [range of fields](#valid-fields) to modify the behavior of the host OS.
+
+__Note:__ The `config.json` file is different from the [`config.txt`](https://www.balena.io/docs/reference/OS/advanced/#configtxt) file also located in the boot partition, which is used by the Raspberry Pi to set device configuration options.
+
+## Using config.json to provision a device
+
+The `config.json` file is located in the root of the `resin-boot` partition, and you may edit this file by mounting a flashed SD card and editing the file directly. Editing this file will only work if you edit before the device's first boot. Once the device has been provisioned, edits to this file have no effect as the contents of the file are copied to the device boot partition at `/mnt/boot/config.json`.
```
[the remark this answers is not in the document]
pdcastro: @shaunmulligan or @alexgg might be able to help on this point as well...

pages/reference/OS/configuration.md
garethtdavies: There is some additional material here on this issue just published via @hedss WIP masterclass https://hackmd.io/7z1NBeDoTy2rqchIC9ogHg?view#52-DNS-Issues

pages/reference/OS/configuration.md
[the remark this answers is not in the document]
hedss: Yep, you can edit the `config.json` on-device, `/mnt/boot` is a mapping to the boot partition (first) of the media. We shouldn't really be mentioning `/resin-boot`, these are backups to ensure that should a HUP fail. Weirdly, as I started typing here, this thread: https://www.flowdock.com/app/rulemotion/user_happiness/threads/o2DA9mNhyUEafWsSUMauNi0t3se appeared. Worth reading. :)

pages/reference/OS/configuration.md
hedss: I agree we should share this advice, but we definitely want to really stress the point that if they make changes that makes the file unreadable, they're going to be in a position where they won't be able to access the device again. Saying that, if it's already provisioned and had connected to the VPN *and* there's another device online on the same network, we could probably salvage it. But that's not a support ticket we want to see :)

hedss: And actually, adding to this, if people do want to mess around with it, you want to do it in a safe way: https://github.com/balena-io-playground/configizer

garethtdavies:
OK, thanks, everyone. Updated this file to incorporate everything above. Apologies, I squashed so you can't see review changes but enough was changed to probably warrant a re-scan.

I've also done a lot more testing on this, not least as that `/resin-boot` directory threw me off. I was also confused as the supervisor will only overwrite settings on a provisioned device it manages - so you can't say change persistent logging on a provisioned device but you can add SSH keys / udev rules (basically anything that configizer allows). This issue is also interesting to make this behavior consistent https://github.com/balena-io/balena-supervisor/issues/998. 

As noted above, I'll arrange the content to avoid duplicating once the text is OK.

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,193 @@
...
+}
+```
+
+## Editing config.json on a provisioned device
+
+**Warning:** Editing config.json on a provisioned device should be done very carefully as any mistakes to this file can leave a device inaccessible.
+
+There are various ways of safely modifying the values of `config.json` on a provisioned device:
+
+- Using the supervisor API, you can update the device [hostname][hostname].
```
pdcastro:
```
Using the supervisor API, you can update the device [hostname and proxy settings][hostconfig]

[hostconfig]: /reference/supervisor/supervisor-api/#patch-v1devicehost-config
```

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,193 @@
+---
+title: Configuration
+excerpt: Instructions for editing and customizing config.json for modifying the behavior of {{ $names.os.lower }}
+---
+
+# Configuring {{ $names.os.lower }}
+
+The behavior of {{ $names.os.lower }} can be configured by editing the `config.json` file. This file is located in the [boot partition][boot-partition] and accepts a [range of fields](#valid-fields) to modify the behavior of the host OS.
+
+Before the device has been provisioned, you may edit `config.json` by mounting a flashed SD card (which will mount as `resin-boot`) and editing the file directly. The boot partition is mounted on the device at `/mnt/boot`, and so on the device, the file is located at `/mnt/boot/config.json`. For example, to output a formatted version of `config.json` on a device, use the following commands:
```
pdcastro: * has been provisioned → is provisioned
* This paragraph describes how to edit the `config.json` file ***before*** the device is provisioned. Then there is a whole section **"Sample config.json",** a long section **"Valid fields",** and then the section **"Editing config.json on a provisioned device".** I think this page would be better organised by having the complementary options of editing `config.json` _before and after_ provisioning in the same section, or in consecutive sections, rather than in "distant" sections.
* I am not sure how obvious the meaning of "provisioned" is. I suggest making the word "provisioned" a link to https://www.balena.io/docs/learn/welcome/primer/#device-provisioning

garethtdavies: Thanks again for the feedback. The distinction between before and after provisioning is _really_ confusing as the behavior is inconsistent. I did a lot of testing on this and determined the following behavior:

* Before or after provisioning you can edit _some_ values of config.json so long as the Supervisor doesn't manage them. I mounted an SD card from a provisioned device into my laptop and added the SSH keys or udev rules, for example, and that works just fine.
* When you edit `persistentLogging` before the device is provisioned this would be respected but if you edit this _after_ the device has been provisioned the Supervisor overwrites this setting - it actually reformats the file too (to minify the json). I suspect the same would be true of the hostname too (I'll test that later)

So the distinction between before and after provisioning is very muddled and has caveats, but I may just make this a single section and mention these.

pdcastro: @garethtdavies, what you've described above matches my previous "anecdotal" experience -- but I had not conducted an experiment (or read source code...) to allow me to state it. It really sounds correct though, and actually perhaps you could add a sentence in the lines of:
_"After provisioning, editing the `config.json` as described above is not reliable or advisable because the supervisor may overwrite some fields, such as persistentLogging, with values read from the balenaCloud (API) database."_
Just a thought. :-)

garethtdavies: I think we are getting there.... 😆 This behavior is also confirmed by configizer which only allows specifying of fields not managed by the Supervisor.

Thanks for that suggestion. I think something like that will work and am reformating this now. I don't want to blanket advise against modifying it (with a suitable warning) as for example, adding an SSH key for a prod image is incredibly useful. Those changes _do_ work consistently and I could imagine being popular requests.

garethtdavies: I pushed an update and reorganized so it flows a bit better - i.e. the how to safely edit (three different ways depending on what you want to do...) follows the advice to not do this directly on a provisioned device.

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,193 @@
+---
+title: Configuration
+excerpt: Instructions for editing and customizing config.json for modifying the behavior of {{ $names.os.lower }}
+---
+
+# Configuring {{ $names.os.lower }}
+
+The behavior of {{ $names.os.lower }} can be configured by editing the `config.json` file. This file is located in the [boot partition][boot-partition] and accepts a [range of fields](#valid-fields) to modify the behavior of the host OS.
+
+**Note:** The `config.json` file is different from the [`config.txt`][config-txt] file, also located in the boot partition, which is used by the Raspberry Pi to set device configuration options.
```
lurch: I wonder if it's worth mentioning "...which is (only) used by the Raspberry Pi to...", to highlight the fact that non-RaspberryPi devices don't use `config.txt`? :man_shrugging:

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,193 @@
...
+
+Before the device is [provisioned][provisioned], you may edit `config.json` by mounting a flashed SD card (which will mount as `resin-boot`) and editing the file directly. The boot partition is mounted on the device at `/mnt/boot`, and so on the device, the file is located at `/mnt/boot/config.json`. For example, to output a formatted version of `config.json` on a device, use the following commands:
```
lurch: I wonder if "(with the partition-label `resin-boot`)" is better than "(which will mount as `resin-boot`)"; since I guess the latter is OS-specific? (e.g. think of Windows users)  :man_shrugging:

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,193 @@
...
+
+```shell
+{{ $names.company.lower }} ssh <uuid>
+cat /mnt/boot/config.json | jq '.'
+```
+
+**Warning:** Editing config.json on a provisioned device should be done very carefully as any mistakes to this file can leave a device inaccessible.
```
lurch: nit: "mistakes to this file" doesn't sound right. Perhaps "any mistakes made in the syntax of this file" or "any errors created in this file" ?

pages/reference/OS/configuration.md
lurch: Is there a reliable list of which fields can be edited in `config.json` before provisioning (all of them?), and which fields _shouldn't_ be edited after provisioning, because the supervisor will overwrite any changes? Or is that wildly dependent on the particular version of the supervisor?

pdcastro: > Is there a reliable list of which fields can be edited in `config.json`

I may be misreading the supervisor source code, but it would appear that any of the fields listed in this [schema.ts](https://github.com/balena-io/balena-supervisor/blob/v10.6.26/src/config/schema.ts) file (like `uuid`, `apiKey`, `hostname`, `persistentLogging`...) are managed by the supervisor, and thus cannot be edited manually. Then this block of code seems to read values from a database and update `config.json`:
https://github.com/balena-io/balena-supervisor/blob/v10.6.26/src/config/index.ts#L153-L180

But I am not 100% sure, and perhaps @CameronDiver could comment further. Also, I don't know if it is a good idea to publish such a list, because this might prevent the supervisor from managing the fields in the future. Unless the list was explicitly linked to supervisor versions, alongside a warning that updating the host OS might break the user's app or scripts if they are "manually" editing some fields...

garethtdavies: I assumed the list here (valid fields) was the list of editable fields. 

My understanding then is after provisioning hostname (supervisor API) and persistent logging  (config variable) and the rest we recommend configizer (which only allows these other fields). I guess this is why we don't want users to manually edit this file as it can be unpredictable as noted in the text :)

pages/reference/OS/configuration.md
garethtdavies: Thanks for the feedback. I decided to leave this one out as I think it is mostly redundant given the context here is more to draw attention to the fact a similar filename exists.

pages/reference/OS/configuration.md
garethtdavies: I've left this with the link to https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html for more in-depth discussion.

pages/reference/OS/configuration.md
garethtdavies: Opened an issue to track this https://github.com/balena-io/docs/issues/1283

pages/reference/OS/configuration.md
garethtdavies: This will be addressed in a followup PR. I've extracted this to a partial and from the PR will be generated from the meta-balena repo.

garethtdavies:
Thanks all, addressed the above. There will be a quick follow-up PR to move the new valid fields partial to the meta-balena repo as this will be pulled in dynamically.

I've opened an issue for the potential issue for Network manager docs (dnsmasq behavior) https://github.com/balena-io/docs/issues/1283

garethtdavies:
@balena-ci retest

garethtdavies:
@balena-cli retest

pages/reference/OS/configuration.md
```diff
@@ -0,0 +1,76 @@
...
+{{> "general/config-json-fields"}}
+
+[boot-partition]: /reference/OS/overview/2.x/#stateless-and-read-only-rootfs
+[config-txt]: /reference/OS/advanced/#configtxt
+[country-codes]: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
+[hostname]: /reference/supervisor/supervisor-api/#patch-v1devicehost-config
+[configuration]: /learn/manage/configuration/
+[nm-connectivity]: https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html
+[configizer]:{{ $links.githubPlayground }}/configizer
+[masterclass-os]:{{ $links.githubProjects }}/balenaos-masterclass#12-advanced-editing-configjson-by-hand
```
hedss: I've moved these all over to the `balena-io` org now.

shared/general/config-json-fields.md
```diff
@@ -0,0 +1,118 @@
...
+- `0.resinio.pool.ntp.org`
+- `1.resinio.pool.ntp.org`
+- `2.resinio.pool.ntp.org`
+- `3.resinio.pool.ntp.org`
+
+### dnsServers
+
+(string) A space-separated list of preferred DNS servers to use for name resolution.
+
+- When `dnsServers` is not defined, or empty, Google's DNS server (8.8.8.8) is added to the list of DNS servers obtained via DHCP or statically configured in a NetworkManager connection profile.
```
hedss: It actually uses both the Google primary and secondary servers (`8.8.8.8` and `8.8.4.4`), see a `/etc/resolv.dnsmasq` config file on-device.

garethtdavies: Just checking this as this info came from @alexgg comment above https://github.com/balena-io/docs/pull/1249#discussion_r354797800. Specifically `And /run/dnsmasq.servers is the file that is configured from config.json with either the user provided servers or 8.8.8.8.` 

On my sample device, this appears to be true and `8.8.4.4` is not added.

```
root@3a628fc:~# cat /run/dnsmasq.servers

server=8.8.8.8
```

hedss: OK, that's interesting, maybe later OS changes have changed this (I looked on a 2.38).

garethtdavies: I think it's this line? https://github.com/balena-os/meta-balena/blob/a7c5f47ef43c9ef79b770ec6ac83989130f5f4ed/meta-balena-common/recipes-support/resin-vars/resin-vars/resin-vars#L90

garethtdavies:
@balena-ci retest

garethtdavies:
@balena-ci retest

PR #1258, opened 2019-12-06
docs: Update label names for balena

### LINKED ISSUE #1257: Change label names on "Define a container" doc.
This page still uses io.resin.* labels. They should be renamed with io.balena.* where new labels exist. A disclaimer can be put in place to indicate that for older versions of supervisor the old naming should be used.

Example disclaimer can be found here:
https://www.balena.io/docs/learn/develop/runtime/#dbus-communication-with-host-os

garethtdavies:
Closes #1257 

Change-type: patch

### DIFF
--- a/shared/general/labels-version-note.md
+++ b/shared/general/labels-version-note.md
@@ -1,0 +1,1 @@
+__Note:__ If you have devices in your app that have a supervisor version lower than 7.22.0, then you should use the `io.resin.features.` form of the labels to ensure that all devices obey the label. Earlier supervisor versions will not understand the `io.balena.features` label.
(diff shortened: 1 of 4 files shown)


shared/general/labels-version-note.md
```diff
@@ -0,0 +1 @@
+__Note:__ If you have devices in your app that have a supervisor version lower than 7.22.0, then you should use the `io.resin.features.` form of the labels to ensure that all devices obey the label. Earlier supervisor versions will not understand the `io.balena.features` label.
```
lurch: Is it worth adding a link to a page that explains how to determine the supervisor version? :man_shrugging:

xginn8: @lurch do you mean something more than the field that's exposed in the devices dashboard?
2019-12-08-122203_568x148_scrot

lurch: Sorry, I don't actually use balena, so I wasn't aware of what info was displayed in the dashboard! :fearful: 
I'll crawl back into my hole...

PR #1270, opened 2019-12-13
docs: Remove Artik links

### LINKED ISSUE #1265: Broken Link - Samsung ARTIK 530s 1G
The Getting Started Guide for the Samsung ARTIK 530s 1G contains a link to a page that is no longer active.  The first link here: https://www.balena.io/docs/learn/getting-started/artik533s/nodejs/#what-you-will-need ...is a link to https://www.artik.io/modules/artik-530/.   

However, clicking that link redirects to a different page, https://developer.artik.io/ which states that "This site is permanently offline."

We should probably just remove the link altogether I think.

garethtdavies:
Closes #1265 

Change-type: patch
Connects-to: 1265

### DIFF
--- a/config/dictionaries/device.json
+++ b/config/dictionaries/device.json
@@ -299,7 +299,6 @@
     "arch": "armv7hf",
     "bootMedia": "SD card",
     "icon": "/img/device/artik10.svg",
-    "link": "https://www.artik.io/modules/artik-1020/",
     "type": "released"
   },
   {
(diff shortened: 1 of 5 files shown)


lurch:
The artik530 change also removed the `[device]:{{ $device.link }}` from the bottom of the page - should the artik533 change do the same?

garethtdavies:
@balena-ci retest

Issue #1271, opened 2019-12-14
wording patch to dbus hostname change warning


dt-rush:
The wording in this doc in this part is not clear:

> Changing the device hostname is no longer possible via this method, due to the fact that the /etc/hostname file is stored on the read-only root partition. To change the device hostname, use the [balena supervisor API](https://www.balena.io/docs/reference/supervisor/supervisor-api/#patch-v1-device-host-config).

Ought to be made more clear:

> Changing the device hostname via a dbus-send method invocation of `org.freedesktop.hostname1.SetHostname` is no longer possible, due to the fact that this would attempt to write to `/etc/hostname`, which on the host OS is stored in the read-only root partition. To change the device hostname, use the [balena supervisor API](https://www.balena.io/docs/reference/supervisor/supervisor-api/#patch-v1-device-host-config).

garethtdavies:
This was merged so closing.

PR #1272, opened 2019-12-14
wording patch to dbus hostname change warning

dt-rush:
Change-type: patch
Connects-to: #1271
Signed-off-by: dt-rush <nickp@balena.io>

### DIFF
--- a/pages/learn/develop/runtime.md
+++ b/pages/learn/develop/runtime.md
@@ -54,7 +54,7 @@ root@raspberrypi3-cc723d7:/# printenv | grep {{ $names.company.allCaps }}
 
 ### Dbus communication with host OS
 
-In some cases its necessary to communicate with the host OS systemd to perform actions on the host, for example changing the hostname. To do this you can use [dbus][dbus-link]. In order to ensure that you are communicating to the host OS systemd and not the systemd in your container it is important to set `DBUS_SYSTEM_BUS_ADDRESS` for all dbus communication. The setting of that environment variable is different for older and newer devices (based on the {{ $names.company.lower }} supervisor version), choose the line that is correct for your device's OS version (can be found in your device dashboard):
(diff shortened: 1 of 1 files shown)


garethtdavies:
In addition, it seems this section is outdated and could do with a review https://www.balena.io/docs/learn/develop/runtime/#dbus-communication-with-host-os, in particular:

```In some cases its necessary to communicate with the host OS systemd to perform actions on the host, for example changing the hostname. To do this you can use dbus. ```

This seems redundant given it's immediately followed by:

```Changing the device hostname via a dbus-send method invocation ... is no longer possible```

If there is a specific version that made this incompatible we should mention it (if we want to keep the section about it being possible). I'm guessing Supervisor 1.7.0 is _really_ old and is that still necessary?

**Comment by garethtdavies** [comment]
@dt-rush I couldn't make a suggested change so for simplicity just pushed another commit removing the example of dbus hostname. That work for you?

File: pages/learn/develop/runtime.md [hunk 1]
```diff
@@ -54,7 +54,7 @@ root@raspberrypi3-cc723d7:/# printenv | grep {{ $names.company.allCaps }}
...
 
-In some cases its necessary to communicate with the host OS systemd to perform actions on the host, for example changing the hostname. To do this you can use [dbus][dbus-link]. In order to ensure that you are communicating to the host OS systemd and not the systemd in your container it is important to set `DBUS_SYSTEM_BUS_ADDRESS` for all dbus communication. The setting of that environment variable is different for older and newer devices (based on the {{ $names.company.lower }} supervisor version), choose the line that is correct for your device's OS version (can be found in your device dashboard):
+In some cases its necessary to communicate with the host OS systemd to perform actions on the host. To do this you can use [dbus][dbus-link]. In order to ensure that you are communicating to the host OS systemd and not the systemd in your container it is important to set `DBUS_SYSTEM_BUS_ADDRESS` for all dbus communication. The setting of that environment variable is different for older and newer devices (based on the {{ $names.company.lower }} supervisor version), choose the line that is correct for your device's OS version (can be found in your device dashboard):
```
lurch: "some cases its necessary" -> "some cases it's necessary"

garethtdavies:
@balena-ci retest

PR #1274, opened 2019-12-16
deps: Updating dependabot dependencies

garethtdavies:
Cherry-picked these individual commits into a single PR to save rebasing after each one.

I'm unsure of how we would properly test this. I've pushed this to staging and tested locally and see no ill-effects https://www.balena-staging.io/docs/learn/getting-started/raspberrypi3/nodejs/


garethtdavies:
@balena-ci retest

Issue #1254, opened 2019-12-05
Update base images to include Raspbian


garethtdavies:
Raspian is missing from https://www.balena.io/docs/reference/base-images/base-images/  but is available here https://hub.docker.com/r/balenalib/rpi-raspbian/tags.

The list of architectures, distributions and language stacks should be automated to reflect what is available on DockerHub.

lurch:
NIT1: There's a 'b' in "Raspbian" :wink: 

NIT2: For the link you posted, DockerHub says "OS/ARCH: linux/amd64" for each image - is that just a DockerHub bug / limitation?

garethtdavies:
I don't know whether this is a Docker thing (will investigate during resolving this issue) but that behavior is consistent with other images e.g. https://hub.docker.com/r/balenalib/aarch64-ubuntu-node/tags

lurch:
I did a bit of clicking around and e.g. https://hub.docker.com/r/arm64v8/postgres/tags says `linux/arm64`

klutchell:
The `balenalib/rpi-rasbian` images are not proper balena base images and don't contain our helper functions like `install_packages`, `cross-build`, `balena-idle`, etc.

As far as I can tell they only exist to fill a gap because `armv6` is not available from official `debian` base images. You can see the need by comparing these two Dockerfiles.

https://github.com/balena-io-library/base-images/blob/8e39e5184f4ebca0bd95411b159e98557f8e3f9b/balena-base-images/rpi/debian/buster/run/Dockerfile#L1-L2

https://github.com/balena-io-library/base-images/blob/8e39e5184f4ebca0bd95411b159e98557f8e3f9b/balena-base-images/armv7hf/debian/buster/run/Dockerfile#L1-L2

PR #1285, opened 2019-12-20
baseImages: Updating base image versions

garethtdavies:
@nghiant2710 I've manually updated the base image page for updated version information and to get a better sense of how it works before trying to pull this information in dynamically from the contracts.

I wrote a simple script to parse the `version` field for each variant from here for Node.js for example https://github.com/balena-io/contracts/blob/master/contracts/sw.stack/node/contract.json. A couple of questions as result:

* There are some images that do not have contracts but do exist in DockerHub, specifically Node 10.10 which is referenced in these docs https://hub.docker.com/layers/balenalib/raspberrypi3-debian-node/10.10.0-stretch-run/images/sha256-15c6d3683496a4813d487e79796f30debf1a637a42ebca258c3b6224ea15784f. As I understand these point to the corresponding frozen date image. So only those with contracts get updates?
* As per #1254 there is a Raspbian image but again doesn't have a contract as I would expect under 
`sw.os` https://github.com/balena-io/contracts/tree/master/contracts/sw.os. Is there support for Raspbian?

Change-type: patch

### DIFF
--- a/pages/reference/base-images/base-images.md
+++ b/pages/reference/base-images/base-images.md
@@ -10,91 +10,90 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
 ## Features Overview
 
 - Multiple Architectures:
-	- armv5e
-	- armv6
-	- armv7hf
-	- aarch64
-	- amd64
-	- i386
+  - armv5e
+  - armv6
+  - armv7hf
+  - aarch64
+  - amd64
+  - i386
 - Multiple Distributions:
-	- [Debian](https://www.debian.org/): jessie, sid, stretch and buster
-	- [Alpine](https://alpinelinux.org/): 3.5, 3.6, 3.7, 3.8, 3.9, 3.10 and edge
-	- [Ubuntu](https://www.ubuntu.com/): trusty (14.04), xenial (16.04), artful (17.10), bionic (18.04), and cosmic (18.10)
-	- [Fedora](https://getfedora.org/): 26, 28, 29 and 30
+  - [Debian](https://www.debian.org/): jessie (8), stretch (9), buster (10) and sid
+  - [Alpine](https://alpinelinux.org/): 3.6, 3.7, 3.8, 3.9, 3.10, 3.11 and edge
+  - [Ubuntu](https://www.ubuntu.com/): xenial (16.04), bionic (18.04), cosmic (18.10), disco (19.04) and eoan (19.10)
+  - [Fedora](https://getfedora.org/): 26, 28, 29, 30 and 31
 - Multiple language stacks:
-	- [Node.js](https://nodejs.org/en/): 12.7.0, 11.15.0, 10.16.1, 8.16.1 and 6.17.1
-	- [Python](https://www.python.org/): 2.7.16, 3.5.7, 3.6.9 and 3.7.4
-	- [openJDK](https://openjdk.java.net/): 7-jdk/jre, 8-jdk/jre, 10-jdk/jre and 11-jdk/jre
-	- [Golang](https://golang.org/): 1.12.7, 1.11.12 and 1.10.8
-	- [Dotnet](https://docs.microsoft.com/en-gb/dotnet/core/): 2.2-sdk/runtime/aspnet,3.0-sdk/runtime/aspnet,2.1-sdk/runtime/aspnet
+  - [Node.js](https://nodejs.org/en/): 8.11.1, 8.17.0, 9.11.2, 10.18.0, 11.15.0, 12.14.0 and 13.5.0
+  - [Python](https://www.python.org/): 2.7.17, 3.4.10, 3.5.7, 3.6.9, 3.7.5 and 3.8.0
+  - [openJDK](https://openjdk.java.net/): 7-jdk/jre, 8-jdk/jre and 11-jdk/jre
+  - [Golang](https://golang.org/): 1.10.8, 1.11.13, 1.12.13 and 1.13.4
+  - [Dotnet](https://docs.microsoft.com/en-gb/dotnet/core/): 2.1-sdk/runtime/aspnet, 2.2-sdk/runtime/aspnet and 3.1-sdk/runtime/aspnet
 - [`run`](#run-vs-build) and [`build`](#run-vs-build) variants designed for multistage builds.
 - [cross-build](#building-arm-containers-on-x86-machines) functionality for building ARM containers on x86.
 - Helpful package installer script called `install_packages` inspired by [minideb](https://github.com/bitnami/minideb#why-use-minideb).
 
-
 ## How to Pick a Base Image
 
-<!-- TODO: Add image on how image hierarchy and discuss  -->
-When starting out a project its generally easier to have a "fatter" image which contains a lot of prebuilt dependencies and tools. These images help you get setup faster and work out the requirements for your project. For this reason its recommended to start with `-build` variants and as your project progresses switch to a `-run` variant with some [docker multistage build magic][multistage-build-docs] to slim your deploy image down. In most cases your project can just use a Debian based distribution, which is the default if not specified, but if you know the requirements of your project or prefer specific distros, Ubuntu, Alpine and Fedora images are available. The set of `balenalib` base images follow a simple naming scheme described below, which will help you select a base image for your specific needs.
+<!-- TODO: Add image on how image hierarchy and discuss -->
+When starting out a project, it's generally easier to have a "fatter" image, which contains a lot of prebuilt dependencies and tools. These images help you get setup faster and work out the requirements for your project. For this reason, it's recommended to start with `-build` variants, and as your project progresses, switch to a `-run` variant with some [docker multistage build magic][multistage-build-docs] to slim your deploy image down. In most cases, your project can just use a Debian based distribution, which is the default if not specified, but if you know the requirements of your project or prefer specific distros, Ubuntu, Alpine, and Fedora images are available. The set of `balenalib` base images follow a simple naming scheme described below, which will help you select a base image for your specific needs.
 
 ### How the Image Naming Scheme Works
 
-With over 26000 `balenalib` base images to choose from it can be overwhelming to decide which image and tag is correct for your project. To pick the correct image, it helps to understand how the images are named as that indicates what is installed in the image. In general the naming scheme for the `balenalib` image set follows the pattern below:
+With over 26000 `balenalib` base images to choose from, it can be overwhelming to decide which image and tag are correct for your project. To pick the correct image, it helps to understand how the images are named as that indicates what is installed in the image. In general, the naming scheme for the `balenalib` image set follows the pattern below:
+
 ```
 balenalib/<hw>-<distro>-<lang_stack>:<lang_ver>-<distro_ver>-(build|run)-<yyyymmdd>
 ```
 
-#### Image Names:
+#### Image Names
 
-- `<hw>` is either architecture or device type and is **mandatory**. If using Dockerfile.templates you can replace this with `%%BALENA_MACHINE_NAME%%` or `%%BALENA_ARCH%%`. For a list of available device names and architectures see the [Device types](/reference/base-images/devicetypes/).
-- `<distro>`  is the linux distribution, currently there are 4 distributions, namely Debian, Alpine, Ubuntu and Fedora. This field is optional, and will default to Debian if left out.
-- `<lang_stack>` is the programming language pack, currently we support Node.js, Python, OpenJDK and Go. This field is optional and if left out, no language pack will be installed, so you will just have the distribution.
+- `<hw>` is either architecture or device type and is **mandatory**. If using Dockerfile.templates, you can replace this with `%%BALENA_MACHINE_NAME%%` or `%%BALENA_ARCH%%`. For a list of available device names and architectures, see the [Device types](/reference/base-images/devicetypes/).
+- `<distro>` is the Linux distribution. Currently there are 4 distributions, namely Debian, Alpine, Ubuntu and Fedora. This field is optional and will default to Debian if left out.
+- `<lang_stack>` is the programming language pack, currently we support Node.js, Python, OpenJDK and Go. This field is optional, and if left out, no language pack will be installed, so you will just have the distribution.
 
-#### Image Tags:
+#### Image Tags
 
-In the tags, all of the fields are optional and if they are left out, they will default to their `latest` pointer.
+In the tags, all of the fields are optional, and if they are left out, they will default to their `latest` pointer.
 
-- `<lang_ver>` is the version of the language stack, for example node.js 10.10, it can also be substituted for `latest`.
-- `<distro_ver>` is the version of the linux distro, for example in the case of Debian there are 4 valid versions, namely `sid`, `jessie`, `buster` and `stretch`.
-- For each combination of distro and stack we have two variants called `run` and `build`. The build variant is much heavier as it has a number of tools preinstalled to help with building source code. You can see an example of the tools that are included in the Debian Stretch variant [here](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/build/Dockerfile). The `run` variants are stripped down and only include a few useful runtime tools, see an example [here](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile).
-If no variant is specified, the image defaults to `run`
-- <yyyymmdd> the last optional field on tags is the date tag. This is useful for production deployments as these base images are non-moving tags, so no packages in these will update ever.
+- `<lang_ver>` is the version of the language stack, for example, Node.js 10.10, it can also be substituted for `latest`.
+- `<distro_ver>` is the version of the Linux distro, for example in the case of Debian, there are 4 valid versions, namely `sid`, `jessie`, `buster` and `stretch`.
+- For each combination of distro and stack, we have two variants called `run` and `build`. The build variant is much heavier as it has a number of tools preinstalled to help with building source code. You can see an example of the tools that are included in the Debian Stretch variant [here]({{ $links.githubLibrary }}/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/build/Dockerfile). The `run` variants are stripped down and only include a few useful runtime tools, see an example [here]({{ $links.githubLibrary }}/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile). If no variant is specified, the image defaults to `run`
+- The last optional field on tags is the date tag `<yyyymmdd>`. This is useful for production deployments as these base images are non-moving tags, so no packages in these will update ever.
 
 #### Examples
 
-`balenalib/raspberrypi3-node:10.10`
+`balenalib/raspberrypi3-node:10.18`
 
-- `<hw>` : raspberrypi3 - The raspberry pi 3 device type.
+- `<hw>` : raspberrypi3 - The Raspberry Pi 3 device type.
 - `<distro>` : omitted, so it defaults to Debian.
 - `<lang>` : node - the Node.js runtime and npm will be installed
-- `<lang_ver>` : 10.10 - This gives us node.js version 10.10.x whatever is the latest patch version provided on balenalib
-- `<distro_ver>` : omitted, so it defaults to `stretch`
+- `<lang_ver>` : 10.18 - This gives us Node.js version 10.18.x whatever is the latest patch version provided on balenalib
+- `<distro_ver>` : omitted, so it defaults to `buster`
 - `(build|run)` : omitted, so the image defaults to the slimmed down `run` variant
-- `<yyyymmdd>` : omitted, we don’t have a date frozen image, so new updates pushed to our 10.10 tag, for example patch versions from node.js will automatically be inherited when they are available.
+- `<yyyymmdd>` : omitted, we don't have a date frozen image, so new updates pushed to our 10.18 tag, for example patch versions from Node.js will automatically be inherited when they are available.
 
-`balenalib/i386-ubuntu-python:latest-trusty-build-20181029`
+`balenalib/i386-ubuntu-python:latest-bionic-build-20191029`
 
 - `<hw>` : i386 - the intel 32 bit architecture that runs on Intel Edison
 - `<distro>` : ubuntu
 - `<lang>` : python
-- `<lang_ver>` : `latest` points to the latest python 2 version, which currently is 2.7.15
-- `<distro_ver>` : trusty is Ubuntu 14.04
+- `<lang_ver>` : `latest` points to the latest Python 2 version, which currently is 2.7.17
+- `<distro_ver>` : bionic is Ubuntu 18.04
 - `(build|run)` : `build` - to include things like `build-essential` and `gcc`
-- `<yyyymmdd>` : 20181029 is a date frozen image - so this image will never be updated on dockerhub. Pinning to a date frozen base image is a good idea if you are running a fleet in production and are sensitive to dependencies updating and/or bandwidth constrained.
+- `<yyyymmdd>` : 20191029 is a date frozen image - so this image will never be updated on Docker Hub. Pinning to a date frozen base image is a good idea if you are running a fleet in production and are sensitive to dependencies updating and/or bandwidth constrained.
 
 ### run vs. build
 
 For each combination of `<hw>`-`<distro>`-`<lang>` there is both a `run` and a `build` variant. These variants are provided to allow for easier multistage builds.
 
-The `run` variant is designed to be a slim and minimal variant with only runtime essentials packaged into it. An example of the packages installed in can be seen in the [`Dockerfile`](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile#L7) of [`balenalib/armv7hf-debian:run`](https://github.com/balena-io-library/base-images/tree/master/balena-base-images/armv7hf/debian/stretch/run).
+The `run` variant is designed to be a slim and minimal variant with only runtime essentials packaged into it. An example of the packages installed in can be seen in the [`Dockerfile`]({{ $links.githubLibrary }}/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile#L7) of [`balenalib/armv7hf-debian:run`]({{ $links.githubLibrary }}/base-images/tree/master/balena-base-images/armv7hf/debian/stretch/run).
 
-The `build` variant is a heavier image that includes many of the tools required for building from source such as `build-essential`, `gcc`, etc. As an example you can see the types of packages installed in the `balenalib/armv7hf-debian:build` variant [here](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/build/Dockerfile#L51).
+The `build` variant is a heavier image that includes many of the tools required for building from source such as `build-essential`, `gcc`, etc. As an example, you can see the types of packages installed in the `balenalib/armv7hf-debian:build` variant [here]({{ $links.githubLibrary }}/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/build/Dockerfile#L51).
 
-These variants make building multistage projects easier, take for example installing an I2C node.js package which requires a number of build time dependencies to build the native `i2c` node module, but we don't want to send all of those down to our device. This is the perfect time for multistage builds and to use the `build` and `run` variants.
+These variants make building multistage projects easier, take for example, installing an I2C node.js package, which requires a number of build time dependencies to build the native `i2c` node module, but we don't want to send all of those down to our device. This is the perfect time for multistage builds and to use the `build` and `run` variants.
 
 ```Dockerfile
 FROM balenalib/raspberrypi3-debian-node:10.10-stretch-build as build
-RUN npm install --only=production  i2c
+RUN npm install --only=production i2c
 
 # The run time container that will go to devices
 FROM balenalib/raspberrypi3-debian-node:10.10-stretch-run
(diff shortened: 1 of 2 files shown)


nghiant2710:
Great work @garethtdavies 👍 

Regarding your questions:
1. We only keep the latest version in contracts so you won't find information about old version there.
2. We do support raspbian and all devices with `arch.sw = rpi` will be built off `balenalib/rpi-raspbian`. As the raspbian images are not available on dockerhub, we create the `balenalib/rpi-raspbian` from this repo https://github.com/balena-io-library/resin-rpi-raspbian and all base images of `raspberry-pi`  device type (RPi 1 and Zero) will be Raspbian. For some reasons discussed long time ago, we want to have `raspberry-pi-debian` as an alias of `rpi-raspbian` so you will see `rpi` in `debian` contract here https://github.com/balena-io/contracts/blob/master/contracts/sw.os/debian/contract.json

pages/reference/base-images/base-images.md
```diff
@@ -18,15 +18,15 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
+	- [Ubuntu](https://www.ubuntu.com/): xenial (16.04), bionic (18.04), cosmic (18.10), disco (19.04) and eoan (19.10)
+	- [Fedora](https://getfedora.org/): 26, 28, 29, 30 and 31
 - Multiple language stacks:
-	- [Node.js](https://nodejs.org/en/): 12.7.0, 11.15.0, 10.16.1, 8.16.1 and 6.17.1
-	- [Python](https://www.python.org/): 2.7.16, 3.5.7, 3.6.9 and 3.7.4
-	- [openJDK](https://openjdk.java.net/): 7-jdk/jre, 8-jdk/jre, 10-jdk/jre and 11-jdk/jre
-	- [Golang](https://golang.org/): 1.12.7, 1.11.12 and 1.10.8
-	- [Dotnet](https://docs.microsoft.com/en-gb/dotnet/core/): 2.2-sdk/runtime/aspnet,3.0-sdk/runtime/aspnet,2.1-sdk/runtime/aspnet
+	- [Node.js](https://nodejs.org/en/): 13.1.0, 12.13.0, 11.15.0, 10.17.0, 9.11.2, 8.16.1 and 8.11.1
+	- [Python](https://www.python.org/): 2.7.17, 3.8.0, 3.7.5, 3.6.9, 3.5.7 and 3.4.10
```
lurch: 2.7.x, 3.8.x, 3.7.x, 3.6.x, 3.5.x and 3.4.x looks like a strange version-ordering system? :wink: 

Also, it seems some version-lists here are sorted largest -> smallest (e.g. Node.js), and others are sorted smallest -> largest (e.g. openJDK) ?

pages/reference/base-images/base-images.md
```diff
@@ -112,18 +112,18 @@ Currently balenalib supports the following OS distributions and Language stacks,
...
 
 | Language | Default  	                  | Supported Architectures                      |
 |---------|------------------------------|----------------------------------------------|
-| Node.js | v12.7.0                      | armv6, armv7hf, aarch64, amd64, i386         |
-| Python  | v2.7.16                       | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
-| OpenJDK | v1.8.0_181                    | armv7hf, aarch64, amd64, i386, armv6         |
-| Go      | 1.12.7                        | armv7hf, aarch64, amd64, i386, armv6         |
-| Dotnet  | 2.2-sdk                       | armv7hf, aarch64, amd64                |
+| Node.js | v13.1.0                      | armv6, armv7hf, aarch64, amd64, i386         |
+| Python  | v2.7.17                       | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
```
lurch: I wonder if it's also worth having a separate Python3 language tag, bearing in mind the upcoming EOL of Python2 ? https://pythonclock.org/

garethtdavies:
Thanks for this explanation @nghiant2710 . I've tested this quite a bit now, but I am still a bit confused about Raspbian and how to document it.

I understand those device types with device type `raspberry-pi` will use Raspbian by default, but later variants, notably, the Pi3 does not. If I use the machine name `raspberrypi3`  or %%BALENA_MACHINE_NAME%% I am always going to get Debian if I don't specify an OS. I can specify Ubuntu, Fedora, and Alpine but not Raspbian, so to get Raspbian so on the Pi3 I can get it e.g.:

```
FROM balenalib/raspberry-pi (i.e. a device type of raspberry pi)
FROM balenalib/rpi-raspbian:jessie
FROM balenalib/raspberry-pi-debian
```

I guess I'm expecting these to work, but they, of course, do not:

```
FROM balenalib/raspberrypi3-raspbian
FROM balenalib/%%BALENA_MACHINE_NAME%%-raspbian
```

So, I'm not sure how we should document this for Raspbian. Basically, if you want to use Raspbian, use the base image `balenalib/rpi-raspbian` with available tags of `jessie, `stretch` and `buster`?

pages/reference/base-images/base-images.md
```diff
@@ -112,18 +112,18 @@ Currently balenalib supports the following OS distributions and Language stacks,
 
 | Distribution | Default                | Supported Architectures                      |
 |---------|------------------------------|----------------------------------------------|
-| Debian  | Debian GNU/Linux 9 (stretch) | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
+| Debian  | Debian GNU/Linux 10 (Buster) | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
 | Alpine  | Alpine Linux v3.10            | armv6, armv7hf, aarch64, amd64, i386 		|
 | Ubuntu  | 18.04 LTS (Bionic Beaver)    | armv7hf, aarch64, amd64, i386                |
-| Fedora  | Fedora 29 (Twenty Nine)      | armv7hf, aarch64, amd64, i386                |
+| Fedora  | Fedora 30 (Thirty)      | armv7hf, aarch64, amd64, i386                |
```
xginn8:
```suggestion
| Fedora  | Fedora 30               | armv7hf, aarch64, amd64, i386                |
```

pages/reference/base-images/base-images.md
```diff
@@ -62,25 +62,25 @@ If no variant is specified, the image defaults to `run`
...
 
-`balenalib/i386-ubuntu-python:latest-trusty-build-20181029`
+`balenalib/i386-ubuntu-python:latest-bionic-build-20191029`
 
 - `<hw>` : i386 - the intel 32 bit architecture that runs on Intel Edison
 - `<distro>` : ubuntu
 - `<lang>` : python
-- `<lang_ver>` : `latest` points to the latest python 2 version, which currently is 2.7.15
-- `<distro_ver>` : trusty is Ubuntu 14.04
+- `<lang_ver>` : `latest` points to the latest python 2 version, which currently is 2.7.17
```
xginn8:
```suggestion
- `<lang_ver>` : `latest` points to the latest Python 2 version, which currently is 2.7.17
```

pages/reference/base-images/base-images.md
```diff
@@ -62,25 +62,25 @@ If no variant is specified, the image defaults to `run`
...
+- `<distro_ver>` : bionic is Ubuntu 18.04
 - `(build|run)` : `build` - to include things like `build-essential` and `gcc`
-- `<yyyymmdd>` : 20181029 is a date frozen image - so this image will never be updated on dockerhub. Pinning to a date frozen base image is a good idea if you are running a fleet in production and are sensitive to dependencies updating and/or bandwidth constrained.
+- `<yyyymmdd>` : 20191029 is a date frozen image - so this image will never be updated on dockerhub. Pinning to a date frozen base image is a good idea if you are running a fleet in production and are sensitive to dependencies updating and/or bandwidth constrained.
```
xginn8:
```suggestion
- `<yyyymmdd>` : 20191029 is a date frozen image - so this image will never be updated on Docker Hub. Pinning to a date frozen base image is a good idea if you are running a fleet in production and are sensitive to dependencies updating and/or bandwidth constrained.
```

lurch:
In case it helps to clear up any confusion (at least from the Raspbian side, if not from the Balena side :wink: ) - the Raspbian images available from https://www.raspberrypi.org/downloads/raspbian/ support all models of the Raspberry Pi (Pi1, Pi Zero, Pi2, Pi3, Pi4) with a single image (which is presumably why there's no `raspberrypi3-raspbian`), and it's a 32-bit kernel and 32-bit userland.

I dunno how closely Balena's `raspbian` images follow the Raspberry Pi `raspbian` images though.

nghiant2710:
@garethtdavies, you're right, only `raspberry-pi` device type is `raspbian`, as you can run `armv7` debian images on Pi2 so we use it as the base images for Pi2 and Pi3. However, we do add raspbian package source and pre-install userland in the Pi2 and Pi3 base images (basically armv7hf debian will raspbian userland installed).

Another thing is `debian` is the default OS for the base images, if you don't specify OS in the base images then you will get `debian` (or `raspbian` in case of `raspberry-pi`).

garethtdavies:
Thanks all for the discussion, that's much clearer to me now 😄  At the root of this is simply what we should include in the docs with reference to Raspbian.  To that end, I've added a note to the docs about Raspbian. @nghiant2710 can you review this note https://github.com/balena-io/docs/pull/1285/files#diff-aa7d2966be46f689a147702028e9ac59R129 or have any other suggestions about any other missing information or where this should be included?

I've also updated this file for the latest contracts, and I have done a general cleanup, but I'll follow up with additional PRs for auto-generating these docs.

pages/reference/base-images/base-images.md
```diff
@@ -10,67 +10,66 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
 
 `balenalib/raspberrypi3-node:10.17`
 
-- `<hw>` : raspberrypi3 - The raspberry pi 3 device type.
+- `<hw>` : raspberrypi3 - The Raspberry Pi 3 device type.
 - `<distro>` : omitted, so it defaults to Debian.
 - `<lang>` : node - the Node.js runtime and npm will be installed
-- `<lang_ver>` : 10.17 - This gives us node.js version 10.17.x whatever is the latest patch version provided on balenalib
-- `<distro_ver>` : omitted, so it defaults to `stretch`
+- `<lang_ver>` : 10.17 - This gives us Node.js version 10.18.x whatever is the latest patch version provided on balenalib
```
lurch: Huh? Mix of 10.17 and 10.18

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
...
+```Dockerfile
 FROM balenalib/raspberrypi3
 
 RUN install_packages wget git
 
 CMD ["bash", "start.sh"]
 ```
 
-This will run an `apt-get update -qq`, then install `wget` and `git` via apt-get with `-y --no-install-recommends` flags and it will by default try this 2 times before failing. You can see the source of `install_packages` [here](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile#L25-L49).
+This will run an `apt-get update -qq`, then install `wget` and `git` via apt-get with `-y --no-install-recommends` flags, and it will by default try this 2 times before failing. You can see the source of `install_packages` [here]({{ $links.githubLibrary }}/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile#L25-L49).
```
lurch: Should this be `#L26-L49` ?

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
...
+In many IoT projects, your containers will want to interact with some hardware, often this hardware is plugged in at runtime, in the case of USB or serial devices. In these cases, you will want to enable [`udevd`][udevd-link] in your container. In `balenalib` images this can easily be done either by adding `ENV UDEV=1` in your `Dockerfile` or by setting an [environment variable](https://www.balena.io/docs/learn/manage/serv-vars/).
 
-You will also need to run your container `privileged`, by default any [balenaCloud](https://www.balena.io/cloud/) projects that don't contain a `docker-compose.yml` will run their containers `privileged`. If you are using a multicontainer project you will need to add `privileged: true` to each of the service definitions for the services that need hardware access.
+You will also need to run your container `privileged`, by default. Any [balenaCloud](https://www.balena.io/cloud/) projects that don't contain a `docker-compose.yml` will run their containers `privileged`. If you are using a multicontainer project, you will need to add `privileged: true` to each of the service definitions for the services that need hardware access.
```
lurch: Was the full-stop put in the wrong place here? Looks like maybe it should say "You will also need to run your container `privileged`. By default, any ..." ?

pages/reference/base-images/base-images.md
```diff
@@ -10,67 +10,66 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
+  - armv7hf
+  - aarch64
+  - amd64
+  - i386
 - Multiple Distributions:
-	- [Debian](https://www.debian.org/): jessie, sid, stretch and buster
-	- [Alpine](https://alpinelinux.org/): 3.6, 3.7, 3.8, 3.9, 3.10 and edge
-	- [Ubuntu](https://www.ubuntu.com/): xenial (16.04), bionic (18.04), cosmic (18.10), disco (19.04) and eoan (19.10)
-	- [Fedora](https://getfedora.org/): 26, 28, 29, 30 and 31
+  - [Debian](https://www.debian.org/): jessie, stretch, buster and sid
```
lurch: I wonder if it's worth putting the Debian version numbers in brackets, similar to Ubuntu below?

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
...
 
-#### Notices
+Not all OS distro and language stack versions are compatible with each other. Notice that there are some combinations that are not available in the `balenalib` base images.
 
-Not all OS distro and language stack versions are compatible with each other, pleace notice that there are some combinations which are not available in the `balenalib` base images.
 - [Nodejs dropped 32-bit builds](https://github.com/nodejs/build/issues/885) a while ago so i386-based nodejs images (Debian, Fedora and Ubuntu) v8.x and v6.x are official. New series (v10.x and v12.x) are using unofficial builds so please notice when using these images.
 - armv6 binaries were officially dropped from v12 and v12 armv6 support is now considered unofficial.
 - The Nodejs v6.x and v8.x series are not available for i386 Alpine Linux base images v3.9 and edge as node crashes with segfault error, we are investigating the issue and will add them back as soon as the issue is resolved.
```
lurch: Nitpick: Nodejs v6.x isn't included in the list of supported Node.js versions earlier on the page :wink:

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
...
-| Dotnet  | 3.1-sdk                       | armv7hf, aarch64, amd64                |
+Devices with a [device type](/reference/base-images/devicetypes/) of `raspberry-pi` (Raspberry Pi1 and Zero) will be built from `balenalib/rpi-raspbian` and will be [Raspbian]({{ $links.githubLibrary }}/resin-rpi-raspbian) base images. Raspberry Pi2 and Pi3 base images have the Raspbian package source added and Raspbian userland pre-installed.
 
-#### Notices
+Not all OS distro and language stack versions are compatible with each other. Notice that there are some combinations that are not available in the `balenalib` base images.
 
-Not all OS distro and language stack versions are compatible with each other, pleace notice that there are some combinations which are not available in the `balenalib` base images.
 - [Nodejs dropped 32-bit builds](https://github.com/nodejs/build/issues/885) a while ago so i386-based nodejs images (Debian, Fedora and Ubuntu) v8.x and v6.x are official. New series (v10.x and v12.x) are using unofficial builds so please notice when using these images.
 - armv6 binaries were officially dropped from v12 and v12 armv6 support is now considered unofficial.
```
lurch: Is the `v12` on this line referring to Nodejs?

pages/reference/base-images/base-images.md
garethtdavies: Thanks, I left this as while it isn't supported there are v6 base images available on Docker Hub.

pages/reference/base-images/base-images.md
garethtdavies: Nice catch, which is why I want to autogenerate this as much as possible 😄

pages/reference/base-images/base-images.md
```diff
@@ -66,7 +66,7 @@ In the tags, all of the fields are optional, and if they are left out, they will
 - `<hw>` : raspberrypi3 - The Raspberry Pi 3 device type.
 - `<distro>` : omitted, so it defaults to Debian.
 - `<lang>` : node - the Node.js runtime and npm will be installed
-- `<lang_ver>` : 10.17 - This gives us Node.js version 10.18.x whatever is the latest patch version provided on balenalib
+- `<lang_ver>` : 10.18 - This gives us Node.js version 10.18.x whatever is the latest patch version provided on balenalib
```
lurch: The 10.18 here is now fixed, but the example image-name a few lines up is still `balenalib/raspberrypi3-node:10.17` ? :eyes:

pages/reference/base-images/base-images.md
```diff
@@ -130,9 +130,9 @@ Devices with a [device type](/reference/base-images/devicetypes/) of `raspberry-
 
 Not all OS distro and language stack versions are compatible with each other. Notice that there are some combinations that are not available in the `balenalib` base images.
 
-- [Nodejs dropped 32-bit builds](https://github.com/nodejs/build/issues/885) a while ago so i386-based nodejs images (Debian, Fedora and Ubuntu) v8.x and v6.x are official. New series (v10.x and v12.x) are using unofficial builds so please notice when using these images.
-- armv6 binaries were officially dropped from v12 and v12 armv6 support is now considered unofficial.
-- The Nodejs v6.x and v8.x series are not available for i386 Alpine Linux base images v3.9 and edge as node crashes with segfault error, we are investigating the issue and will add them back as soon as the issue is resolved.
+- [Node.js dropped 32-bit builds](https://github.com/nodejs/build/issues/885) a while ago so i386-based nodejs images (Debian, Fedora and Ubuntu) v8.x and v6.x are official. New series (v10.x and v12.x) are using unofficial builds so please notice when using these images.
```
lurch: What does "so please notice when using these images" mean? :wink:

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
...
+#### Notes
 
-| Language | Default  	                  | Supported Architectures                      |
-|---------|------------------------------|----------------------------------------------|
-| Node.js | v12.7.0                      | armv6, armv7hf, aarch64, amd64, i386         |
-| Python  | v2.7.16                       | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
-| OpenJDK | v1.8.0_181                    | armv7hf, aarch64, amd64, i386, armv6         |
-| Go      | 1.12.7                        | armv7hf, aarch64, amd64, i386, armv6         |
-| Dotnet  | 2.2-sdk                       | armv7hf, aarch64, amd64                |
+Devices with a [device type](/reference/base-images/devicetypes/) of `raspberry-pi` (Raspberry Pi1 and Zero) will be built from `balenalib/rpi-raspbian` and will be [Raspbian]({{ $links.githubLibrary }}/resin-rpi-raspbian) base images. Raspberry Pi2 and Pi3 base images have the Raspbian package source added and Raspbian userland pre-installed.
```
lurch: "Raspberry Pi2 and Pi3 base images have the Raspbian package source added and Raspbian userland pre-installed."

Is this true for just Debian on Raspberry Pi, or for the other OSes too? (I guess it wouldn't make sense to add the Raspbian package source to Fedora?)

Also, I think "Raspberry Pi2 and Pi3" ought to be "Raspberry Pi2 and Raspberry Pi3" ? :man_shrugging:

pages/reference/base-images/base-images.md
```diff
@@ -10,91 +10,90 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
 
 ### How the Image Naming Scheme Works
 
-With over 26000 `balenalib` base images to choose from it can be overwhelming to decide which image and tag is correct for your project. To pick the correct image, it helps to understand how the images are named as that indicates what is installed in the image. In general the naming scheme for the `balenalib` image set follows the pattern below:
+With over 26000 `balenalib` base images to choose from, it can be overwhelming to decide which image and tag are correct for your project. To pick the correct image, it helps to understand how the images are named as that indicates what is installed in the image. In general, the naming scheme for the `balenalib` image set follows the pattern below:
+
 ```
 balenalib/<hw>-<distro>-<lang_stack>:<lang_ver>-<distro_ver>-(build|run)-<yyyymmdd>
```
lurch: Is it possible to specify `hw`, `distro`, `lang_stack` and `distro_ver` but **not** `lang_ver`?  Or is `lang_ver` compulsory if `lang_stack` is used?

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
...
+| Ubuntu       | 18.04 LTS (bionic)           | armv7hf, aarch64, amd64, i386                |
+| Fedora       | Fedora 30                    | armv7hf, aarch64, amd64, i386                |
+
+| Language | Default | Supported Architectures                      |
+| -------- | ------- | -------------------------------------------- |
+| Node.js  | v13.5.0 | armv6, armv7hf, aarch64, amd64, i386         |
+| Python   | v2.7.17 | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
+| OpenJDK  | v11-jdk | armv7hf, aarch64, amd64, i386, armv6         |
+| Go       | 1.13.4  | armv7hf, aarch64, amd64, i386, armv6         |
+| Dotnet   | 3.1-sdk | armv7hf, aarch64, amd64                      |
```
lurch: Some of the versions here have a `v` at the front and some don't? I guess that'll be fixed when this page is more auto-generated? :slightly_smiling_face:

pages/reference/base-images/base-images.md
lurch: There's also part of the page which says "`latest` points to the latest Python 2 version, which currently is 2.7.17" and I guess that means there's currently no way to automatically get the latest Python 3 version?

pages/reference/base-images/base-images.md
garethtdavies: Thanks for these nits as good to clean up. I've removed `v` as clearly these shouldn't be specified in the tag and could lead to confusion.

pages/reference/base-images/base-images.md
lurch: ...and there's also the Raspberry Pi 4 of course :wink: Maybe "Raspberry Pi2 and above" ?

garethtdavies: No for Pi4, as we only offer 64-bit support (and I need to explicitly mention that for Pi3 variant).  I'm checking the distros and looks like only Debian adds Raspbian package source via https://github.com/balena-io-library/base-images/blob/master/balena-base-images/device-base/raspberrypi3/debian/jessie/build/Dockerfile. Maybe @nghiant2710 can confirm this though before I update.

pages/reference/base-images/base-images.md
garethtdavies: Yes as no tags are compulsory. Will default to `latest` if no tag set. Tested via:

```
docker run -it balenalib/raspberrypi3-debian-golang:stretch-run bash
root@ed271cd90107:/go# go version
go version go1.13.4 linux/arm

root@ed271cd90107:/go# cat/etc-os-release
bash: cat/etc-os-release: No such file or directory
root@ed271cd90107:/go# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
VERSION_CODENAME=stretch
```

pages/reference/base-images/base-images.md
garethtdavies: This issue is also related as this doesn't work for `debian:sid` variant https://github.com/balena-io-library/base-images/issues/602

pages/reference/base-images/base-images.md
lurch: Ah cool! I only asked because it's not clear (from these docs) how the 'punctuation' (the hyphens and colons) would 'collapse' if you specified `hw`, `distro`, `lang_stack` and `distro_ver` but not `lang_ver`'

Could there ever be an edge-case with lang_ver and distro_ver (for the chosen lang_stack and distro) both being numeric, leading to a 'clash' between `balenalib/<hw>-<distro>-<lang_stack>:<lang_ver>` and `balenalib/<hw>-<distro>-<lang_stack>:<distro_ver>` ? (IYSWIM)
Although I'm sure all you clever folks already have a solution for that.

garethtdavies: Interesting Alpine and Python would fit the bill. I'm guessing there is a simple precedence

`docker run -it balenalib/raspberrypi3-alpine-python:3.7 bash` 

So this gives me Python 3.7 and the latest Alpine (`PRETTY_NAME="Alpine Linux v3.10"`). This makes intuitive sense (to me anyway 😄 ) given the lang ver is the first to be specified. If you wanted Alpine 3.7 and latest Python you'd have to do `docker run -it balenalib/raspberrypi3-alpine-python:latest-3.7 bash` which indeed gives Python 2.7.16 and `PRETTY_NAME="Alpine Linux v3.7"`

garethtdavies: And I guess rather than trying to explain that in text additional examples would be a good idea.

pages/reference/base-images/base-images.md
lurch: Interesting - why no Raspbian on RPi4 on balena, when the images at https://www.raspberrypi.org/downloads/raspbian/ support the RPi4?

garethtdavies: There is no balena `raspberrypi4` device unlike the Pi3, there is only `raspberrypi4-64`.

pages/reference/base-images/base-images.md
lurch: And I guess (for completeness) `balenalib/raspberrypi3-alpine-python:3.7-3.7` would give you Python3.7 on Alpine3.7?
Does the distro_ver also support a `latest` tag (i.e. `balenalib/raspberrypi3-alpine-python:3.7-latest`) or does `latest` only work for lang_ver ?

garethtdavies: Yes to the first question, 3.7 for both.

```
docker run -it balenalib/raspberrypi3-alpine-python:3.7-3.7 bash
bash-4.4# python --version
Python 3.7.4
bash-4.4# cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.7.3
PRETTY_NAME="Alpine Linux v3.7"
HOME_URL="http://alpinelinux.org"
BUG_REPORT_URL="http://bugs.alpinelinux.org"
```

Latest also works with distro so your example as expected gives Python 3.7 and latest Alpine 3.10.3

pages/reference/base-images/base-images.md
garethtdavies: I missed this one. I'd have thought using `3` would do it but that doesn't seem to work as expected:

```
docker run -it balenalib/raspberrypi4-64-alpine-python:3-3.11-run bash
Python 3.6.9
```

garethtdavies: Maybe that's the default package available i.e via `apt-get install python3`?

pages/reference/base-images/base-images.md
```diff
@@ -10,91 +10,90 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
-- For each combination of distro and stack we have two variants called `run` and `build`. The build variant is much heavier as it has a number of tools preinstalled to help with building source code. You can see an example of the tools that are included in the Debian Stretch variant [here](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/build/Dockerfile). The `run` variants are stripped down and only include a few useful runtime tools, see an example [here](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile).
-If no variant is specified, the image defaults to `run`
-- <yyyymmdd> the last optional field on tags is the date tag. This is useful for production deployments as these base images are non-moving tags, so no packages in these will update ever.
+- `<lang_ver>` is the version of the language stack, for example, Node.js 10.10, it can also be substituted for `latest`.
+- `<distro_ver>` is the version of the Linux distro, for example in the case of Debian, there are 4 valid versions, namely `sid`, `jessie`, `buster` and `stretch`.
```
lurch: Should this mention that `latest` can also be used for the `<distro_ver>` tags?

pages/reference/base-images/base-images.md
```diff
@@ -10,91 +10,90 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
-- `<distro>`  is the linux distribution, currently there are 4 distributions, namely Debian, Alpine, Ubuntu and Fedora. This field is optional, and will default to Debian if left out.
-- `<lang_stack>` is the programming language pack, currently we support Node.js, Python, OpenJDK and Go. This field is optional and if left out, no language pack will be installed, so you will just have the distribution.
+- `<hw>` is either architecture or device type and is **mandatory**. If using Dockerfile.templates, you can replace this with `%%BALENA_MACHINE_NAME%%` or `%%BALENA_ARCH%%`. For a list of available device names and architectures, see the [Device types](/reference/base-images/devicetypes/).
+- `<distro>` is the Linux distribution. Currently there are 4 distributions, namely Debian, Alpine, Ubuntu and Fedora. This field is optional and will default to Debian if left out.
+- `<lang_stack>` is the programming language pack, currently we support Node.js, Python, OpenJDK and Go. This field is optional, and if left out, no language pack will be installed, so you will just have the distribution.
```
lurch: So I guess if you specify a `<lang_ver>` but not a `<lang_stack>`, the lang_ver will just be ignored?

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
 
 ### Supported Architectures, Distros and Languages
 
-Currently balenalib supports the following OS distributions and Language stacks, if you would like to see others added, create an issue on the [balena base images repo](https://github.com/balena-io-library/base-images/issues).
+Currently, balenalib supports the following OS distributions and Language stacks, if you would like to see others added, create an issue on the [balena base images repo]({{ $links.githubLibrary }}/base-images/issues).
+
+| Distribution | Default                      | Supported Architectures                      |
```
lurch: Presumably in both this table and the next, the "Default" column is the version you'd get if you used the `latest` tag for the `<distro_ver>` or `<lang_ver>` fields?
Is it worth mentioning that explicitly? (e.g. by changing the "Default" column-header to "Default (latest)" ? ) :man_shrugging: 

---

Hmmm, in the "Multiple Distributions:" and "Multiple language stacks:" lists earlier on in the page, I wonder if it'd be worth highlighting the versions corresponding to the "latest" tag by e.g. displaying them in bold? Like:
 - [Node.js](https://nodejs.org/en/): 8.11.1, 8.17.0, 9.11.2, 10.18.0, 11.15.0, 12.14.0 and **13.5.0**
 - [Python](https://www.python.org/): **2.7.17**, 3.4.10, 3.5.7, 3.6.9, 3.7.5 and 3.8.0

pages/reference/base-images/base-images.md
```diff
@@ -10,91 +10,90 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
-`balenalib/raspberrypi3-node:10.10`
+`balenalib/raspberrypi3-node:10.18`
 
-- `<hw>` : raspberrypi3 - The raspberry pi 3 device type.
+- `<hw>` : raspberrypi3 - The Raspberry Pi 3 device type.
 - `<distro>` : omitted, so it defaults to Debian.
 - `<lang>` : node - the Node.js runtime and npm will be installed
-- `<lang_ver>` : 10.10 - This gives us node.js version 10.10.x whatever is the latest patch version provided on balenalib
-- `<distro_ver>` : omitted, so it defaults to `stretch`
+- `<lang_ver>` : 10.18 - This gives us Node.js version 10.18.x whatever is the latest patch version provided on balenalib
```
lurch: The "Multiple language stacks:" list further up the page lists "Node.js 10.18.0" but here you're using a `<lang_ver>` of `10.18`. Does that imply that `balenalib/raspberrypi3-node:10.18`, `balenalib/raspberrypi3-node:10.18.0` and `balenalib/raspberrypi3-node:10` are all equally valid base-image names? Or is there some other logic at work?

Possibly worthy of more explanation? :man_shrugging:

pages/reference/base-images/base-images.md
garethtdavies: It won't be ignored and it'd probably just mean the image doesn't exist (unless it happened to coincide with an available tag) - but either way, it's unlikely to be what you want. For example, you can see the `raspberrypi3-alpine` tags here so if you specified a <lang_ver> tag of 10.8 (e.g. node) then there simply isn't a corresponding image.

pages/reference/base-images/base-images.md
garethtdavies: Yes. You can confirm this via `docker pull balenalib/raspberrypi3-node:10.18` and inspecting the output of `docker images` which shows they are all the same image (for this specific example). As I noted earlier I think a few more examples would go a long way for this sort of information.

```
docker images
REPOSITORY                                                         TAG                   IMAGE ID            CREATED             SIZE
balenalib/raspberrypi3-node                                        10                    3bf113fa9176        15 hours ago        231MB
balenalib/raspberrypi3-node                                        10.18                 3bf113fa9176        15 hours ago        231MB
balenalib/raspberrypi3-node                                        10.18.0               3bf113fa9176        15 hours ago        231MB
```

pages/reference/base-images/base-images.md
lurch: Cool, I was just checking what the edge-case behaviour was :slightly_smiling_face:

lurch: Hmmm, that's curious - https://hub.docker.com/layers/balenalib/raspberrypi3-alpine/latest/images/sha256-054582a87ef96ba2243334500bb580246dba066967e9f6ad0c3bee655fde281d says "LABEL io.balena.device-type=raspberry-pi2" ??

pages/reference/base-images/base-images.md
```diff
@@ -108,92 +107,97 @@ CMD ["node", "main.js"]
 
 ### Supported Architectures, Distros and Languages
 
-Currently balenalib supports the following OS distributions and Language stacks, if you would like to see others added, create an issue on the [balena base images repo](https://github.com/balena-io-library/base-images/issues).
+Currently, balenalib supports the following OS distributions and Language stacks, if you would like to see others added, create an issue on the [balena base images repo]({{ $links.githubLibrary }}/base-images/issues).
+
+| Distribution | Default                      | Supported Architectures                      |
+| ------------ | ---------------------------- | -------------------------------------------- |
+| Debian       | Debian GNU/Linux 10 (buster) | armv5e, armv6, armv7hf, aarch64, amd64, i386 |
+| Alpine       | Alpine Linux v3.10           | armv6, armv7hf, aarch64, amd64, i386         |
```
lurch: Hmmm, I've just noticed that this table says that the Default version of Alpine is v3.10, and by comparing Digests on https://hub.docker.com/r/balenalib/raspberrypi3-alpine/tags you can also confirm that the `latest` and `3.10` tags point to the same image, and yet Alpine 3.11 is mentioned in the "Multiple Distributions:" list on this page, and there's a `3.11` tag on DockerHub? :confused:

pages/reference/base-images/base-images.md
lurch: Hmmm, I wonder if this _might_ explain the behaviour you discovered at https://github.com/balena-io/docs/pull/1285#discussion_r362953784 where `balenalib/raspberrypi4-64-alpine-python:3-3.11-run` points to an image containing Python 3.6.9 (whereas you'd expect it to contain Python 3.8.0) ...

**_Perhaps_** when `python:3.6.9` got added, it also added the `python:3.6` and `python:3` tags, inadvertently "overwriting" the `python:3` tag which was previously pointing to `python:3.8.0`?

That's just a guess though, and only @nghiant2710 will know for sure.

pages/reference/base-images/base-images.md
lurch: _Maybe_ that's explained by https://github.com/balena-io/contracts/commit/84035e0310ef5f14fe3fd956986091084e5ca680#r36690619 but that's a bit of a random stab in the dark :wink:

garethtdavies: Yeah looks like a mistake as mentioned in that linked contract above. In general though `latest` doesn't necessarily mean "latest" 😕https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375  In this case though it looks like it should have been tagged latest.

lurch: That's interesting, but I've no idea how and/or if Docker tags relate to the base-image tags used by Balena :wink:

garethtdavies: I'm not sure I understand that comment. The tags on Docker Hub are the base-image tags.

lurch: The medium.com article you've linked to mentions tags like `marc/test:latest` and it seems like the `latest` tag there is automatically added by Docker. I was assuming that that's actually unrelated to the `latest` tag embedded within a Balena image-name like `balenalib/raspberrypi3-alpine-python:latest-3.7` ? :man_shrugging: :man_shrugging:

garethtdavies:
Squashed and rebased. I'm keen to get this in, in its current state, as the current docs are outdated and this fixes that. I'll take some of the unresolved suggestions to a follow-up PR as I'll likely need to tweak the layout to generate this doc from the contracts anyway.

lurch:
Apologies if my comments / questions have delayed this PR, but _hopefully_ they were more of a help than a hindrance.

garethtdavies:
Oh no @lurch they have been super helpful and have helped clarify a lot of things. I didn't just want to merge this in under the assumption everything was resolved r.e. discussions above, so the note to acknowledge that. We can carry on the discussion but I don't want to delay getting at least the updated version information in 👍

pages/reference/base-images/base-images.md
lurch: Created https://github.com/balena-io-library/base-images/issues/606 to track this inconsistency

PR #1216, opened 2019-11-04
Adding page for masterclass links

garethtdavies:
Change-type: minor
Signed-off-by: Gareth Davies gareth@balena.io

PR #1295, opened 2019-12-30
docs: Embedding masterclasses into documentation

garethtdavies:
This is a WIP to look at the feasibility of incorporating the Masterclasses into the main docs. The major advantage of this is they will be searchable.

I've pulled the masterclass README via `fetch-external.sh`, so it is pulling directly from the GitHub repos so they can continue to live in balena-projects.

There is some minor formatting that would need to be done. For example, the headings should be updated for search and navigation as there are multiple h1 tags.  Some text/links might need tweaking to be generic, as could be accessed in multiple places (I've not really checked this, so may not be an issue). Image paths will also need to be updated to be absolute rather than relative as currently.

I've located it in the navigation where it was at present though obviously, we can modify the navigation as necessary.

There's currently an issue with the external docs link for the cli pages as these also match the existing `cli` page.

I pushed to staging to give an idea (though as noted above there are some formatting issues)

https://www.balena-staging.io/docs/learn/more/masterclasses/overview/

If we think this idea is worth pursuing, I'll resolve the formatting issues above in the relevant repos.

Change-type: minor


garethtdavies:
As a better example, I updated a branch of the fleet management masterclass and it is now pulling from that so see this:

https://www.balena-staging.io/docs/learn/more/masterclasses/fleet-management-masterclass/

xginn8:
I like this a lot! I think this'll make discovery much easier, and it's always better to share docs directly rather than disparate GH projects.

One small concern I have is that aiui this'll only redeploy on `docs` merge, whereas the masterclasses may change faster than the docs (this is a general concern for fetched docs).

I also think as a step 2 it'd be worth looking into pulling these automatically using a repo tag or something, as hopefully we'll have these MCs proliferate :)

garethtdavies:
Thanks @xginn8, yeh all these external docs have these issues so we need to find a good way to deploy this - webhooks might be the simplest solution there and I guess I could bring this discussion to an arch call.

I thought you would mention repo tags/topics 😆 - I do agree completely but I did a bit of work on that and couldn't make it work (including your earlier comment here https://github.com/balena-io/docs/pull/1216#issuecomment-549616375). We now have topics for all masterclasses (I've asked for a couple to be removed that are not complete) but it seems the GitHub API for topics is lacking in that regard.

The response from `https://api.github.com/search/topics?q=balena-masterclass` is simply:

```
{
    "total_count": 1,
    "incomplete_results": false,
    "items": [
        {
            "name": "balena-masterclass",
            "display_name": null,
            "short_description": null,
            "description": null,
            "created_by": null,
            "released": null,
            "created_at": "2019-12-06T20:38:30Z",
            "updated_at": "2019-12-06T20:38:30Z",
            "featured": false,
            "curated": false,
            "score": 82.770805
        }
    ]
}
```

The only way I could get the repo information was via `https://api.github.com/orgs/balena-io-projects/repos` but there doesn't seem a way to filter this via topic (other than manually parsing each result).

hedss:
Love the idea, and webhooks would be the best way to do this as you say. Watch for any repo with a `-masterclass` suffix to have a `master` merge, then pull it in automatically.

garethtdavies:
I've updated this PR to point to all the formatted branches of the PR so we can test it on staging see https://resin-docs-preview.herokuapp.com/docs/learn/more/masterclasses/overview/

I'll submit the PRs for each formatted branches in upstream repos then simply a case of updating to use those once merged. Formatting was mostly a case of fixing headers (which need to be like that for the docs) but I also tidied a few things up in line with [Markdownlint](https://github.com/DavidAnson/markdownlint) such as named fenced code blocks - there are no notable content changes.

Given the frequency of updates of these masterclasses (i.e. low) and the frequency of updates to the docs (usually daily), I'm not sure the lack of integration to trigger a build is a showstopper for this right now.

xginn8:
@garethtdavies mind updating the skeleton as well? https://github.com/balena-io-projects/balena-masterclass-skeleton

garethtdavies:
Updated now to use master branches so this is ready to merge. Deployed to staging for one final test and read through https://resin-docs-preview.herokuapp.com/docs/learn/more/masterclasses/overview/

I've created an issue to track the triggering of a docs build from external docs here https://github.com/balena-io/docs/issues/1354, which I think we can implement in due course.

tools/fetch-external.sh
```diff
@@ -4,7 +4,7 @@
 # get latest CLI docs
 cd pages/reference/ && {
   curl -O -L https://github.com/balena-io/balena-cli/raw/master/doc/cli.markdown
-  mv cli.markdown cli.md
+  mv cli.markdown balena-cli.md
```
garethtdavies: In case this needs clarification, the external docs use a regex and so I needed to make this less generic to avoid conflicts with the masterclass CLI. I added in all the relevant redirects.

garethtdavies:
@balena-ci retest

garethtdavies:
@balena-ci retest

garethtdavies:
@hedss, not sure if you had any more feedback or wanted a final review of this before it is merged in?

Preview here: https://resin-docs-preview.herokuapp.com/docs/learn/more/masterclasses/overview/

hedss:
@garethtdavies Have read through. This is great, thank you very much :)

Issue #1262, opened 2019-12-11
Outdated docs on mounting external media


garethtdavies:
Discussed here: https://www.flowdock.com/app/rulemotion/r-beginners/threads/chrtgsfaA4ugpC1IGFlhJXzD-es

Using the approach listed in the docs users were not able to mount a USB stick. The page also has outdated init system references. @tmigone has an alternate approach that works and docs should be updated to reflect this.

Related to #863 #1181

garethtdavies:
Closed via #1298

PR #1298, opened 2020-01-03
storage: Updated outdated docs on mounting external media

tmigone:
Connects-to: #1262
Change-type: patch
Signed-off-by: Tomás Migone <tomas@balena.io>


package-lock.json
```diff
@@ -4383,7 +4383,7 @@
         },
         "htmlparser2": {
           "version": "3.8.3",
-          "resolved": "https://registry.npmjs.org/htmlparser2/-/htmlparser2-3.8.3.tgz",
+          "resolved": "http://registry.npmjs.org/htmlparser2/-/htmlparser2-3.8.3.tgz",
```
lurch: Hmmm?

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,79 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
 
 ### Mounting external storage media
 
-Mounting external storage media, such as SD cards or USB thumb drives, within your application (running inside Docker) works somewhat different compared to mounting devices directly in Linux. Here we include a set of recommendations that helps you can get started.
+Since the release of multicontainer on the balenaCloud platform we no longer recommend the use of an initsystem in the container. This affects the way we deal with external storage since previously we relied on `systemd`/`OpenRC` and `/etc/fstab`.
```
lurch: Should "balenaCloud" here be using one of your funky variable-names?

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,79 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
+**Mounting**
+
+To mount an external drive you can use Linux's `mount` command. You can either mount a device by it's device name (i.e: `/dev/sda1`):
+
+```bash
+mount -t <fstype> -o rw <device-name> <mount-point>
 ```
 
-For block devices (SD cards, USB sticks), `/etc/fstab` entries would result in this error at runtime: `Running in a container, ignoring fstab device entry for ...`. Instead, you have to use [systemd .mount files](https://www.freedesktop.org/software/systemd/man/systemd.mount.html). Let's assume you want to mount an external SD card to `/mnt/storage`. Then you have to create a file with the name `mnt-storage.mount` and the content such as:
+Or mount it by it's device label (i.e: `my-usb-drive`):
```
lurch: Is mounting by UUID also supported?  (although I guess that probably doesn't make sense in a fleet-of-devices context)

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,79 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
+If your filesystem is not supported you can contact us through our [forums](https://forums.balena.io/) and we will be glad to help.
 
-Normally systemd mounts entries from `/etc/fstab` on startup automatically, but running within Docker, it will only mount entries that are not block devices, such as `tmpfs` entries. For non-block devices, adding entries `/etc/fstab` is sufficient, for example in your Dockerfile:
-```Dockerfile
-RUN echo "tmpfs  /cache  tmpfs  rw,size=200M,nosuid,nodev,noexec  0 0" >> /etc/fstab
+__Note:__ You can get a list of device names, labels and filesystem types by running `lsblk -f` (both on the host or container).
+
+**Preparing the container**
+
+In order to be able to detect external media dynamically you will need to run the container on privileged mode and enable `udevd` on it. This can be easily done if you are using [balena base images](https://www.balena.io/docs/reference/base-images/base-images/#working-with-dynamically-plugged-devices) by:
```
lurch: "on privileged mode" or "in privileged mode" ?

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,79 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
-Because of the keyword filter, `localmount` cannot be automatically started (using `rc-update add`) and won't appear in the output of `rc-status`, even when it works correctly.
+This rules will trigger everytime we plug or unplug a block partition device and run the scripts we provide (`/usr/src/mount.sh` or `/usr/src/unmount.sh`).
+
+Copy both the rules and scripts to your container:
+```Dockerfile
+COPY usb.rules /etc/udev/rules.d/usb.rules
+COPY scripts /usr/src/scripts
+```
+
+Finally we need to write the `mount.sh` and `unmount.sh` scripts. This scripts will use `mount` and `unmount` commands in the same way we described on the **Mounting** and **Unmounting** sections above.
```
lurch: "This scripts" -> "These scripts" ?

garethtdavies: Also, unmount ->umount commands...

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,79 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
+
+Copy both the rules and scripts to your container:
+```Dockerfile
+COPY usb.rules /etc/udev/rules.d/usb.rules
+COPY scripts /usr/src/scripts
+```
+
+Finally we need to write the `mount.sh` and `unmount.sh` scripts. This scripts will use `mount` and `unmount` commands in the same way we described on the **Mounting** and **Unmounting** sections above. 
+
+You can find a fully working example of automounting/unmounting devices with UDev rules on this [project](https://github.com/balena-io-playground/balena-storage).
```
garethtdavies: This one becomes {{ $links.githubPlayground }}/balena-storage

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,79 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
-Then start the `localmount` service manually in your start script:
-```Bash
-rc-service localmount start
+ACTION=="add", SUBSYSTEM=="block", ENV{DEVTYPE}=="partition", RUN+="/bin/sh -c '/usr/src/scripts/mount.sh'"
+ACTION=="remove", SUBSYSTEM=="block", ENV{DEVTYPE}=="partition", RUN+="/bin/sh -c '/usr/src/scripts/unmount.sh'"
 ```
-After running that command, the device should be mounted and ready to use in your application.
 
-Because of the keyword filter, `localmount` cannot be automatically started (using `rc-update add`) and won't appear in the output of `rc-status`, even when it works correctly.
+This rules will trigger everytime we plug or unplug a block partition device and run the scripts we provide (`/usr/src/mount.sh` or `/usr/src/unmount.sh`).
```
garethtdavies: "This rules" -> "These rules" will trigger

pages/learn/develop/runtime.md
tmigone: It is supported, however as you suggest it usually does not make much sense. Nonetheless, I'm changing this section slightly to account for this and also bringing back a general note on how to device selection. Thanks!

package-lock.json
tmigone: I have no idea how this happened. @garethtdavies perhaps related to the fact I'm not using the Dockerfile. Changing it back to `https`.

garethtdavies: I also have no idea. A brief Google search and there seem to be quite a few mentions of it but I haven't researched more than a cursory skim read of something like https://github.com/npm/npm/issues/20106

tmigone: Yep, I don't think it's worth the time tbh.

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,80 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
+Before you start it's a good idea to check if the {{ $names.os.lower }} kernel you are running was compiled with support for the filesystem you want to use. To do so, you can run this command on the **host** which will produce a list of supported filesystems: `cat /proc/filesystems`.
 
-**Using systemd**
+If your filesystem is not supported you can contact us through our [forums](https://forums.balena.io/) and we will be glad to help.
 
-Normally systemd mounts entries from `/etc/fstab` on startup automatically, but running within Docker, it will only mount entries that are not block devices, such as `tmpfs` entries. For non-block devices, adding entries `/etc/fstab` is sufficient, for example in your Dockerfile:
-```Dockerfile
-RUN echo "tmpfs  /cache  tmpfs  rw,size=200M,nosuid,nodev,noexec  0 0" >> /etc/fstab
-```
+__Note:__ You can get a list of device names, labels and filesystem types by running `lsblk -f` (both on the host or container).
```
garethtdavies: This is entirely optional but maybe this note fits better under the General Tips heading as that's where using the output of this is discussed?

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,80 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
+- Adding `privileged: true` to your container's service definition on the `docker-compose.yml` file
+- Adding `ENV UDEV=on` to your container's `Dockerfile`
+
+This will ensure that the host propagates udev events into the container, enabling us to manipulate the device from within it.
+
+**General tips for external media**
+
+Devices can be selected in many ways, for example by it's device name (`/dev` entry), label, or UUID. From a practical point of view, we recommend using labels (`LABEL=...` entries). Labels can easily be made the same across multiple cards or thumb drives, while you can still identify each device by their UUID. Also, `/dev` entries are not static on some platforms, and their value depends on which order the system brings up the devices. Device names or UUIDs are a good choice when you can easily identify or predict their values, for example within the context of a UDev rule.
```
garethtdavies: nit: it's->its

pages/learn/develop/runtime.md
```diff
@@ -249,74 +249,80 @@ At the time of writing there is an inconsistency in the behaviour of `/tmp` dire
...
-**General tips for external media**
+Copy both the rules and scripts to your container:
+```Dockerfile
+COPY usb.rules /etc/udev/rules.d/usb.rules
+COPY scripts /usr/src/scripts
+```
+
+Finally we need to write the `mount.sh` and `unmount.sh` scripts. These scripts will use `mount` and `umount` commands in the same way we described on the **Mounting** and **Unmounting** sections above. 
+
+You can find a fully working example of automounting/unmounting devices with UDev rules on this [project]({{ $links.githubPlayground }}/balena-storage).
```
garethtdavies: nit: on this project->in this project

pages/learn/develop/runtime.md
tmigone: Agreed. It was originally there before I removed that section. Now that it's back I should move it.

PR #1311, opened 2020-01-09
images: Updating base image listings

### LINKED ISSUE #1309: Links to available tags for base image builds broken
e.g. the link at the end of the second row of this table:

https://www.balena.io/docs/reference/base-images/base-images-ref/#RaspberryPi

Should be:

https://registry.hub.docker.com/r/balenalib/raspberrypi3-node/tags

not 

https://registry.hub.docker.com/u/balenalib/raspberrypi3-node/tags/manage/

Looks like all the similar links on the page need fixing.

garethtdavies:
Closes #1309 

Change-type: patch
Connect-to: #1309 


lurch:
Is it worth making `:https://registry.hub.docker.com/r/balenalib/` into one of your magic auto-expanding variables? :wink:

garethtdavies:
> Is it worth making `https://registry.hub.docker.com/r/balenalib/` into one of your magic auto-expanding variables? 😉

Yeah, either that or fully expand the links if we pull directly from https://github.com/balena-io-library/base-images/tree/master/docs (so it displays correctly there). Given this will need to be updated again very soon, I'll defer this to the next one 😄

PR #1312, opened 2020-01-09
typo: is -> are to match noun

xginn8:
Change-type: patch

### DIFF
--- a/shared/network/2.x.md
+++ b/shared/network/2.x.md
@@ -344,7 +344,7 @@ settings = dbus.Interface(proxy, "org.freedesktop.NetworkManager.Settings")
 settings.AddConnection(con)
 ```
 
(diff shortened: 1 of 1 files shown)


shared/network/2.x.md
```diff
@@ -344,7 +344,7 @@ settings = dbus.Interface(proxy, "org.freedesktop.NetworkManager.Settings")
...
+The above example will add a new WiFi connection file to your host OS **NetworkManager** configuration in `/etc/NetworkManager/system-connections`. It should be noted that you will not see a new file created in `resin-boot/system-connections` because there isn't a two way binding. The way it works is that on device boot, the files in `resin-boot/system-connections` are copied into `/mnt/state/root-overlay/etc/NetworkManager/system-connections` without removing any files from the destination. That directory is then bind mounted to `/etc/NetworkManager/system-connections` in the root filesystem on the host.
```
garethtdavies:
```suggestion
The above example will add a new WiFi connection file to your host OS **NetworkManager** configuration in `/etc/NetworkManager/system-connections`. It should be noted that you will not see a new file created in `resin-boot/system-connections` because there isn't a two-way binding. On device boot, the files in `resin-boot/system-connections` are copied into `/mnt/state/root-overlay/etc/NetworkManager/system-connections` without removing any files from the destination. That directory is then bind mounted to `/etc/NetworkManager/system-connections` in the root filesystem on the host.
```
Entirely optional but do we need `The way it works...`

PR #1313, opened 2020-01-09
integrations: Pulling in Google ioT integration from external repo

garethtdavies:
Change-type: patch


garethtdavies:
This is a draft as some tweaks needed to the upstream content

config/links.coffee
```diff
@@ -33,3 +33,4 @@ module.exports =
     "supervisor-api": 'https://github.com/balena-io/balena-supervisor/edit/master/docs/API.md'
     "update-locking": "https://github.com/balena-io/balena-supervisor/edit/master/docs/update-locking.md"
     "diagnostics": "https://github.com/balena-io/device-diagnostics/edit/master/diagnostics.md"
+    "google": "https://github.com/balenalabs/google-iot/edit/master/README.md"
```
lurch: I wonder if `google` might be a bit too generic, and using `google-iot` might be better? :man_shrugging:

garethtdavies: Yeh, this needs to be the same as the filename (which already exists as `google.md`) and I was trying to avoid a redirect. That being said `cli` is already causing me headaches due to being too generic so this may be worth implementing now.

garethtdavies:
I pushed to staging here: https://www.balena-staging.io/docs/learn/develop/integrations/google-iot/

.gitignore
```diff
@@ -20,3 +20,5 @@ pages/reference/sdk/python-sdk.md
 pages/reference/supervisor/supervisor-api.md
 pages/learn/deploy/release-strategy/update-locking.md
 pages/reference/diagnostics.md
+pages/learn/integrations/google.md
+pages/learn/integrations/google-iot.md
```
lurch: Should only one of these two lines be here? :man_shrugging:

PR #1317, opened 2020-01-14
partials: Importing partials from meta-balena repo

garethtdavies:
Change-type: patch


garethtdavies:
@xginn8 I'm trying to extract a section of text from the meta-balena README. See this FD thread https://www.flowdock.com/app/rulemotion/r-resinos/threads/wd4MfhbLhl9ZJkkH4oZ4feibGSV

I'm going to need to do this again to extract more partials from this page (e.g. config.json). This PR works (specifically this section https://github.com/balena-io/docs/pull/1317/files#diff-0164a2a83e76801e22ae67e3be5f2eefR45) but I not sure this is the best way of doing it. As a bash guru 😄any suggestions to techniques to do this that I can research?

xginn8:
> @xginn8 I'm trying to extract a section of text from the meta-balena README. See this FD thread https://www.flowdock.com/app/rulemotion/r-resinos/threads/wd4MfhbLhl9ZJkkH4oZ4feibGSV
> 
> I'm going to need to do this again to extract more partials from this page (e.g. config.json). This PR works (specifically this section https://github.com/balena-io/docs/pull/1317/files#diff-0164a2a83e76801e22ae67e3be5f2eefR45) but I am not sure this is the best way of doing it. As a bash guru smileany suggestions to additional techniques to do this that I can research?

I'd recommend a markdown extractor like https://github.com/maboiteaspam/markdown-extract perhaps:

```bash
❯ curl -q -L https://raw.githubusercontent.com/balena-os/meta-balena/development/README.md | node md-extract.js Modems
We currently test as part of our release process and provide explicit support for the following modems:
* USB modems (tested on Raspberry Pi 3, Balena Fin, Intel NUC and Nvidia TX2)
  * Huawei MS2131i-8
  * Huawei MS2372
* mPCI modems (tested on Balena Fin and Nvidia TX2 Spacely carrier)
  * Huawei ME909s-120
  * Quectel EC20
  * SIM7600E
```

garethtdavies:
Thanks @xginn8 that works really well for extracting "Modems" and "WiFi Adapters". It falls down though on config.json as only extracts that heading whereas I need all sub-headings too and this doesn't seem configurable via the binary interface 🤔 

```
➜ curl -q -L https://raw.githubusercontent.com/balena-os/meta-balena/development/README.md | ./node_modules/.bin/md-extract "config.json" 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 11093  100 11093    0     0  55883      0 --:--:-- --:--:-- --:--:-- 56025
The behaviour of balenaOS can be configured by setting the following keys in the config.json file in the boot partition. This configuration file is also used by the supervisor.
```

xginn8:
> Thanks @xginn8 that works really well for extracting "Modems" and "WiFi Adapters". It falls down though on config.json as only extracts that heading whereas I need all sub-headings too and this doesn't seem configurable via the binary interface thinking
> 
> ```
> ➜ curl -q -L https://raw.githubusercontent.com/balena-os/meta-balena/development/README.md | ./node_modules/.bin/md-extract "config.json" 
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100 11093  100 11093    0     0  55883      0 --:--:-- --:--:-- --:--:-- 56025
> The behaviour of balenaOS can be configured by setting the following keys in the config.json file in the boot partition. This configuration file is also used by the supervisor.
> ```

Hmm, interesting. Maybe we can find a better tool to precisely extract the markdown text? I think that's a better long-term solution than sed magic, but I'm having trouble finding anything decent (opened https://github.com/balena-os/meta-balena/pull/1806 to sort the British in that snippet :) )

garethtdavies:
Thanks, yeh I didn't come up with anything suitable either on a search. Isn't it ultimately going to come down to essentially the same thing as `sed` anyway i.e. simply extracting text from a pattern match? Maybe I can make the `sed` command more generic so it extracts everything up to the next same level header rather than being reliant on the specific text for the closing tag.

xginn8:
> Thanks, yeh I didn't come up with anything suitable either on a search. Isn't it ultimately going to come down to essentially the same thing as `sed` anyway i.e. simply extracting text from a pattern match? Maybe I can make the `sed` command more generic so it extracts everything up to the next same level header rather than being reliant on the specific text for the closing tag.

I think it's even a little more nuanced, you'll want to grab all text up to any header that is less than or equal to (in number of `#`'s) than the current one. I'm going to see if I can get this working in an awkscript or similar.

xginn8:
> Thanks, yeh I didn't come up with anything suitable either on a search. Isn't it ultimately going to come down to essentially the same thing as `sed` anyway i.e. simply extracting text from a pattern match? Maybe I can make the `sed` command more generic so it extracts everything up to the next same level header rather than being reliant on the specific text for the closing tag.

@garethtdavies I reckon this'll do the trick:

```bash
#!/bin/bash

HEADER=$1

match=0

while read -r line; do
    local_match=$(grep -c "#* ${HEADER}$" <<< "${line}")
    if (( local_match > 0 )); then
        match=1
        level=$(awk '/^#*/ { print gsub(/#/, "") }' <<< "${line}")
    elif (( local_match + match > 0 )); then
        new_level=$(awk '/^#*/ { print gsub(/#/, "") }' <<< "${line}")
        if (( new_level > 0 && new_level <= level )); then
            match=0
        else
            echo "${line}"
        fi
    fi
done
```

garethtdavies:
I'm not sure I am using this correctly (saving the above script to extract-markdown.sh) and running like this only outputs the following header.

```
➜ curl -q -L https://raw.githubusercontent.com/balena-os/meta-balena/development/README.md | ./extract-markdown.sh "The OS"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 11093  100 11093    0     0  52483      0 --:--:-- --:--:-- --:--:-- 52573

### SSH and Avahi services
```

FWIW I also pushed an update of the original script which extracts the three partials correctly but is prone to breaking if a header structure changes.

xginn8:
> I'm not sure I am using this correctly (saving the above script to extract-markdown.sh) and running like this only outputs the following header.
> 
> ```
> ➜ curl -q -L https://raw.githubusercontent.com/balena-os/meta-balena/development/README.md | ./extract-markdown.sh "The OS"
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100 11093  100 11093    0     0  52483      0 --:--:-- --:--:-- --:--:-- 52573
> 
> ### SSH and Avahi services
> ```
> 
> FWIW I also pushed an update of the original script which extracts the three partials correctly but is prone to breaking if a header structure changes.

@garethtdavies i updated the script above, sorry about that (missing a space)

garethtdavies:
@xginn8 That's awesome and works great! Pushed an update to incorporate those changes. Let me know if there's a better way of implementing than the way I have done via `fetch-external.sh`.

tools/fetch-external.sh
```diff
@@ -2,28 +2,54 @@
...
+  curl -O -L https://github.com/balena-io/balena-supervisor/raw/master/docs/update-locking.md
+  cd -
+}
+
+# get meta-balena README and extract partials
+cd shared/meta-balena/ && {
+  curl -O -L https://raw.githubusercontent.com/balena-os/meta-balena/development/README.md
+  mv README.md meta-balena.md
+  # Extract modem text
+  cat meta-balena.md | ../../tools/extract-markdown.sh "Modems" > supported-modems.md
```
xginn8:
```suggestion
  ../../tools/extract-markdown.sh "Modems" < meta-balena.md > supported-modems.md
```

Save A Cat™

tools/fetch-external.sh
```diff
@@ -2,28 +2,54 @@
...
+  # Extract Wifi adapters   
+  cat meta-balena.md | ../../tools/extract-markdown.sh "Wifi Adapters" > supported-wifi-adapters.md
```
xginn8: ditto

tools/fetch-external.sh
```diff
@@ -2,28 +2,54 @@
...
+  # Extract config.json
+  cat meta-balena.md | ../../tools/extract-markdown.sh "config.json" > config-json.md
```
xginn8: ditto

tools/fetch-external.sh
garethtdavies: 😹

garethtdavies:
This PR doesn't implement the config-json.md extraction as this has a couple of upstream PRs (including one that means the tests won't fail here) but included now as will be making use of this soon (and the WiFi one).

garethtdavies:
Updated now that balena-os/meta-balena#1804 has landed in master to additionally pull config.json valid fields content in remotely.

garethtdavies:
@xginn8 I just noticed when testing this, the extract script removes some formatting of the JSON i.e. it turns this:

```json
"os": {
 "network" : {
  "connectivity": {
    "uri" : "https://api.balena-cloud.com/connectivity-check",
    "interval" : "300"
  }
 }
}
```

into this:

```json
"os": {
"network" : {
"connectivity": {
"uri" : "https://api.balena-cloud.com/connectivity-check",
"interval" : "300"
}
}
}
```

Is there a way to edit the script to preserve the whitespace?

xginn8:
> @xginn8 I just noticed when testing this, the extract script removes some formatting of the JSON i.e. it turns this:
> 
> ```json
> "os": {
>  "network" : {
>   "connectivity": {
>     "uri" : "https://api.balena-cloud.com/connectivity-check",
>     "interval" : "300"
>   }
>  }
> }
> ```
> 
> into this:
> 
> ```json
> "os": {
> "network" : {
> "connectivity": {
> "uri" : "https://api.balena-cloud.com/connectivity-check",
> "interval" : "300"
> }
> }
> }
> ```
> 
> Is there a way to edit the script to preserve the whitespace?

@garethtdavies i dumped this little helper into a tiny project and just fixed your issue, check out: https://github.com/xginn8/markdown-extractor

garethtdavies:
@pdcastro would you mind giving this a review when you get a chance as follows up on https://github.com/balena-io/docs/pull/1249 and pulls and extracts the valid fields of config.json from the meta-balena repo (with some awk wizardry courtesy of @xginn8).

It does the same for supported modems ref [this thread](https://www.flowdock.com/app/rulemotion/r-resinos/threads/wd4MfhbLhl9ZJkkH4oZ4feibGSV) in addition to supported WiFi adapters.

tools/extract-markdown.sh
```diff
@@ -0,0 +1,26 @@
...
+        if (( DEBUG > 0 )); then echo "line=${line}"; fi
+        new_level=$(awk '/^#* / { print gsub(/#/, "") }' <<< "${line}")
+        if (( DEBUG > 1 )); then echo "new_level=${new_level}"; fi
+        if (( new_level > 0 && new_level <= level )); then
+            break # also could exit if you want non-greedy matching, leave as break for greedy matching
+        else
+            echo "${line}"
+        fi
+    fi
+done
```
pdcastro: My only comment here is that I would have chosen to write a Node / Javascript script instead of this shell script. Bash scripts are `awk`-ward :-)

PR #1324, opened 2020-01-19
Fix env variables

terrierscript:
I read https://github.com/balenalabs/google-iot and found may `GOOGLE_IOT_SERVICE_JSON` is outdated


garethtdavies:
Hi @terrierscript and thank you for your contribution 🎉 As it turns out there was an open PR to completely update this page by getting the content from this updated guide https://github.com/balenalabs/google-iot. I've since merged this, which in the process actually resolves this issue.

PR #1325, opened 2020-01-21
Remove a wrong note about mixpanel

roman-mazur:
All the data is submitted using our own backend if a recent enough supervisor is used.
So, no extra connection to a third-party domain.

Change-type: patch


garethtdavies:
@roman-mazur this is for 1.x versions so can a recent enough supervisor be used?

roman-mazur:
Ah, good catch. It's difficult to track when we had a working combination of the backend/supervisor back in the past. So closing.

PR #1326, opened 2020-01-21
Fix page views reporting

roman-mazur:
In December, mixpanel client update broke our reporting because it switched from
GET to POST. GET is not handled by our backend.

This change forces the supported method to be used.

Change-type: patch


nazrhom:
@balena-ci retest

templates/_track.html
```diff
@@ -15,9 +15,11 @@
...
 for(g=0;g<i.length;g++)f(c,i[g]);b._i.push([a,e,d])};b.__SV=1.2;a=f.createElement("script");a.type="text/javascript";a.async=!0;a.src="//cdn.mxpnl.com/libs/mixpanel-2-latest.min.js";e=f.getElementsByTagName("script")[0];e.parentNode.insertBefore(a,e)}})(document,window.mixpanel||[]);
 
-window.jQuery.get('https://api.resin.io/config', (config) => {
+var baseUrl = 'https://api.balena-cloud.com';
+window.jQuery.get(baseUrl + '/config', (config) => {
   mixpanel.init(config.mixpanelToken, {
```
lurch: On a totally unrelated note, that `https://api.balena-cloud.com/config` URL currently returns 141KB of data. That seems like an _awful_ lot of data to download when all that's actually needed is the `mixpanelToken` value :wink:

roman-mazur: Fair enough :) We should improve it too.

roman-mazur:
@balena-ci retest please

Issue #1213, opened 2019-10-31
Update image partition layout 


garethtdavies:
https://www.balena.io/os/docs/architecture/#Image-Partition-Layout is outdated and the layout is actually https://github.com/balena-os/meta-balena/blob/development/meta-balena-common/classes/image_types_resin.bbclass#L26

lurch:
Also, the text on that page says 'resin-state' but the images say 'resin-conf'.

lurch:
Closed by #1333 ?

garethtdavies:
Thanks @lurch but not quite as I need to copy this all over to the balena.io/os site (which is what this issue originally referenced). I just need to update the last bit of https://www.balena.io/docs/reference/OS/overview/2.x regarding Yocto and then I'll push all the changes over and close this.

lurch:
Ahhh, I didn't spot the `/os/` part of the original URL :wink: 
Rather than having the docs manually copied from one site (and/or repo?) to the other, is this another place where your "copy external docs into this repo" script could come in useful?

garethtdavies:
That would be ideal but that's not currently possible in the same way as the others have been implemented (e.g. the /os/ site is private). Agree though copying from one to the other isn't a good long term solution so hopefully will resolve this sooner rather than later.

garethtdavies:
This will be closed via https://github.com/balena-io/resin-site/pull/954

PR #1333, opened 2020-01-27
os: Update image partition layout

garethtdavies:
Updates text and images on https://www.balena.io/docs/reference/OS/overview/2.x/ for image partition layout and state partition. 

Change-type: patch
Connects-to: #1213 


garethtdavies:
Probably one for @ZubairLK or @alexgg to review

PR #1336, opened 2020-01-30
docs: Expand documentation on granting support access

### LINKED ISSUE #1127: Expand docs on enabling support access
currently we only have https://www.balena.io/docs/learn/manage/actions/#grant-support-access-to-application

We should expand on that explaining how to enable per app, per device and the implications of enabling it, linking to security docs.

garethtdavies:
Closes #1127 

Change-type: patch
Connects-to: #1127


garethtdavies:
@balena-ci retest

PR #1344, opened 2020-02-04
outline development anti-patterns

xginn8:
Change-type: patch

### DIFF
--- a/pages/learn/more/anti-patterns.md
+++ b/pages/learn/more/anti-patterns.md
@@ -1,0 +1,86 @@
+---
+title: Development Anti-patterns
+---
+# Anti-patterns, or how to break your {{ $names.os.lower }} devices
+
+As you are building your application and preparing for production, there are a number of things to keep in mind that can cause problems on a device. Many of these items may seem obvious or even desired, but often cause more trouble than they are worth.
+
+## Networking
+One of the biggest causes of issues during deployment is either a misconfigured local network or otherwise a network that does not adhere to the {{ $names.os.lower }} [networking requirements][networking-reqs].
+
+### Connectivity dependent on container (e.g. kernel module for wifi dongle, or udev rule for modem)
+Since there can be cases where application containers fail to start, having any networking configuration depend on those application containers without a fallback is generally discouraged. Using a project like [wifi-connect][wifi-connect] can help configure the networking stack after launch, but devices should be connected as reliably as possible at boot.
+
+### Iptables rules that block supervisor/VPN traffic
+It is enticing to use custom iptables rules to filter traffic before it hits the application, to limit the surface of the API. Those rules combined with host networking however can prevent the supervisor or VPN from successfully connecting to {{ $names.cloud.lower }}, and therefore should be treated with caution. For a full list of networking requirements, consult [this document][networking-reqs].
+
+### Manually changing the system clock, or blocking NTP requests
+Since {{ $names.os.lower }} and the supervisor communicate with {{ $names.cloud.lower }} using an HTTPS API, it is important that time is synchronized on the device. If the system date/time drifts substantially, SSL certificate validation may fail and the device may unexpectedly lose the ability to reach HTTPS websites or update the {{ $names.cloud.lower }} web dashboard, and may even no longer be reachable over ssh or the {{ $names.cloud.lower }} VPN. {{ $names.os.upper }} provides a [number of mechanisms][time-sync] to keep time as up-to-date as possible, but ensuring [NTP is accessible over the network][networking-reqs] is critical.
+
+## Local storage
+
+### Writing to files in the container file system (and not a volume)
+Commonly, users misunderstand the distinction between the container's file system and a volume. Named volumes are preserved across updates unless specifically dereferenced or destroyed, while anything written to the container's file system will be purged during the next application update.  In addition, the container's file system often uses the AUFS driver, which typically has worse performance both in disk space and CPU utilization when compared to a named volume's standard Linux ext4 file system. For an in-depth comparison of the two, it is recommended to complete the following [services masterclass][services-mc]. Additionally, the [supervisor provides an API][supervisor-api] to manage named volumes.
+
+### Running out of disk space
+One of the most common issues moving from a project to a product is considering what is written to local disk. Often, either logs or important data are written locally but without giving thought to the data's lifecycle. It is important to ensure that some mitigations are in place to prevent any local storage from filling. [Some examples include][logging-solutions]:
+* configuring logrotate
+* external logging service
+* logging only to stderr/stdout to take advantage of {{ $names.engine.lower }}'s built-in logging features
+
+If the data partition (`/mnt/data`) fills up completely, {{ $names.engine.lower }} may not be able to restart the application container on device reboot, new application releases may fail to apply, and the only way to recover the device may be by deleting the application image through manual intervention, including loss of all data written to the container's file system (named volumes are preserved).
+
+### Excessive writes causing SD / Flash failure / corruption
+If the device type uses flash memory or an SD card, any write-heavy workload will cause that storage medium to wear prematurely and may cause devices to crash irrecoverably. Limiting any writing to temporary locations in-memory is one strategy to avoid having to replace SD cards in the field. There are further SD card recommendations [available here][sd-cards]
+
+## Host OS services
+
+### Modifying the OS image
+Any modifications made to the host OS before flashing or at runtime carry undue risk. If the host OS has been modified in some way (other than connection details), any subsequent behavior cannot be guaranteed.
+
+### Modifying the host’s systemd services via D-Bus
+While {{ $names.os.lower }} provides lots of flexibility in terms of interacting with the host OS via D-Bus, with great power comes great responsibility. Since these interactions can potentially cause {{ $names.os.lower }} devices to disconnect from {{ $names.cloud.lower }} or even fall offline, these API calls should be carefully guarded and considered in the application code.
+
+### DoSing the supervisor with API requests
+Since the supervisor is the brain of the update process for {{ $names.os.lower }} devices, any interactions with the API provided should be carefully considered and rate-limited. If a container engages in denial-of-service style behavior, the supervisor may be unable to recover and perform any necessary updates.
+
+### Incorrect config.txt/config.json settings
+These two configuration files are critical for the boot process on most devices. Specifically, a malformed `config.json` will cause any {{ $names.os.lower }} devices to not boot properly, and a mangled `config.txt` (on certain Raspberry Pi-based devices) may cause a loss in functionality at boot. There are some tools that can help manage these files across devices, like [configizer][configizer] and [dtparam tuneables][dtparams].
+
+## Managing Resources
+
+### Creating a reboot loop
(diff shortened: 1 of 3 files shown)


xginn8:
leaving the WIP while sections are still missing, but @garethtdavies wdyt on the completed bits?

pages/learn/welcome/production-plan.md
```diff
@@ -118,6 +118,10 @@ When updating software on production devices, it makes sense to be cautious and
 
 The {{ $names.company.lower }} API includes functionality that allows for [fine-grained control of updates][staged-releases]. With these endpoints, you can turn off automatic release updates, set a target release for the whole fleet, or set target releases for a specific subset of devices. The [device tag][tags] and [filter][filters] functionality makes it easy to assign groups of devices to certain phases of your deployment.
 
+## Further reading
+
+For more information about common anti-patterns, see [How to break balenaOS (development anti-patterns)][anti-patterns].
```
xginn8: @garethtdavies i had to remove the templatized reference to balenaOS here as it broke the build, not sure if that's a known issue or not.

pdcastro:
@xginn8, I've just thought of another (anti-)pattern (?) for your consideration: users should make a conscious decision regarding pinning their application's base image to a specific version, for example:
```
FROM balenalib/raspberrypi3-debian-node:8-run-20200115
```
instead of:
```
FROM balenalib/raspberrypi3-debian-node:8-run
```
In past support threads, what I found is that users would one day, perhaps several months after a previous release, push an update (perhaps an urgent fix to an app bug), and what they find is that "things break" (the build fails) and they panic because "they haven't changed anything", "what did balena do?!". They don't realise that a new base image version is being used because they never pinned it to a specific dated release.

Alternatively, users are actively pushing updates on a daily basis (during development) and occasionally they find that "everything rebuilt from scratch" and took a long time, and they ask us "why didn't the balenaCloud builder use the cache?", "our app is very large and takes a long time to build, please fix the cache". Well, the cache wasn't used because there was an update to the base image, without the user realising.

pages/learn/welcome/production-plan.md
garethtdavies: Interesting and I've not seen that before. I actually built this locally with the variable syntax and it seemed to work `[How to break {{ $names.os.lower }} (development anti-patterns)][anti-patterns]`

xginn8: Well, I'll be, i don't know what i'm doing differently now but it's working fine. sorry for the trouble!

xginn8:
> @xginn8, I've just thought of another (anti-)pattern (?) for your consideration: users should make a conscious decision regarding pinning their application's base image to a specific version, for example:
> 
> ```
> FROM balenalib/raspberrypi3-debian-node:8-run-20200115
> ```
> 
> instead of:
> 
> ```
> FROM balenalib/raspberrypi3-debian-node:8-run
> ```
> 
> In past support threads, what I found is that users would one day, perhaps several months after a previous release, push an update (perhaps an urgent fix to an app bug), and what they find is that "things break" (the build fails) and they panic because "they haven't changed anything", "what did balena do?!". They don't realise that a new base image version is being used because they never pinned it to a specific dated release.
> 
> Alternatively, users are actively pushing updates on a daily basis (during development) and occasionally they find that "everything rebuilt from scratch" and took a long time, and they ask us "why didn't the balenaCloud builder use the cache?", "our app is very large and takes a long time to build, please fix the cache". Well, the cache wasn't used because there was an update to the base image, without the user realising.

That's a great one @pdcastro, and very reminiscent of what Greg advised [here](https://forums.balena.io/t/im-a-balena-fleetoperations-engineer-amaa/30792/8) too:

> For code development, pinning versions of libraries, OS versions, base images, as much as possible, so things don’t change unexpectedly on a repush (when silently new versions are pulled in).

I will add that to the doc!

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Creating a reboot loop
+One common anti-pattern is to trigger reboots if some condition is met (using the supervisor API or D-Bus directly).  Often this "fix" can cause reboot loops that are unexpected and cause more trouble than they fix. Nearly all components of {{ $names.os.lower }} are health-checked and restarted if found to be non-functional, including the [operating system itself][watchdog] (when available).
+
+### Causing an out-of-memory (OOM) scenario
+When developing on low-footprint devices or for a heterogeneous fleet, it is easy to forget about managing memory consumption. Due to the nature of [Linux's out-of-memory killer][linux-oom], these events are often difficult to catch and debug. Individual services can be limited by {{ $names.engine.lower }} using [directives in `docker-compose.yml`][docker-compose].
+
+## Managing Applications
+
+### Not pinning base image versions
```
xginn8: @pdcastro specifically I folded in your comments here

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+FROM balenalib/raspberrypi3-debian-node:8-run-20200115
+```
+
+instead of:
+
+```
+FROM balenalib/raspberrypi3-debian-node:8-run
+```
+
+During app development, pinning versions of various components as much as possible prevents pieces from changing unintentionally later on during a redeployment when new versions are silently pulled in (like libraries, OS versions, base images, etc). In the same vein, pinning versions encourages builder cache reuse as components cannot change underneath the build process itself, ensuring a more reliable and repeatable build.
```
pdcastro: 👍

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Connectivity dependent on container (e.g. kernel module for wifi dongle, or udev rule for modem)
+Since there can be cases where application containers fail to start, having any networking configuration depend on those application containers without a fallback is generally discouraged. Using a project like [wifi-connect][wifi-connect] can help configure the networking stack after launch, but devices should be connected as reliably as possible at boot.
+
+### Iptables rules that block supervisor/VPN traffic
+It is enticing to use custom iptables rules to filter traffic before it hits the application, to limit the surface of the API. Those rules combined with host networking however can prevent the supervisor or VPN from successfully connecting to {{ $names.cloud.lower }}, and therefore should be combined with extreme scrutiny. For a full list of networking requirements, consult [this document][networking-reqs].
+
+### Running docker-in-docker with host networking and iptables enabled
+(can be folded into the previous issue, leaving for completeness at review)
```
pdcastro: Don't forget. :-)

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Manually changing the system clock, or blocking NTP requests
+Since {{ $names.os.lower }} & the supervisor communicate with {{ $names.cloud.lower }} using an HTTP API, it is important that time is synchronized on the device to ensure that certificates are trusted. {{ $names.os.upper }} provides a [number of mechanisms][time-sync] to keep time as up-to-date as possible, but ensuring [NTP is accessible over the network][networking-reqs] is critical.
```
pdcastro: * HTTP API → HTTPS API  (or "HTTP API over HTTPS", but sounds a bit pedantic...).  My point being that HTTPS (or SSL) are important to mention.
* certificates → SSL certificates
* certificates are trusted → certificate validation succeeds
* We know what it means (I think I do), :-) but users would benefit from expanding a bit on "why" time is relevant for SSL certificates. What about:

> Since {{ $names.os.lower }} and the supervisor communicate with {{ $names.cloud.lower }} using an HTTPS API, it is important that time is synchronized on the device. If the system date/time drifts substantially, SSL certificate validation may fail and the device may unexpectedly lose the ability to reach HTTPS websites or update the {{ $names.cloud.lower }} web dashboard, and may even no longer be reachable over ssh or the {{ $names.cloud.lower }} VPN. {{ $names.os.upper }} provides a [number of mechanisms][time-sync] to keep time as up-to-date as possible, but ensuring [NTP is accessible over the network][networking-reqs] is critical.

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Local storage
+
+### Writing to files in the container file system (and not a volume)
+Commonly, users misunderstand the distinction between the container's file system and a volume. Named volumes are preserved across updates unless specifically dereferenced or destroyed, while anything written to the container's file system will be purged during the next application update. For an in-depth comparison of the two, it is recommended to complete the following [services masterclass][services-mc]. Additionally, the [supervisor provides an API][supervisor-api] to manage named volumes.
```
pdcastro: I understand there are performance and disk space implications as well, especially in the AUFS case. Suggestion:
> [...] during the next application update. In addition, the container's filesystem often uses the [AUFS driver](https://docs.docker.com/storage/storagedriver/aufs-driver/), which typically has worse performance both in disk space and CPU utilization when compared to a named volume's standard Linux [ext4 filesystem](https://en.wikipedia.org/wiki/Ext4). For an in-depth comparison [...]

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Running out of disk space
+One of the most common issues moving from a project to a product is considering what is written to local disk. Often, either logs or important data are written locally but without giving thought to the data's lifecycle. It is important to ensure that some mitigations are in place to prevent any local storage from filling. [Some examples include][logging-solutions]:
+* configuring logrotate
+* external logging service
+* logging only to stderr/stdout to take advantage of {{ $names.engine.lower }}'s built-in logging features
+
```
pdcastro: Correct but perhaps not scary enough. :-) I suggest adding some "motivation" here:
> If the `/mnt/data` partition is allowed to fill up completely, balenaEngine may not be able to restart the application container on device reboot, new application releases may fail to apply, and the only way to recover the device may be by deleting the application image through manual intervention (ssh to the host OS), including loss of all data written to the container's filesystem (named volumes are preserved).

pages/learn/more/anti-patterns.md
```diff
@@ -0,0 +1,96 @@
...
+## Host OS services
+
+### Udev rules in container that conflict with host
+needs more information
+
+### Modifying the OS image
+Any modifications made to the host OS before flashing or at runtime carry undue risk. If the host OS has been modified in some way (other than connection details), any subsequent behavior cannot be guaranteed.
+
+### Overwriting the D-Bus socket
+needs more information
```
pdcastro: Indeed. I see there are few of these lines around.

xginn8: This list was cribbed from @shaunmulligan's notes, but in some cases I'm not 100% what he means or has seen before. If this one (or the others) seems obvious to you, please help me write it!

pages/learn/more/anti-patterns.md
xginn8: TIL! thanks for pointing that out :)

pages/learn/more/anti-patterns.md
xginn8: That looks good to me! Added.

pages/learn/more/anti-patterns.md
xginn8: Thanks for pointing these out, I think these are good changes. Updated!

PR #1346, opened 2020-02-05
primer: Updating balena primer for push workflow

garethtdavies:
Updating to replace git workflow and updated screenshot for device summary page.

Change-type: patch


pages/learn/welcome/primer.md
```diff
@@ -29,11 +29,11 @@ First, the device connects to the network and performs its early provisioning, w
 
 ## Code deployment
 
-Code deployment begins when you type `git push {{ $names.company.short }} master` in your command line, moving your application code from a local repository to the {{ $names.company.lower }} platform. Here's what the process looks like:
+`balena push` is the recommended method for deployment and [development](/learn/develop/local-mode/) on the balenaCloud platform. To use `balena push` you need to first [install the balena CLI](/reference/cli/#install-the-cli) and ensure you are logged in to your account with `balena login`.
 
-<img src="/img/common/how_deploy_works.png" width="80%">
+When you run the  `balena push <APP_NAME or DEVICE_IP>` command from your development machine it will essentially take your project (or repository) folder, compress it and send it to the [balenaCloud build server](/learn/deploy/deployment/#the-balenacloud-build-server) or local balenaOS device in [localMode](/learn/develop/local-mode/) where it will be built.
```
rahul-thakoor: not sure about `DEVICE_IP` here since I believe it will be a device in local mode and won't use the cloud builders

garethtdavies: Right, it's trying to cover both cases hence the "or local balenaOS device in local mode". I'll update this to just be the cloud and then as a separate sentence say you can also build on the device in local mode to be more clear but give them a link to learn more...

I need to update this section for variables as well which was taken from the deployment section.

PR #1349, opened 2020-02-11
base-images: Bumping node lang versions

garethtdavies:
Reference: https://github.com/balena-io/contracts/pull/146

Change-type: patch


garethtdavies:
@balena-ci retest

PR #1350, opened 2020-02-11
base-images: Adding nitrogen8mm-dwe to machine names and architectures

garethtdavies:
Ref: https://github.com/balena-io/contracts/pull/143

The table is also formatted hence the diff output.

Change-type: patch


garethtdavies:
@balena-ci retest

PR #1351, opened 2020-02-11
base-images: Updating base image lists from contracts

garethtdavies:
Updated this to be generated directly from the contracts repo rather than the standalone script that was being used before but was missing several devices. As per [this comment](https://github.com/balena-io/contracts/pull/144#issuecomment-581312116) this device was excluded. 

I ran a [broken link checker](https://www.npmjs.com/package/broken-link-checker) tool on this and can confirm there are 0 broken links (4410 good ones). Otherwise, I've kept the layout consistent with that was there before. 

For reference, it was generated via [this](https://github.com/garethtdavies/base-images-docs-generator/blob/master/base-images-ref.py). This ideally needs refactoring but it works for now...

Change-type: patch


xginn8:
@garethtdavies i will approve you if you move https://github.com/garethtdavies/base-images-docs-generator/ to _at least_ the playground

garethtdavies:
Right, I was going to do that after a bit of refactoring but I can move it now as it is a playground :)

PR #1362, opened 2020-02-17
redirect: Add redirect for Masterclass overview

garethtdavies:
Change-type: patch


garethtdavies:
@balena-ci retest

PR #1372, opened 2020-02-24
reference: Update API resources for application tag

garethtdavies:
Added a missing resource of application tag. 

Is there an endpoint that lists all available resources or where would be the easiest place to look to track down any additional missing endpoints?

Change-type: patch


garethtdavies:
@balena-ci retest

Issue #816, opened 2018-05-29
Updated instructions for modifying config.txt


zwalchuk:
Modifying config.txt directly can only be done before first boot, after that the settings are tracked in the cloud

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_bvqr)

Issue #675, opened 2018-01-08
Document configuration for using dtb overlay to switch from uart0 to uart1


zwalchuk:
If users load a dtb overlay to switch from uart0 to uart1 (the pi3-miniuart-bt overlay
for example), this would make bluetooth use the MiniUART instead of the PL011 UART.
Using this overlay, users should also set core_freq=250 or force_turbo=1 in config.txt

Issue #296, opened 2016-08-17
Document how config.txt additions are applied and the device behaviour


shaunmulligan:
We need to detail the mechanism for how the supervisor applies changes to the config.txt file.

we also need to mention the current caveats and limitations.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_kii)

shaunmulligan:
also need to mention that changing or adding to /boot in the container will have no affect on the device.

vipulgupta2048:
Closing #296 via #1378

PR #1378, opened 2020-02-27
Updating advanced boot configuration for Raspberry Pi

garethtdavies:
Change-type: patch
Connects-to: #816
Connects-to: #675
Connects-to: #296


pages/reference/OS/advanced.md
```diff
@@ -109,26 +106,31 @@ dtoverlay=i2c-rtc,ds1307
...
 
 To disable the Raspberry Pi rainbow splash screen, add the `disable_splash=1` entry to `config.txt`.
 
 __Note:__ This setting disables the Raspberry Pi rainbow splash screen but does not disable the {{ $names.company.lower }} logo splash screen. If you would like to replace the {{ $names.company.lower }} logo with your custom splash logo, replace `splash/resin-logo.png` located in the [boot partition][boot-partition] of the image.
 
 ### Further Reading
 
-There are more details on the options available in `config.txt` over at
-[elinux's RPi Config page][elinux].
+There are more details on the options available in `config.txt` over at [elinux's RPi Config page][elinux].
```
lurch: The eLinux page is very out of date - you ought to be (only) linking to https://www.raspberrypi.org/documentation/configuration/config-txt/README.md

shared/general/config-variables-pi
```diff
@@ -0,0 +1,6 @@
+Name | Default | Description
+--- | --- | ---
+RESIN_HOST_CONFIG_disable_splash | 1 | Enable / Disable the {{ $names.company.lower }} splash screen
```
lurch: There's another file in this PR where it says about `disable_splash` ... "This setting disables the Raspberry Pi rainbow splash screen but does not disable the {{ $names.company.lower }} logo splash screen." - seems to be some contradiction here?

pages/reference/OS/advanced.md
```diff
@@ -109,26 +106,31 @@ dtoverlay=i2c-rtc,ds1307
...
 
-[elinux]:http://elinux.org/RPiconfig
+[boot-partition]:/reference/OS/overview/2.x/#image-partition-layout
+[cli]:/reference/cli/reference/balena-cli/#envs
+[config-txt]:https://www.raspberrypi.org/documentation/configuration/config-txt/
 [config-vars]:/learn/manage/configuration
+[config-vars-fleet]:/learn/manage/configuration/#managing-fleet-configuration-variables
+[config-vars-device]:/learn/manage/configuration/#managing-device-configuration-variables
+[device-provisioning]:/learn/welcome/primer/#device-provisioning
+[device-tree-overlay]:https://github.com/raspberrypi/linux/blob/rpi-4.1.y/arch/arm/boot/dts/overlays/README
```
lurch: `rpi-4.1.y` is very old - this probably ought to be updated to https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/overlays/README ? (or whatever kernel branch it is that BalenaOS uses)

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
 
 ```
-gpu_mem=16
 gpu_mem_256=64
 gpu_mem_512=128
+gpu_mem_1024=256
 ```
 
-Each of these values is specified in megabytes.
+`gpu_mem` is the default amount of memory, which, by default, is set to 16MB (specified by `gpu_mem=16`). This may well be less than you require depending on your application (particularly applications that make heavy use of the Raspberry Pi's graphics card). As per the [Raspberry Pi documentation][gpu-memory], values of `gpu_mem` over 512 are not recommended, will provide no performance improvements, and are untested.
```
lurch: "graphics card" -> "graphics capabilities"? (The Pi doesn't really have a "graphics card" in the traditional sense)

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
-__Note:__ Configuration variables defined through the API will not apply to devices in [local mode][local-mode].
+<img alt="Add fleet configuration variable" src="/img/configuration/variable_editor_config.png" width="60%">
+
+__Note:__ In addition to the dashboard, these configuration variables can be also be set using the API or any of its clients, including the [SDK][sdk] and [CLI][cli].
 
 ### GPU Memory
 
-You can configure the amount of memory shared with the GPU by adding entries to
-`config.txt` as follows:
+The [amount of memory][gpu-memory] that is addressable from the GPU my be configured by adding entries to `config.txt`. You can target different memory sizes, specified in MB, of 256, 512, and 1024 corresponding to the RAM on the Raspberry Pi. Values will be ignored for those devices whose memory size does not match. For the Raspberry Pi 4, which has versions with RAM greater than 1GB, the minimum and maximum values are the same as for a 1GB device.
```
lurch: "You can target different memory sizes, specified in MB, of 256, 512, and 1024 corresponding to the RAM on the Raspberry Pi." doesn't really make much sense - perhaps it could be reworded? Maybe something like "You can also set specific values of gpu_mem for Raspberry Pis with 256, 512 or 1024 MB of RAM." ?

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
 
 ```
-gpu_mem=16
```
lurch: IMHO `gpu_mem` should be left in. It's a "global value" which is applied if one of the more memory-specific options isn't supplied.

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
-#### Modifying `config.txt` locally before the first boot
+```shell
+$ balena ssh <uuid>
+$ cat /mnt/boot/config.txt
+```
 
-The `config.txt` is located in the root of the `{{ $names.company.lower }}-boot` partition, and you can modify it by mounting the SD card on a computer. This will only work if you edit the SD card before the device's first boot.
+### Modifying `config.txt` locally before the first boot
 
+Before the device is [provisioned][device-provisioning], you may edit `config.txt` by mounting a flashed SD card (with the partition label `resin-boot`) and editing the file directly. This will only work if you edit the file before the device's first boot as after device provisioning, any changes will be overwritten by the device supervisor with values read from the {{ $names.cloud.lower }} API.
```
lurch: NIT: should the comma be moved to just after "device's first boot" ?

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
 
-__Warning:__ This page contains details of the more *involved* configuration
-options that expose more functionality but potentially enable you to break
-things should you configure them incorrectly - tread carefully!
+__Warning:__ This page contains details of advanced configuration options that expose more functionality, but any mistakes may potentially leave a device inaccessible.
 
 ## Raspberry Pi
 
-### config.txt
+The Raspberry Pi exposes device [configuration options][config-txt] via a text file on the [boot partition][boot-partition] named [`config.txt`][config-txt]. You can change boot options simply by editing this file, either by manually editing before the device's first boot or editing the device [configuration variables][config-vars]. By default, the following values are set for Raspberry Pi devices:
```
lurch: "simply by editing this file" -> "in this file" ?
"manually editing" -> "manually editing it" ?

lurch:
I'm not sure if this PR actually addresses #675 ?

garethtdavies:
> I'm not sure if this PR actually addresses #675 ?

Many thanks for the review, and I'll address those changes shortly. As for the issue, you are right and I had already raised this in an internal chat for further investigation as I wanted this PR to resolve that issue one way or another, so I linked it here for reference. So when I get some feedback I'll update this PR (or remove that `connect-to` if no longer relevant).

pages/reference/OS/advanced.md
garethtdavies: Given this is already linked a couple times in the first paragraph I've just removed this section.

shared/general/config-variables-pi
garethtdavies: Great catch, this was pulled in from another part of the docs and not updated (this was also recently clarified in the dashboard).

pages/reference/OS/advanced.md
chrisys: I think it's worth clarifying a bit further here, added a suggestion but feel free to reword:

```suggestion
Before the device is [provisioned][device-provisioning], you may edit `config.txt` by mounting a flashed SD card (with the partition label `resin-boot`) and editing the file directly. This will only work if you edit the file before the device's first boot as after device provisioning, any changes will be overwritten by the device supervisor with values read from the {{ $names.cloud.lower }} API. For this reason if you wish changes added before the first boot to persist, they must also be specified in the dashboard.
```

garethtdavies: Yes definitely, I wasn't aware of that. I assumed they would be added during provisioning.

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
+  - 1 start and 1 stop bit.
+  - No parities.
+  - Break generation.
+  - 8 symbols deep FIFOs for receive and transmit.
+  - SW controlled RTS, SW readable CTS.
+  - Auto flow control with programmable FIFO level.
+  - 16550 like registers.
+  - Baudrate derived from system clock.
+
+To enable the mini UART, an entry should be added to `config.txt` as follows:
```
chrisys: This may raise questions about the fact we have a toggle for this in device configuration - how does that toggle relate to this setting?

pages/reference/OS/advanced.md
lurch: Aha! Reading the old version of this page, it sounded like the supervisor just "appended" new values to config.txt, but I guess you're saying that the supervisor just overwrites the entirety of config.txt?

garethtdavies: @chrisys I'm confused by this comment as I just tested this and works as initially outlined but I might have missed something.

To test, I simply created a new application and added a new device (Raspberry Pi 4). I flashed the image with Etcher and edited `config.txt` before the first boot. I added `start_x=1` and then provisioned the device. As expected a custom configuration value was already created for me in the dashboard, so they are persisted so long as they are created prior to device provisioning.

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,91 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
+The BCM2837 on the Raspberry Pi 3 has 2 built-in [UARTs][uart] (as did its predecessors), however, to support the Bluetooth functionality the fully-featured PL011 UART was moved from the header pins to the Bluetooth chip and the mini UART made available on header pins 8 & 10.
 
-By default we assign 16MB of memory to the GPU (specified by `gpu_mem`.) This
-may well be less than you require depending on your application (particularly
-applications which make heavy use of the Pi's graphics card.)
+This has a number of consequences for users of the serial interface:
 
-### Enable serial interface
+- The /dev/ttyAMA0 previously used to access the UART now connects to Bluetooth.
+- The mini UART is now available on /dev/ttyS0, disabled by default.
```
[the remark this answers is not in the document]
garethtdavies: I just tested this and yes that's correct for a development image and disabled on prod. Will update.

pages/reference/OS/advanced.md
garethtdavies: I tested again on a production image with some more obscure values to double check and they get added to the dashboard automatically if you do this prior to device provisioning. 

CleanShot 2020-03-02 at 10 28 37@2x

pages/reference/OS/advanced.md
garethtdavies: I'll update this for clarity but essentially it's the same as other config options we have prepopulated i.e. if you edit this before device provisioning it'll be respected. After device provisioning, it has to be done via the dashboard. Tested this by enabling UART `enable_uart=1` on a production image prior to first boot and it worked.

This also applies to Rahul's comment as for dev images this is already enabled.

pages/reference/OS/advanced.md
garethtdavies: @lurch so in summary, as you point out, yes the supervisor does overwrite the file rather than append but if you add these before the device is provisioned these are created in the dashboard (API) so the supervisor is aware of them so they won't be removed. If you edit `config.txt` after provisioning any changes you make are simply going to get overwritten by the supervisor so you must use the dashboard.

lurch: If that's indeed the case, it's probably worth explicitly mentioning in the docs that any values manually added to config.txt before device-provisioning are also automatically added to the dashboard when the device is provisioned?

garethtdavies:
I also think #675 is covered by the content [here](https://www.raspberrypi.org/documentation/configuration/uart.md) which can be linked to.

pages/reference/OS/advanced.md
```diff
@@ -22,7 +22,7 @@ $ cat /mnt/boot/config.txt
 
 ### Modifying `config.txt` locally before the first boot
 
-Before the device is [provisioned][device-provisioning], you may edit `config.txt` by mounting a flashed SD card (with the partition label `resin-boot`) and editing the file directly. This will only work if you edit the file before the device's first boot, as after device provisioning any changes will be overwritten by the device supervisor with values read from the {{ $names.cloud.lower }} API.
+Before the device is [provisioned][device-provisioning], you may edit `config.txt` by mounting a flashed SD card (with the partition label `resin-boot`) and editing the file directly. Any values added to `config.txt` will be added to the dashboard during device provisioning. This will only work if you edit the file before the device's first boot, as after device provisioning, any changes will be overwritten by the device supervisor with values read from the {{ $names.cloud.lower }} API.
```
lurch: In one sentence it refers to "the dashboard" and in another sentence it refers to "the balena API". I guess these are more or less the same thing, but perhaps this could be clarified / tidied up?

garethtdavies:
@shaunmulligan would you be able to have a review of this please, particularly the UART section? There are quite a few changes so have previewed it here: https://resin-docs-preview.herokuapp.com/docs/reference/OS/advanced/

garethtdavies:
@balena-ci rebase

garethtdavies:
Thanks, let's get this in as clearly an improvement on existing content and can follow up if anything needs additional clarification.

pages/reference/OS/advanced.md
```diff
@@ -5,94 +5,94 @@ excerpts: Configuration options to expose more device functionality to {{ $names
...
+  - 1 start and 1 stop bit.
+  - No parities.
+  - Break generation.
+  - 8 symbols deep FIFOs for receive and transmit.
+  - SW controlled RTS, SW readable CTS.
+  - Auto flow control with programmable FIFO level.
+  - 16550 like registers.
+  - Baudrate derived from system clock.
+
+The mini UART is enabled by default for development images. For production images either enable it using [configuration variables][config-vars] or before device provisioning by adding the following entry to `config.txt`:
```
lurch, line 83: Perhaps the earlier line that says "This is disabled by default for [production images][image-variants] and enabled by default for [development images][image-variants]." should be merged into this line?

garethtdavies: Sorry, I merged this at the exact time you left this comment. Duly noted and will follow up.

PR #1384, opened 2020-03-07
projects: add commento selfhosted reference

igolaizola:
Change-type: patch
Signed-off-by: igolaizola <11333576+igolaizola@users.noreply.github.com>


garethtdavies:
Thanks for your contribution, and sorry for the slow response. We have decided to restructure this page as it is outdated. Unfortunately, as this will now be sourced dynamically, the community section affected by this PR will no longer exist in the current form. As a result, I am closing this PR, but we will let you know if we reinstate this community section or similar so we can link to this.

PR #1389, opened 2020-03-11
integrations: Removing Azure integration

garethtdavies:
This integration is outdated and doesn't work. There is an internal discussion [here](https://www.flowdock.com/app/rulemotion/resin-propaganda/threads/wUn-Zd1-tHu1QN_2rSZ9t12NPSm) to replace it, but removing until this is available to avoid potential confusion.

Change-type: patch


garethtdavies:
@balena-ci retest

PR #1390, opened 2020-03-11
Getting started updates for all supported devices

### LINKED ISSUE #650: Add a getting started guide for QEMU

### LINKED ISSUE #988: Getting started docs needed for PocketBeagle
Getting started docs are needed for the Pocket Beagleboard.

Getting started URL as specified in coffee file is [https://docs.resin.io/pocketbeagle/nodejs/getting-started/#adding-your-first-device](https://docs.resin.io/pocketbeagle/nodejs/getting-started/#adding-your-first-device)

### LINKED ISSUE #1132: Getting Started guide is not consistent with UI instructions for Tinker S board
The UI correctly says:

* Use the form on the left to configure and download balenaOS for your new device.
* Write the OS file you downloaded to your SD card. We recommend using Etcher.
* Insert the freshly burnt SD card into the Asus Tinker Board S.
* Warning! This will also completely erase internal storage media, so please make a backup first.
* Before powering up the Asus Tinker Board S, make sure that the jumper between the 5V power supply and the HDMI connector is in the MASKROM mode, as ilustrated in this link https://tinkerboarding.co.uk/wiki/index.php/Setup
* The device is performing a shutdown. Please wait until all LEDs are off.
* Remove the balenaOS installation media.
* Before powering up the Asus Tinker Board S, make sure that the jumper between the 5V power supply and the HDMI connector is in the parking (no function) mode, as ilustrated in this link https://tinkerboarding.co.uk/wiki/index.php/Setup
* Remove and re-connect power to the board.
* Your device should appear in your application dashboard within a few minutes. Have fun!

But the getting started guide talks about flashing the eMMC directly with etcher, which does not appear to work.

### LINKED ISSUE #1321: Wrong images for Jetson Nano getting started guide
Some of the images show Raspberry Pi devices instead of the nano.

garethtdavies:
This PR is staged here: https://resin-docs-preview.herokuapp.com/docs/learn/getting-started

The content for existing devices should be the same other than updated screenshots for all devices. Other devices fall back to instructions in the devices API endpoint and can be updated with custom instructions as needed.

Example new board with fallback instructions: https://resin-docs-preview.herokuapp.com/docs/learn/getting-started/jetson-xavier/nodejs/

The screenshots are generated via Selenium implementation that can be viewed [here](https://github.com/balena-io-playground/getting-started-guide-scripts/blob/master/add_application.py). There are a few images that are too complex to automate (such as the ones downloading services), so in that case, I've made the images generic by blurring the device.

The device dictionary is [being generated](https://github.com/balena-io-playground/getting-started-guide-scripts/blob/master/device_config.py) from the v1 endpoint of the API. The Pi4 notably had a URL that wasn't the device slug, so that one has been updated with a redirect.

I've moved the generic content into the same file and only using imports for variable content. I don't have a strong opinion on this and can revert; it just made my life easier when editing this and makes it more obvious which content is intended to be variable. I've also done a purge of content that is no longer used (sample projects). 

As a result, it should be relatively easy to add a new device with the default instructions, and custom content can be appended.

Closes #1321
Closes #1132
Closes #988
Closes #650 

Change-type: minor


config/dictionaries/device.json
```diff
@@ -1,389 +1,1361 @@
...
+        "i": "Remove and re-connect power to the board."
+      }
+    ]
   },
   {
     "id": "intel-edison",
     "name": "Intel Edison",
     "arch": "i386",
-    "bootMedia": "eMMC",
+    "bootMedia": "SD card",
```
garethtdavies: Looks like this is falling back to SD card due to a lack of instructions. In this case it won't matter as has custom content but I'll resolve this specific board.

config/dictionaries/device.json
```diff
@@ -1,389 +1,1361 @@
...
+  {
+    "id": "surface-pro-6",
+    "name": "Microsoft Surface 6",
     "arch": "amd64",
-    "bootMedia": "USB drive",
-    "icon": "/img/device/intel-nuc.svg",
-    "link": "https://www.intel.com/content/www/us/en/products/boards-kits/nuc.html",
-    "type": "released"
+    "bootMedia": "USB key",
+    "state": "EXPERIMENTAL",
```
xginn8: i'm sure this is pulled from somewhere else, but i thought these were changed to `NEW`? just curious how it fits together

garethtdavies: So the API returns this and then it seems like it is down to each individual implementation to format it. The docs do this here: https://github.com/balena-io/docs/blob/master/templates/_devices.html#L22 but I've seen similar code in the dashboard (and I think SDK). Either way, I'll update this config file to use the updated ones, as though it isn't used anywhere currently and I left in to minimize changes I could see us using that in the future.

garethtdavies: Oh, new boards default to `NEW` so I think it's just old ones.

pages/learn/getting-started.md
```diff
@@ -25,42 +25,188 @@ If you don't already have a {{ $names.company.lower }} account, make sure to [si
...
+```
+
+Your application will then be downloaded and executed by all the devices you have connected in your application fleet. The first push is slower to deploy, but all subsequent pushes are much faster due to [Docker layer sharing][dockerLayerDocs]. You can see the progress of the device code updates on the device dashboard:
+
+<img alt="Service download progress" src="/img/common/device/download-progress.png" width="100%">
+
+<!-- This is language/project specific -->
 {{import "getting-started/postPush"}}
 
-{{import "getting-started/deviceUrl"}}
```
shaunmulligan: What was the reason for removing these types of partials? I think this partial could definitely be re-used in an FAQ and perhaps elsewhere where one just wants to reference how to enable a URL.

garethtdavies: > I've moved the generic content into the same file and only using imports for variable content. I don't have a strong opinion on this and can revert; it just made my life easier when editing this and makes it more obvious which content is intended to be variable.

Thanks and quoting myself :), this was more for my benefit in editing and no good rationale. Happy to add some back in where they could be reused like this one you suggest.

shaunmulligan: yeah it makes sense to only add things if they are useful in other places. When i first wrote this, the Getting started was the first use of the partials, so i tried to write most things for reuse in the docs and elsewhere, but agree it only makes sense if its actually useful and reusuable.

garethtdavies:
@balena-ci retest

PR #1400, opened 2020-03-24
Note that files in /run/secrets/ can only be copied using RUN cp

OdysLam:
flowdock discussion: https://www.flowdock.com/app/rulemotion/i-cli/threads/MfZnNVs4gmIRqJ4AVP0dVWNlhrn

Change-type: minor
Signed-off-by: Odysseas Lamtzidis odysseas@balena.io

### DIFF
--- a/shared/deployment/secrets/_default.md
+++ b/shared/deployment/secrets/_default.md
@@ -39,7 +39,17 @@ build-secrets:
           dest: my-recipe
 ```
 
-This will mount the `super-secret-recipe` file into `/run/secrets/my-recipe` file in every build container. Note that the `/run/secrets` folder is only available during the image build, and not present in the image that is deployed to the devices.
+This will mount the `super-secret-recipe` file into `/run/secrets/my-recipe` file in every build container. However, the `/run/secrets` folder will not be added to the build context that is sent to the Docker daemon (or balenaEngine),
+and therefore:
+
+* The `/run/secrets/` folder will not be present in the image that is deployed to the devices.
+* The `COPY` or `ADD` Dockerfile directives will **not** be able to copy files from that folder.
+* The folder will be available "for the duration of `RUN` directives", such that secrets can be
+  accessed by scripts/code at image build time, for example:
+
+` ``shell
+RUN cat /run/secrets/my-recipe | secret_processing_script.sh
+` ``
 
 To add a secret file to a specific service's build:
 
@@ -74,7 +84,7 @@ build-secrets:
 
 This will mount the `super-secret-recipe` file as `/run/secrets/my-recipe` for all services, and `super-secret-recipe-2` as `/run/secrets/my-recipe2` for `service1` only. Again, note that the `/run/secrets` folder is only available during the image build, and not present in the image that is deployed to the devices.
 
-Subdirectories are supported in both the source (`.balena/secrets`) and the destination (`/run/secrets`)
+Subdirectories are supported in both the source (`.balena/secrets`) and the destination (`/run/secrets`).
 
 ### Build variables
 


shared/deployment/secrets/_default.md
```diff
@@ -74,7 +74,9 @@ build-secrets:
 
 This will mount the `super-secret-recipe` file as `/run/secrets/my-recipe` for all services, and `super-secret-recipe-2` as `/run/secrets/my-recipe2` for `service1` only. Again, note that the `/run/secrets` folder is only available during the image build, and not present in the image that is deployed to the devices.
 
-Subdirectories are supported in both the source (`.balena/secrets`) and the destination (`/run/secrets`)
+Subdirectories are supported in both the source (`.balena/secrets`) and the destination (`/run/secrets`).
+
+It is worth mentioning that if you want to copy the file `super-secret-recipe` from the `/run/secrets/my-recipy` folder to the container, you can't use the Dockerfile primitive `COPY`, but rather you have to use `RUN cp /run/secrets/my-recipy <destination>`.
```
xginn8:
```suggestion
It is worth mentioning that if you want to copy the file `super-secret-recipe` from the `/run/secrets/my-recipe` folder to the container you must use the RUN directive, e.g. `RUN cp /run/secrets/my-recipe <destination>`.
```

garethtdavies:
Thanks for the PR 👍 While I agree clarity is needed here, I find this specific example a little counterintuitive (since this section predominantly deals with _not_ copying this stuff to the container). 

I wonder if we can generalize this more in line with the [masterclass](https://www.balena.io/docs/learn/more/masterclasses/cli-masterclass/#8-building-with-secrets-and-variables) and outline what can be done via `RUN` e.g. from masterclass "this file could obviously contain a raft of different functions, including a script that gets executed, text for filling in files, etc."

shaunmulligan:
@OdysLam I agree with @garethtdavies on this, I don't think we should be encouraging people to be copying these secrets into the images, that negates their design purpose which is to be build time secrets, that expressly don't end up in the built image. They should be used to pull in private git repos or install private NPM packages, etc. If runtime secrets are needed, the currently proposed way of doing them are base64 encoded ENV vars

OdysLam:
@garethtdavies @shaunmulligan  .The  issue is that bug/error provides no feedback whatsoever, it just tells that it could not perform COPY because it can't find the file.

I think we should at least mention that while underlining that this is not the intended functionality and provide an alternative workflow for copying files (e.g key.pem) into the device for use by the services.

I literally spent 20-30min trying to understand why the Dockerfile did not work and trying different random things.

garethtdavies:
@OdysLam I still think it makes sense to give more details to describe how you must use `RUN` to work with these build time secrets/variables and also linking to the masterclass for a walkthrough.

I don't really understand the alternative workflow and might be missing something, in that a normal `COPY` would work if it were not included as build-time secrets (which are specifically designed to not propagate to the device)? In this instance wouldn't you just `.gitignore` the file to keep out of any repositories and then push to the device as normal?

OdysLam:
The idea was to put everything in .balena and just ignore that folder so to make sure that I am not forgeting anytihng. 

I am not saying that we should encourage that workflow, but I think we should explicitly say that something won't work, especially if the error that it produces can not be traced back to the cause.  (I had @pdcastro tell me that won't work as I couldn't figure out why it wasn't working).

In any case, how do you think that we should handle it? Happy to make any PR.

shared/deployment/secrets/_default.md
```diff
@@ -74,7 +74,9 @@ build-secrets:
 
 This will mount the `super-secret-recipe` file as `/run/secrets/my-recipe` for all services, and `super-secret-recipe-2` as `/run/secrets/my-recipe2` for `service1` only. Again, note that the `/run/secrets` folder is only available during the image build, and not present in the image that is deployed to the devices.
 
-Subdirectories are supported in both the source (`.balena/secrets`) and the destination (`/run/secrets`)
+Subdirectories are supported in both the source (`.balena/secrets`) and the destination (`/run/secrets`).
+
+It is worth mentioning that if you want to copy the file `super-secret-recipe` from the `/run/secrets/my-recipe` folder to the container you must use the RUN directive, e.g. `RUN cp /run/secrets/my-recipe <destination>`.
```
pdcastro: How about this: there was already a paragraph earlier in this document that touched the topic:
```
This will mount the `super-secret-recipe` file into `/run/secrets/my-recipe` file in
every build container. Note that the `/run/secrets` folder is only available during
the image build, and not present in the image that is deployed to the devices.
```

I suggest amending that paragraph instead of adding a new paragraph here. Amending it like:

```
This will mount the `super-secret-recipe` file into `/run/secrets/my-recipe` file in
every build container. However, the `/run/secrets` folder will not be added to the
"build context" (tar stream) that is sent to the Docker daemon (or balenaEngine),
and therefore:

* The `/run/secrets/` folder will not be present in the image that is deployed to the devices.
* The `COPY` or `ADD` Dockerfile directives will **not** be able to copy files from that folder.
* The folder will be available "for the duration of `RUN` directives", such that secrets can be
  accessed for example by scripts/code at image build time:

` ``sh
RUN cat /run/secrets/my-recipe | secret_processing_script.sh
` ``
```
The `RUN cat` example above is probably less controversial than `RUN cp`. :-)

garethtdavies: @OdysLam the above clarification from Paulo looks great. Do you agree and want to update this PR to include?

OdysLam: Totally onboard @garethtdavies. Thank you @pdcastro for putting the time for these changes. 

Will update the PR today.

shared/deployment/secrets/_default.md
```diff
@@ -72,12 +72,22 @@ build-secrets:
...
+every build container. However, the `/run/secrets` folder will not be added to the
+"build context" (tar stream) that is sent to the Docker daemon (or balenaEngine),
+and therefore:
+
+* The `/run/secrets/` folder will not be present in the image that is deployed to the devices.
+* The `COPY` or `ADD` Dockerfile directives will **not** be able to copy files from that folder.
+* The folder will be available "for the duration of `RUN` directives", such that secrets can be
+  accessed by scripts/code at image build time, for example:
+
+```sh
```
garethtdavies:
```suggestion
```shell
```

I know this is an alias but for consistency let's update to shell as is used elsewhere.

shared/deployment/secrets/_default.md
```diff
@@ -72,12 +72,22 @@ build-secrets:
                dest: my-recipe2
 ```
 
-This will mount the `super-secret-recipe` file as `/run/secrets/my-recipe` for all services, and `super-secret-recipe-2` as `/run/secrets/my-recipe2` for `service1` only. Again, note that the `/run/secrets` folder is only available during the image build, and not present in the image that is deployed to the devices.
+This will mount the `super-secret-recipe` file into `/run/secrets/my-recipe` file in
+every build container. However, the `/run/secrets` folder will not be added to the
+"build context" (tar stream) that is sent to the Docker daemon (or balenaEngine),
```
garethtdavies:
```suggestion
build context that is sent to the Docker daemon (or balenaEngine),
```

I'd remove quotes as build context is consistent with terminology we have used elsewhere.

pdcastro:
> Can you squash these commits before this is merged in?

If helpful (sorry otherwise), :-) start with rebasing your local branch with something like:

```sh
git checkout OdysLam-patch-1
git fetch -t origin master:master
git rebase master
```

And then proceed with the squashing / fixup via interactive rebasing:
* `git rebase -i master`
* http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html
* https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together
* https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

When all looks good locally (`git logs`), force push with
```
git push -f origin OdysLam-patch-1
```
or [supposedly safer](https://stackoverflow.com/questions/52823692/git-push-force-with-lease-vs-force):
```
git push --force-with-lease origin OdysLam-patch-1
```
(I don't have as much experience as I should with `--force-with-lease`, it is something that I am working on.) :-)

shared/deployment/secrets/_default.md
```diff
@@ -72,9 +72,21 @@ build-secrets:
                dest: my-recipe2
 ```
 
-This will mount the `super-secret-recipe` file as `/run/secrets/my-recipe` for all services, and `super-secret-recipe-2` as `/run/secrets/my-recipe2` for `service1` only. Again, note that the `/run/secrets` folder is only available during the image build, and not present in the image that is deployed to the devices.
+This will mount the `super-secret-recipe` file into `/run/secrets/my-recipe` file in
+every build container. However, the `/run/secrets` folder will not be added to the
+build context that is sent to the Docker daemon (or balenaEngine),
+and therefore:
```
pdcastro, line 78: Oh wait -- wasn't this supposed to have replace the paragraph in line 42? (Rather than replacing the paragraph in line 75...)

garethtdavies:
@balena-ci rebase

garethtdavies:
@OdysLam I'm merging this now to save any future rebasing. In the future feel free to merge yourself (in fact this is the encouraged behavior) once you are happy and it has an approval. Thanks again!

PR #1407, opened 2020-03-31
deltas: Delta updates enabled for default for >2.47.1

garethtdavies:
Internal discussion: https://www.flowdock.com/app/rulemotion/i-docs/threads/FIhR3qd_HUhcOzLv3VIx2zA5nn1

Change-type: patch


pages/learn/deploy/delta.md
```diff
@@ -8,30 +8,34 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
+To address some of these issues, we have implemented a "binary delta" update process. Instead of initiating a Docker pull when the device is notified about an update, it requests the {{ $names.company.lower }} servers to provide just the differences between the old and new container image. This comparison is done on the full image level, comparing the actual content, regardless of the layers used, resulting in the minimum amount of change required to get from the previous application version to the new one. In the worst case (i.e. completely replaced application image) the binary delta is equal size to the Docker pull (the device needs to download the full image). In most cases, the binary delta will be much smaller.
```
garethtdavies, line 11: To save you searching for the change it's just a missing "in" 

`resulting **in** the minimum amount...`

garethtdavies:
@dfunckt is correct that updating the host OS to >2.47.1 enables delta updates or would it depend on the application/device settings?

pages/learn/deploy/delta.md
```diff
@@ -8,30 +8,34 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
 
 These binary deltas save on the amount of data needed to be downloaded, reduce the storage space requirements on the device to perform an application update, and shorten the time when Docker is updating.
 
-## Enabling delta updates
+## Enabling/Disabling delta updates
 
-Delta update behavior is enabled with the `RESIN_SUPERVISOR_DELTA` configuration variable:
+__Note__: Delta updates are enabled by default for devices running {{ $names.os.lower }} >= 2.47.1.
+
+For any devices running {{ $names.os.lower }} >= 2.47.1, the delta update behavior is enabled by default. For devices running {{ $names.os.lower }} < 2.47.1, updating to >= 2.47.1 via a [self-service update][self-service-update] will enable delta updates for the device. Alternitavely, the delta update behavior may be enabled or disabled application-wide or per-device with the `RESIN_SUPERVISOR_DELTA` configuration variable.
```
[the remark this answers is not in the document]
garethtdavies: It's `RESIN_` in the dashboard so probably best to leave like that for consistency.

CleanShot 2020-03-31 at 16 02 38@2x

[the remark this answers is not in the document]
dfunckt: Yes, deltas are enabled by default for ESR >=2020.04 as well, just like for balenaOS >=2.47.1. Good catch!

pages/learn/deploy/delta.md
```diff
@@ -8,30 +8,34 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
-
-In addition, the wait time before a connection is considered stalled and the time between retries are configurable, using `RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT` and `RESIN_SUPERVISOR_DELTA_RETRY_INTERVAL`. These are both specified in milliseconds. The defaults for these options allow for a 20 minute time frame where no byte is received before giving up, after which the supervisor retries the delta from the beginning.
+Delta updates are resumable, so if the connection drops or otherwise stalls, the update will resume from the last byte received. Updates will be retried 30 times, with an interval of 1 second between retries. In addition, there is a wait time of 30 seconds during updates before a connection is considered stalled. This allows for a 20-minute timeframe where no byte is received before the update fails, after which the supervisor retries the delta from the beginning.
```
dfunckt, line 37: Seems this paragraph cannot be easily converted to something that is 100% accurate for both old rsync and new balenaEngine deltas (eg. I'm not sure if balenaEngine resumes from last byte, and it definitely doesn't have that exact retry behaviour). If we just removed it, what would be left (the paragraph above) would not be very useful so I think this whole *section* can be removed for now and replaced by something that better describes the whole delta update flow from its beginning.

Some key points:

1. Before devices can update using deltas, a delta must be generated between the release a device currently runs and the one it is updating to. There are two ways deltas are generated: (1) automatically between the last successful and the new release during a build on the balenaCloud builder -- these are known as "build-time" deltas. (2) automatically when a device is requesting to update from or to a release for which no delta has been generated before, these are known as "on-demand" deltas and can typically happen if the device was not on the latest release when the new release was pushed.

2. On-demand deltas may take a while to generate, depending primarily on the size of the images involved, and while doing so, the device keeps polling balenaCloud for its readiness (this is the "delta is still being generated remotely; will retry" -- or something to that effect -- log message). If a delta exists however, this step is skipped and the flow continues from the next step.

3. The device is downloading and applying the delta. The old paragraph about the "Download progress" can be reused here.

4. When the update finishes, the rest of the flow is identical to an update without deltas (ie. "regular" update). Previous services are stopped and the new services are started, etc.

Let's sync up on Flowdock if needed and discuss how we can flesh this out to be proper documentation.

garethtdavies: Thanks @dfunckt I'm testing this so I can understand better before rewriting this.

One thing I am currently confused by is what we currently support for sync vs balenaEngine deltas (which is v2 vs v3 right?) I understand anything >=2.47.1 will be the new balenaEngine deltas and on by default but what about older devices that enable this functionality? Is there a clear distinction for when each version is used or are they all the latest balenaEngine deltas now?

dfunckt: @garethtdavies the different versions are an implementation detail users shouldn't care about and I don't think docs should discuss this (@shaunmulligan?). From the user's point of view the only distinction between versions is that we don't generate v2 deltas on the builder during build.

For your reference:

- every device above or on 2.47.1 will get deltas v3 enabled by default
- the devices of every application created since December will get deltas v3 if the user enables deltas on their own
- all other devices will get deltas v2 if the user enables deltas on their own

garethtdavies: Thanks for that explanation @dfunckt. Yes, I'm not suggesting we document the underlying technical differences between delta versions (though undoubtedly that's a good blog post). I'm trying to elucidate the differences that the user will experience when interacting with the different delta versions. 

As an example, seeing the delta size in build output is incredibly descriptive and illustrates what they do perfectly so I'd like to highlight this (e.g. a screenshot). However, I now understand that these are only visible when using the newer versions of deltas as those are the ones generated at build time so I can add a note to that effect.

I'll conduct some more testing and I'll ping you (in FD) for some more questions before finishing this draft :)

garethtdavies:
@xginn8 Is this a good place for delta HUP docs? Could include as a partial or just make note and link to host upgrade docs https://www.balena.io/docs/reference/OS/updates/update-process/

xginn8:
@balena-ci rebase

pages/learn/deploy/delta.md
```diff
@@ -8,30 +8,40 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
+
 ## Delta behavior
 
-If you are using delta updates, you might notice the following changes in {{ $names.company.lower }} behavior:
+Before devices can update using deltas, a delta must be generated between the release a device is currently running and the one it is updating to. There are two ways deltas are generated:
+
+1. Automatically between the last successful and the new release during a build on the {{$names.cloud.lower}} builder - these are known as "build-time" deltas.
+2. Automatically when a device is requesting to update from or to a release for which no delta has been generated before - these are known as "on-demand" deltas and can typically happen if the device was not on the latest release when the new release was pushed, or the user did not use the {{$names.cloud.lower}} builder for the new release.
+
+__Note:__ Delta updates that were enabled for applications created prior to December 2019 (v2) do not generate build-time deltas, and will always generate an on-demand delta.
```
dfunckt, line 38: This is not telling the full story and also feels redundant, especially having to mention a specific date.

pages/learn/deploy/delta.md
```diff
@@ -8,30 +8,40 @@ thumbnail: /img/runtime/ResinSupervisorDelta.png
...
+The *Download progress* bar on the dashboard might show for only a short time, much shorter than in a normal application update. In the most common development patterns, there are usually very small changes between one version of the application image and the next (e.g. fixing typos, adding a new source file, or installing an extra OS package), so when using deltas these changes are downloaded much quicker than before.
 
-In addition, the wait time before a connection is considered stalled and the time between retries are configurable, using `RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT` and `RESIN_SUPERVISOR_DELTA_RETRY_INTERVAL`. These are both specified in milliseconds. The defaults for these options allow for a 20 minute time frame where no byte is received before giving up, after which the supervisor retries the delta from the beginning.
+When the update finishes, the previous services are stopped and the new services are started and the prior image removed.
```
dfunckt, line 44: That is accurate for the default case, but depends on the specific update strategy used. I think it's worth mentioning that and also linking the relevant docs here.

xginn8:
@balena-ci I self-certify!

PR #1417, opened 2020-04-21
Adding note about accessing the VPN behind country level firewalls

### LINKED ISSUE #1415: Explain the potential network connections issues of devices from China
As a fleet owner with devices in China, I'd like to have more information on which features of balenaCloud would have issues due to the great firewall of China.

Adding a bit more detail here might be the best place https://www.balena.io/docs/reference/OS/network/2.x/#network-requirements

We heard that VPN connection from China kinda working but we also observed that some users have issues because of China's firewall rules. This is a known problem and there are several users reporting this: https://github.com/balena-io/balena/issues/466

garethtdavies:
Closes #1415 

Added some general text about accessing the VPN behind country-level firewalls and also a specific FAQ with reference to China. What else should be in here cc @alisondavis17 

Change-type: patch

### DIFF
--- a/pages/faq/troubleshooting/faq.md
+++ b/pages/faq/troubleshooting/faq.md
@@ -9,6 +9,7 @@ title: FAQs
 * [Why does `/data` report weird usage?](#why-does-data-report-weird-usage)
 * [What NTP servers do the devices use?](#what-ntp-servers-do-the-devices-use)
 * [What Network Ports are required?](#what-network-ports-are-required)
+* [Can I use {{ $names.cloud.lower }} in countries with restrictive firewalls such as China?](#can-i-use-balenacloud-in-countries-with-restrictive-firewalls-such-as-china)
 * [Can I access /dev and things like GPIO from the container?](#can-i-access-dev-and-things-like-gpio-from-the-container)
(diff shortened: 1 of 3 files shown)


pages/faq/troubleshooting/faq.md
```diff
@@ -78,6 +77,10 @@ Additionally, you should whitelist the following domains for the relevant ports
 * `*.docker.com`
 * `*.docker.io`
 
+##### Can I use {{ $names.cloud.lower }} in China?
```
xginn8, line 80:
```suggestion
##### Can I use {{ $names.cloud.lower }} in countries with restrictive national firewalls?
```

xginn8: might be better to avoid naming names, just a thought!

garethtdavies: That was my thought too, which is why I only included that bit in a FAQ section, where it probably is appropriate but I'll let someone else decide :)

alisondavis17: I suspect some people will search for China specifically; that's the country we tend to get asked about. Maybe "countries with restrictive firewalls such as China?"

garethtdavies:
@balena-ci retest

garethtdavies:
@balena-ci rebase

PR #1420, opened 2020-04-23
Adding details for my_application resource

### LINKED ISSUE #1418: Update application API endpoint examples to show commands only for user applications
Since the introduction of public apps, the `https://api.balena-cloud.com/v5/application` now returns 100 other public applications in addition to the user's apps.

You can limit this by passing in a filter `https://api.balena-cloud.com/v5/application?$filter=is_public eq false` or there is the `https://api.balena-cloud.com/v5/my_application` endpoint. It's likely that the latter is what most users would expect to be returned by default.

cc @szony who raised this

garethtdavies:
Closes #1418 

Added some extra examples for `my_application` resource and clarification that `application` will return public applications. On the resources page, I added this under the `application` resource as didn't want to duplicate it all as a separate `my_application` resource but I can do this if further clarification is required.

Change-type: patch

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -25,6 +25,13 @@
         "filters": ""
       },
       {
+        "id": "get-my-apps",
+        "summary": "Get all applications associated with authenticated user",
+        "method": "GET",
+        "endpoint": "/v5/my_application",
+        "filters": ""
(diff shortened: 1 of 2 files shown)


szony:
Looks good, nice work @garethtdavies

PR #1421, opened 2020-04-23
Pull in SDK deprecation policy

garethtdavies:
Depends-on: https://github.com/balena-io/balena-sdk/pull/892

This will pull in the SDK deprecation policy as a partial. Requires balena-io/balena-sdk/pull/892 to be merged as pulling from master. Tested the above branch and looks like this:

CleanShot 2020-04-23 at 12 53 59@2x

Change-type: patch


thgreasi:
HQ: https://github.com/balena-io/balena/issues/2032

PR #1423, opened 2020-04-24
Adding user management API examples

### LINKED ISSUE #1412: Document user management via API
It is possible to add/remove users via our API, however it's not documented. There are two end points where we seem to have users but there are no examples shown:
https://www.balena.io/docs/reference/api/resources/application/
https://www.balena.io/docs/reference/api/resources/user/

garethtdavies:
Closes #1412 

These could likely do with some further explanation, e.g. an API masterclass but added examples for now and also added the application membership role resource.

Change-type: patch

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -101,6 +101,23 @@
     ]
   },
   {
+    "id": "application_membership_role",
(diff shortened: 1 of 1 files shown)


garethtdavies:
@balena-ci rebase

PR #1463, opened 2020-06-18
fix up small typo in time mgmt doc

xginn8:
Change-type: patch


xginn8:
@balena-ci I self-certify!

PR #1481, opened 2020-08-12
Update device resource API fields

pipex:
Added 
- mac_address
- api_heartbeat_state

Removed
- owns__device_log


pipex:
Setting as draft while I make sure modifying the resources.json is the only thing I need

shaunmulligan (approved):
looks good @pipex

PR #1492, opened 2020-09-14
update-strategies: fix up number + supervisor reference

xginn8:
Change-type: patch


xginn8:
@balena-ci retest

xginn8:
@balena-ci I self-certify!

Issue #1485, opened 2020-09-03
Document API resource properties


vipulgupta2048:
Docs: https://www.balena.io/docs/reference/api/

For example: what does `should_be_running_release` on device mean? 
This is the kind of information that the API documentation should include to avoid confusion and support tickets. We should document each API request with properties, examples, or if there are any caveats or edge cases customers should know about.  

FD thread: https://www.flowdock.com/app/rulemotion/i-docs/threads/3uohhegEJXWp88JFasKDv-Bld-A
@dt-rush @alisondavis17 :eyes:

vipulgupta2048:
Will prepare a draft and share

vipulgupta2048:
Closing #1485 in favor of #1496

PR #1496, opened 2020-09-22
Document API properties 

vipulgupta2048:
Solves #1485 
Some previews for you all, 
Screenshot from 2020-09-22 15-43-14
Screenshot from 2020-09-22 15-43-21

@dt-rush @pdcastro @alisondavis17 :eyes:

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -20,6 +20,7 @@
       {
         "id": "get-all-apps",
         "summary": "Get all applications",
+        "description": "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.",
         "method": "GET",
         "endpoint": "/v5/application",
         "filters": ""
@@ -27,6 +28,7 @@
       {
         "id": "get-my-apps",
         "summary": "Get all applications associated with authenticated user",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/my_application",
         "filters": ""
@@ -34,6 +36,7 @@
       {
         "id": "app-by-id",
         "summary": "Get application by ID",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/application(<ID>)",
         "filters": ""
@@ -41,6 +44,7 @@
       {
         "id": "app-by-name",
         "summary": "Get application by name",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/application",
         "filters": "?\\$filter=app_name%20eq%20'<NAME>'"
@@ -48,6 +52,7 @@
       {
         "id": "app-with-devices",
         "summary": "Get application by id along with its devices",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/application(<ID>)",
         "filters": "?\\$expand=owns__device"
@@ -55,6 +60,7 @@
       {
         "id": "app-filter-public",
         "summary": "Filter public applications from returned applications",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/application",
         "filters": "?\\$filter=is_public%20eq%20false"
@@ -62,6 +68,7 @@
       {
         "id": "create-app",
         "summary": "Create application",
+        "description": "",
         "method": "POST",
         "endpoint": "/v5/application",
         "filters": "",
@@ -70,6 +77,7 @@
       {
         "id": "delete-app",
         "summary": "Delete application",
+        "description": "",
         "method": "DELETE",
         "endpoint": "/v5/application(<ID>)",
         "filters": ""
@@ -77,6 +85,7 @@
       {
         "id": "disable-auto-update",
         "summary": "Disable automatic update tracking for an application",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/application(<ID>)",
         "filters": "",
@@ -85,6 +94,7 @@
       {
         "id": "enable-auto-update",
         "summary": "Enable automatic update tracking for an application",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/application(<ID>)",
         "filters": "",
@@ -93,6 +103,7 @@
       {
         "id": "set-commit",
         "summary": "Set entire fleet to a specific commit",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/application(<ID>)",
         "filters": "",
@@ -111,6 +122,7 @@
       {
         "id": "get-application-membership_role",
         "summary": "Get all application membership roles",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/application_membership_role",
         "filters": ""
@@ -130,6 +142,7 @@
       {
         "id": "get-application-tags",
         "summary": "Get all tags by application name",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/application_tag",
         "filters": "?\\$filter=application/app_name%20eq%20'<NAME>'"
@@ -137,6 +150,7 @@
       {
         "id": "create-application-tag",
         "summary": "Create a new application tag",
+        "description": "",
         "method": "POST",
         "endpoint": "/v5/application_tag",
         "filters": "",
@@ -145,6 +159,7 @@
       {
         "id": "update-application-tag",
         "summary": "Update an application tag",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/application_tag/(<ID>)",
         "filters": "",
@@ -153,6 +168,7 @@
       {
         "id": "delete-application-tag",
         "summary": "Delete an application tag",
+        "description": "",
         "method": "DELETE",
         "endpoint": "/v5/application_tag/(<ID>)",
         "filters": ""
@@ -173,6 +189,7 @@
       {
         "id": "get-api-keys",
         "summary": "Get all API keys",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/api_key",
         "filters": ""
@@ -180,6 +197,7 @@
       {
         "id": "create-api-key",
         "summary": "Create new API key",
+        "description": "",
         "method": "POST",
         "endpoint": "/api-key/user/full",
         "filters": "",
@@ -237,6 +255,7 @@
       {
         "id": "get-devices",
         "summary": "Get all devices",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/device",
         "filters": ""
@@ -244,6 +263,7 @@
       {
         "id": "devices-by-application",
         "summary": "Get all devices by application",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/device",
         "filters": "?\\$filter=belongs_to__application%20eq%20'<APP_ID>'"
@@ -251,6 +271,7 @@
       {
         "id": "device-by-id",
         "summary": "Get device by ID",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/device(<ID>)",
         "filters": ""
@@ -258,6 +279,7 @@
       {
         "id": "device-by-uuid",
         "summary": "Get device by UUID",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/device",
         "filters": "?\\$filter=uuid%20eq%20'<UUID>'"
@@ -265,6 +287,7 @@
       {
         "id": "rename-device",
         "summary": "Rename device",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
@@ -273,6 +296,7 @@
       {
         "id": "delete-device",
         "summary": "Delete device",
+        "description": "",
         "method": "DELETE",
         "endpoint": "/v5/device(<ID>)",
         "filters": ""
@@ -280,6 +304,7 @@
       {
         "id": "add-note",
         "summary": "Add note to a device",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
@@ -288,6 +313,7 @@
       {
         "id": "move-device",
         "summary": "Move device to another application",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
@@ -296,6 +322,7 @@
       {
         "id": "set-device-to-release",
         "summary": "Set device to a specific release",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
@@ -304,6 +331,7 @@
       {
         "id": "deactivate-device",
         "summary": "Deactivate an offline device (CHARGEABLE)",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
@@ -324,6 +352,7 @@
     {
         "id": "get-device-config-vars",
         "summary": "Get all device config variables for a device",
+        "description": "",
         "method": "GET",
         "endpoint": "/v5/device_config_variable",
         "filters": "?\\$filter=device%20eq%20<DEVICE ID>"
@@ -331,6 +360,7 @@
       {
         "id": "create-device-config-var",
         "summary": "Create a device config variable",
+        "description": "",
         "method": "POST",
         "endpoint": "/v5/device_config_variable",
         "filters": "",
@@ -339,6 +369,7 @@
       {
         "id": "update-device-config-var",
         "summary": "Update a device config variable",
+        "description": "",
         "method": "PATCH",
         "endpoint": "/v5/device_config_variable(<ID>)",
         "filters": "",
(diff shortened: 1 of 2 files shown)


vipulgupta2048:
If this looks good, please let me know and I will add the rest. Thanks!

pdcastro (commented):
It looks good, 👍   _provided that_ the non-empty descriptions are actually helpful. :-)  

> while leaving the ones that are absolutely simple. One-liners should work as well for them.

I would say that even one-liners would be too much (distracting noise) if they don't actually add something new and useful. If it was just to repeat the summary line, then better leave the description empty.

Did you mean to merge this PR with empty description strings? I suggest replacing the Lorem Ipsum placeholder with at least one actual description, then we could get the PR merged.

templates/api.html
```diff
@@ -10,7 +10,8 @@ <h2>Available fields</h2>
 		<div style="float: right; width: 60%; padding-left: 10px;">
 			<h2>Examples</h2>
 			{% for example in $resource.examples %}
-				<h3 id="{{ example.id }}">{{ example.summary}}</h3>
+				<h3 id="{{ example.id }}">{{ example.summary }}</h3>
+				<p>{{ example.description }}</p>
```
pdcastro, line 14: I didn't test, just a thought: In case many of the descriptions were to be initially left empty, we wouldn't want this `<p>` tag to introduce a significant vertical blank space above the example box. Just wondering whether the `<div>` tag might introduce less vertical blank space. Something to test - it may not make any difference, and the issue may be imaginary, in my head only. :-)

vipulgupta2048: @pdcastro Yes, initially my thoughts were to document all the necessary descriptions myself but I think its too much work for myself to take on + a heavier PR will become harder to review. Instead, I will just drop the mechanism in place and get the PR ready with some I think we really need. 

Regarding your theory, I agree a <p> tag might leave some empty space if not dead HTML code in our website which ain't desirable. I will see what I can do about that.

vipulgupta2048: Good news, resolved it with a if statement that checks if there is a description in the first place if no `<p>` tag created.

vipulgupta2048:
@pdcastro Watch out for deletions and let me know. I checked all the resources, currently, we don't have anything that needs documentation. We should provide context on several fields though.

config/dictionaries/resource.json
```diff
@@ -288,14 +313,16 @@
...
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
         "data": "{\n    \"belongs_to__application\": \"<APP ID>\"\n}"
       },
       {
         "id": "set-device-to-release",
-        "summary": "Set device to a specific release",
+        "summary": "Pin device to a specific release",
+        "description": "Making a GET request to fetch current target release hash of the device will output as null unless the device is pinned to a release.",
```
pdcastro, line 325: The summary and the example are for the PATCH method (setting `should_be_running__release` to a given release ID), but the proposed description regards the GET method and the fact that GET will return a `null` value for `should_be_running__release` if this field has not been (explicitly) set. Almost like a description for a different example / different method. :-) But I also see how a user might think that `should_be_running__release` would always have a non-null value (provided that the application has at least one release).

Instead of this description for the PATCH method, how about adding an example for the GET method, immediately before or after the example for the PATCH method? Then the description for the GET could more justifiably refer to the (un)expected results of the GET method. For example:

```
      {
        "id": "get-target-release",
        "summary": "Get the release that a device is pinned to",
        "description": "Note that `should_be_running__release` will be unset (`null`) 
                        if the device has not been explicitly pinned to a release.
                        To find out the target release for an unpinned device, check
                        the `commit` field of the `application` resource instead.",
        "method": "GET",
        "endpoint": "/v5/device(<ID>)",
        "filters": "?\\$select=should_be_running__release",
        "data": ""
      }
```
I believe that _"check the `commit` field of the `application` resource instead"_ is correct, but I am not 100% sure. I have successfully run these `curl` command lines:
```
$ curl "https://api.balena-cloud.com/v5/device(3991606)?\$select=should_be_running__release" \
  -H "Authorization: Bearer $authToken" | jq .
{
  "d": [
    {
      "should_be_running__release": null,
      "__metadata": {}
    }
  ]
}

$ curl "https://api.balena-cloud.com/v5/application(1301645)?\$select=commit" \
  -H "Authorization: Bearer $authToken" | jq .
{
  "d": [
    {
      "commit": "5dadd8fc163e9f5c98f2f417a517193c",
      "__metadata": {}
    }
  ]
}
```

templates/api.html
```diff
@@ -10,8 +10,11 @@ <h2>Available fields</h2>
 		<div style="float: right; width: 60%; padding-left: 10px;">
 			<h2>Examples</h2>
 			{% for example in $resource.examples %}
-				<h3 id="{{ example.id }}">{{ example.summary}}</h3>
-				<pre><code>curl -X {{ example.method }} \
+				<h3 id="{{ example.id }}">{{ example.summary }}</h3>
+        {% if example.description %}
+          <p>{{ example.description }}</p>
+        {% endif %}
+        <pre><code>curl -X {{ example.method }} \
```
pdcastro, line 17: The indentation got wrong - mixed spaces and tabs. Looks like the original file used tabs, and the PR added spaces.

pdcastro:
> I checked all the resources, currently, we don't have anything that needs documentation. We should provide context on several fields though.

_"Providing context on several fields"_ sounds like documenting, for example, the `should_be_running__release` field, which is `null` unless explicitly set, as discussed in other PR comments for device query examples. Relying on examples is not the same as documenting resource fields though. We would not need an example of a `curl` command line for the GET method if we could directly document the `should_be_running__release` field. I don't have a suggestion of how to achieve this, nor am I saying that this PR should solve it, I am just pointing it out. :-)

templates/api.html
vipulgupta2048: Thanks for the catch looks like I have to adjust my editor settings to convert space to tabs by default

config/dictionaries/resource.json
vipulgupta2048: Good one, Paulo. Having another example for GET method of the field makes more sense. A caveat I experienced was that we can't format any information inside the description. I tried adding <pre></pre> tags to get preformatted text formating on several terms but it seems like the StaticJinja template directly renders strings out of JSON.

I will add your example, though I recommend not using `commit` field since that's for the entire application, we already have `is_on__commit` for that. So will use that instead (On more thought, I will add an example for that too!)

vipulgupta2048: Screenshot from 2020-10-12 11-45-39
This is what I meant by no formatting

vipulgupta2048: Let me know if I should create an issue for this or just leave it with these `  quotes

pdcastro: > _I recommend not using `commit` field since that's for the entire application, we already have `is_on__commit` for a particular device. So I will use that instead, let me know your thoughts_

Interesting. I have more questions than answers, which is why we need more documentation. :-) Considering the distinction between "current state" and "target state", `is_on__commit` sounds like current state (the commit that the device **is** currently running), whereas `should_be_running__release` sounds like "target state" (the commit† that a device **should** be running). Is it true / correct that `is_on__commit` regards current state, whereas `should_be_running__release` regards target state? For a non-pinned device, I thought that `application.commit` would be the target state for a device that belongs to that application, hence my previous suggestion: _"To find out the target release for an unpinned device, check the `commit` field of the `application` resource instead."_

† Btw, there is also the question of release ID vs commit hash. `should_be_running__release` contains a release ID:
```
$ curl -H "Authorization: Bearer $authToken" -H "Content-Type: text/plain" "https://api.balena-cloud.com/v5/device(3991606)?\$select=should_be_running__release" | jq .
{
  "d": [
    {
      "should_be_running__release": {
        "__id": 1552722,
        "__deferred": {
          "uri": "/resin/release(@id)?@id=1552722"
        }
      },
      "__metadata": {}
    }
  ]
}
```
... whereas `device.is_on__commit` and `application.commit` are commit hashes. But we can probably let users figure this out for themselves... 🤔

vipulgupta2048: Yes, you are correct in your assumption @pdcastro, but to not document `should_be_running__release` means we assume the customer knows an intricate bit of how balena works by pinning releases and targeting it and assume that they know why `should_be_running__release` is coming out to be null (because they didn't give a target in the first place) Which brings us all the way back to how the issue came to be #1496 

> For a non-pinned device, I thought that application.commit would be the target state for a device that belongs to that application, hence my previous suggestion: "To find out the target release for an unpinned device, check the commit field of the application resource instead."

You are also right about this, but I still like to go for `is_on__commit` if I want to know what commit the device is running on **right now** 

Good catch on `release ID vs commit hash` and yes, we can let them figure it out this one with minor improvement in the docs to mention it for them. Also, where can I find a device ID? I didn't know they even existed until I made the API call.

vipulgupta2048: What about formatting, Paulo? Leave it with single quotes or no single quotes?

vipulgupta2048 (commented):
@pdcastro Making a new thread, 
Is there a reason why we have Release UUID, Release ID (Similarly DeviceID, Device UUID, application ID)? I didn't know we even had device ID tbh till I made the API call yesterday.

config/dictionaries/resource.json
```diff
@@ -288,22 +313,43 @@
...
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
         "data": "{\n    \"should_be_running__release\": \"<RELEASE ID>\"\n}"
+	  },
+	  {
+        "id": "get-current-release",
+        "summary": "Get current release running on device",
+        "description": "The `is_on__commit` field fetches the hash of the current release that the device is running on",
+        "method": "GET",
+        "endpoint": "/v5/device(<ID>)",
```
vipulgupta2048, line 345: @pdcastro Making a new thread,
Is there a reason why we have Release UUID, Release ID (Similarly DeviceID, Device UUID, application ID)? I didn't know we even had device ID tbh till I made the API call yesterday.

config/dictionaries/resource.json
pdcastro: Regarding formatting, I think the visible `<pre>` elements look bad. Visible backticks like \`should_be_running__release\` look less bad, :-) but probably also better without them. So, short of fixing the system to allow markdown formatting, my opinion is to remove the quotes.

> You are also right about this, but I still like to go for `is_on__commit` if I want to know what commit the device is running on **right now**

hmm Not sure we are on the same page here. :-)  Comparing these two statements:

* _"To find out the **target** release for an unpinned device, check the `commit` field of the `application` resource instead."_
* _"To find out the **target** release for an unpinned device, check the `is_on__commit ` field instead."_

I suspect that one of the statements above is incorrect -- we should clarify which one before advising users. :-) The statements above offer an alternative to using `should_be_running__release` when it is `null` (because the device is unpinned). ***If it is true that `should_be_running__release` regards target state rather than current state,*** then we should advise users how to obtain a device's **target** state (possibly `application.commit`), rather than a device's **current** state (possibly `device.is_on__commit`).

Note that I am assuming, but have not verified, that:

* `device.should_be_running__release` regards target state
* `device.is_on__commit` regards current state
* `application.commit` regards target state

I leave it on you to verify that the above is correct. :-)  If these assumptions are correct, the description field could actually advise users further:

_"For an unpinned device, the `device.is_on__commit` field may be used to obtain the device's current state, while the `application.commit` field represents the device's target state."_

Again, I don't know that the above is true, please verify. :-)

config/dictionaries/resource.json
pdcastro: I understand that the ID would change if we changed the database "engine", for example by upgrading to a new version of Postgres that had an incompatible database format, or if we changed from Postgres to MySQL, and the like. As such, the ID is an implementation detail that, ideally, would be hidden from users. By contrast, the UUID is eternally immutable, and would survive the database changes mentioned above.

See also this CLI issue: https://github.com/balena-io/balena-cli/issues/1843

Having said that, we have always exposed the numeric database IDs to end users via the balena API, we even document it in examples and projects and (to some extent) in the CLI, so at the moment the IDs are effectively part of the "public API". I suppose we could at any time have an API breaking change that hid database IDs, exposing UUIDs only, and then have a API translation layer that mapped "old" DB IDs to UUIDs, with a performance hit. This is my understanding...

@Page-, does this sound right?

pdcastro: Coincidentally, a user has just created a new CLI issue where they describe using app IDs as a "security issue" that "an attacker could easily leverage" (their words!):

> https://github.com/balena-io/balena-cli/issues/2063
> As an aside, I'd even prefer that the app ids be internal only and not exposed to any customers or customer APIs. By itself this isn't a serious security issue, but it needlessly exposes info that an attacker could easily leverage.

config/dictionaries/resource.json
vipulgupta2048: > "To find out the target release for an unpinned device, check the is_on__commit field instead."

This is incorrect, in my statement earlier I was showing my preference for `is_on__commit` if I want to find the current state of the device that's all. I agree with everything you mentioned above including recommending users an alternative if one is not available. I am making a final change, after which I can ping on beginners for folks to verify it?

I have added formatting changes and changes the description field as suggested.

config/dictionaries/resource.json
```diff
@@ -288,22 +313,43 @@
...
         "method": "PATCH",
         "endpoint": "/v5/device(<ID>)",
         "filters": "",
         "data": "{\n    \"belongs_to__application\": \"<APP ID>\"\n}"
-      },
+	  },
+	  {
+        "id": "get-target-release",
+        "summary": "Get the release device is pinned to",
+        "description": "Note: should_be_running__release will output unset or null if the device isn't pinned to any release. For an unpinned device, the device.is_on__commit field may be used to obtain the device's current state, while the application.commit field represents the device's target state.",
```
pdcastro, line 325: NIT: _"will output"_ sounds a bit odd to me. How about _"will be"_ instead?  
(An application may print some output to the console when it is executed, but a field by itself does not output anything...) :-)

That's focusing on the writing only, not on whether it is true that `is_on__commit` represents the current state while `application.commit` represents the target state.

config/dictionaries/resource.json
vipulgupta2048: @pdcastro Pushed and resolving.

config/dictionaries/resource.json
```diff
@@ -288,14 +313,34 @@
...
+      },
+      {
+        "id": "get-current-release",
+        "summary": "Get current release running on device",
+        "description": "The is_on__commit field fetches the hash of the current release that the device is running on",
+        "method": "GET",
+        "endpoint": "/v5/device(<ID>)",
+        "filters": "?\\$select=is_on__commit",
+        "data": ""
+      },
```
vipulgupta2048, line 339: Need more eyes on this change folks, let us know if this is accurate :eyeglasses:

config/dictionaries/resource.json
```diff
@@ -288,14 +313,34 @@
...
+        "description": "Note: should_be_running__release will output unset or null if the device isn't pinned to any release. For an unpinned device, the device.is_on__commit field may be used to obtain the device's current state, while the application.commit field represents the device's target state.",
+        "method": "GET",
+        "endpoint": "/v5/device(<ID>)",
+        "filters": "?\\$select=should_be_running__release",
+        "data": ""
+      },
+      {
+        "id": "get-current-release",
+        "summary": "Get current release running on device",
+        "description": "The is_on__commit field fetches the hash of the current release that the device is running on",
```
pdcastro, line 334: * NIT: _"fetches"_ sounds a bit odd to me. How about _"contains"_ instead?  
(An application may fetch some data when it is executed, but a field by itself does not fetch anything...) :-)
* Instead of: _"the current release that the device is running on"_  
I'd suggest: _"the release currently deployed to the device"_  
(A device "runs on an ARM CPU", rather than a device "runs on a release"... I think!)

That's focusing on the writing only, not on whether it is true that `is_on__commit` regards current state rather than target state. (Is `is_on__commit` updated by the supervisor?)

config/dictionaries/resource.json
pdcastro: Also timely: https://github.com/balena-io/open-balena-api/pull/491/files

config/dictionaries/resource.json
dt-rush: As for the latter part, @pdcastro that is accurate. Unpinned devices have unset/null `should_be_running__release` and will target to `application.commit`

config/dictionaries/resource.json
dt-rush: I second that for "contains" over "fetches". These are declarative *properties* not computational results (even if technically everything is a computational result 😅)

config/dictionaries/resource.json
vipulgupta2048: @pdcastro are we lucky? or was this just bound to happen all at once :joy:

config/dictionaries/resource.json
vipulgupta2048: Thanks, resolving!

config/dictionaries/resource.json
vipulgupta2048: Thanks, resolving!

pdcastro (approved):

pdcastro:
@balena-ci rebase

config/dictionaries/resource.json
pdcastro: Play the lottery! I suggest using the numbers of the issues and pull requests. 😄

PR #1520, opened 2020-11-13
i2c: fix up minor typos

xginn8:
Change-type: patch


xginn8:
@balena-ci I self-certify!

PR #1540, opened 2020-12-08
Update ssh key instructions

imron:
Add instructions to edit ~/.ssh/config so that it contains an entry for
`ssh.balena-devices.com` that uses the newly added key, otherwise `balena ssh`
won't know which key to use.

Change-type: patch
Signed-off-by: Imron Alston <imron.alston@elenium.com>

### DIFF
--- a/shared/general/container-ssh.md
+++ b/shared/general/container-ssh.md
@@ -55,6 +55,15 @@ You may either import an existing SSH key from GitHub or manually enter the publ
 
 If you do not have an existing key, you can follow [GitHub's documentation][github-ssh], skipping the step about adding the key to your GitHub account, and instead adding the key to your {{ $names.cloud.lower }} account.
 
+Once you have added a key your your account, you should also add an entry for `ssh.balena-devices.com` to your `~/.ssh/config` file so that the `balena ssh` command knows to use this key:
+
+```shell
+Host ssh.balena-devices.com
+  User <USER>
+  IdentityFile ~/.ssh/<PRIVATE_KEY>
+
+```
+
 ## Using a standalone SSH client
 
 If you prefer to use a standalone SSH client to connect to the device, the SSH server on a device listens on TCP port `22222`. While development images have passwordless root access enabled, production images require an SSH key to be added to the `config.json` file.


shared/general/container-ssh.md
```diff
@@ -55,6 +55,15 @@ You may either import an existing SSH key from GitHub or manually enter the publ
 
 If you do not have an existing key, you can follow [GitHub's documentation][github-ssh], skipping the step about adding the key to your GitHub account, and instead adding the key to your {{ $names.cloud.lower }} account.
 
+Once you have added a key your your account, you should also add an entry for `ssh.balena-devices.com` to your `~/.ssh/config` file so that the `balena ssh` command knows to use this key:
+
+```shell
+Host ssh.balena-devices.com
+  User <USER>
+  IdentityFile ~/.ssh/<PRIVATE_KEY>
+
```
vipulgupta2048, line 64:
```suggestion
```

vipulgupta2048: This blank line can be removed from the markdown code snippet

vipulgupta2048 (changes requested):

shared/general/container-ssh.md
imron: Done

shared/general/container-ssh.md
```diff
@@ -55,6 +55,14 @@ You may either import an existing SSH key from GitHub or manually enter the publ
 
 If you do not have an existing key, you can follow [GitHub's documentation][github-ssh], skipping the step about adding the key to your GitHub account, and instead adding the key to your {{ $names.cloud.lower }} account.
 
+Once you have added a key your your account, you should also add an entry for `ssh.balena-devices.com` to your `~/.ssh/config` file so that the `balena ssh` command knows to use this key:
```
pdcastro, line 58: This is not usually necessary though. I am considering 2 cases:

1- You meant to have a simple / standard setup with a single key for all hostnames, but `balena ssh` didn't work for you until you added `ssh.balena-devices.com` to `~/.ssh/config`. In this case, we should debug what went wrong, so that it works without extra configuration in `~/.ssh/config`, and we might then document "watch out for xxx".

2. You have a complex setup with separate ssh keys for separate hostnames, and deliberately chose to use a different key for balena / for `ssh.balena-devices.com`. I understand that ssh key configuration is a step that many users struggle with even in the simple scenario of a single ssh key for all hostnames, so I'd be wary of adding extra setup steps that would become extra opportunity of misconfiguration and failure. If the intention was to document the steps for this complex setup case, then I would suggest adding a separate section in this document that had a title such as:

```
## Using a different key for balena (advanced and optional)

If you need to use a different ssh key for balena than the ssh key(s) you use for
other services, ...
```

By the way, in this second scenario, we should keep in mind that `balena ssh` supports Windows as well, with shells like PowerShell or `cmd.exe`, and the built-in ssh client provided by Microsoft. Does the `~/.ssh` notation even works in `cmd.exe`? Maybe it does, but it's something that needs confirmation before we add such a configuration suggestion.

pdcastro (changes requested):

shared/general/container-ssh.md
imron: It's the second setup.  I have separate ssh keys for separate services, and without the entry in the .ssh/config file it doesn't use the correct key.

The windows version provided by Microsoft supports the `~./ssh` notation, with `~` being mapped to `%USERPROFILE%`.

I'll update shortly with a separate section.

vipulgupta2048:
@imron Checking back again for an update on this PR. Let us know how we can help you further to complete this change?

vipulgupta2048:
@imron Checking back again for an update on this PR. Let us know how we can help you further to complete this change?

PR #1552, opened 2020-12-22
Rename resin-logo.png to balena-logo.png

alexgg:
The OS is rebranding the splash image file from v2.51 onwards.

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>

### DIFF
--- a/pages/reference/OS/advanced.md
+++ b/pages/reference/OS/advanced.md
@@ -119,7 +119,7 @@ dtoverlay=i2c-rtc,ds1307
 
 To disable the Raspberry Pi rainbow splash screen, add the `disable_splash=1` entry to `config.txt`.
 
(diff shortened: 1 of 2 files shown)


shaunmulligan (commented):
@alexgg will the OS still understand and use `resin-logo.png` after 2.51? because i know some customers and I think the CLI still hardcode that name and this means it will break for them. We should also make sure @thgreasi and @pipex know about this change since they have been working on making the logo settable from the dashboard and I don't know if they have included this name change.

alexgg:
> the CLI still hardcode that name and this means it will break for them. 
@shaunmulligan the CLI was updated in https://github.com/balena-io/balena-cli/issues/1993

> making the logo settable from the dashboard
I assume the changes will get tested with recent releases and those already use `balena-logo`. Anyway I will ping them just in case.

alexgg:
@balena-ci retest

vipulgupta2048:
@alexgg is this good to go? I am rebasing your branch and the tests will end up passing to now that we have patched the builds but please do remove the merge commit (https://github.com/balena-io/docs/pull/1552/commits/ff7258faaaafb8497e7b6c34d5ef6b45027df0a6) before merging

vipulgupta2048:
@balena-ci rebase

vipulgupta2048 (approved):
Looking good!

PR #1559, opened 2021-01-12
networking: devices no longer require access to DockerHub

xginn8:
Change-type: patch


shaunmulligan (commented):
Does this have any requirements on OS version?

xginn8:
> Does this have any requirements on OS version?

All supervisors should be served from balenaCloud directly (defined in the backend), so nothing device-dependent.

shaunmulligan (approved):
lgtm

PR #1561, opened 2021-01-12
host OS updates: drop specific DT upgradability references

xginn8:
since any MC-capable device can be upgraded to any other host OS, let's
drop these references.

Change-type: patch


shaunmulligan (approved):
LGTM

PR #1573, opened 2021-01-28
Patch CI/CD builds

vipulgupta2048:
npm install in the dockerfile triggers the prepare.sh script (As postinstall), which needs
to be copied before the install starts. If it's not present then it leads to an error. This is the reason for CI/CD checks failing.

The `prepare.sh` triggers npm build which needs all docs to be present, hence copying all files before npm install is needed in the current setup. 

Change-type: patch


Dockerfile
```diff
@@ -3,11 +3,8 @@ FROM library/node:latest
 RUN mkdir /src
 WORKDIR /src
 
-COPY package.json .
-RUN npm install
-
 COPY . .
-RUN tools/prepare.sh
+RUN npm install
```
xginn8, line 7:
```suggestion
RUN npm ci
```

pdcastro: > npm install in the dockerfile triggers the prepare.sh script (As postinstall), which needs to be copied before the install starts. 

By the way, I always felt a bit uneasy about that `postinstall` script in `package.json`:
```
"postinstall": "./tools/prepare.sh",
```
It feels to me like abusing the scope of `npm install` (a more common use of the `postinstall` script is to apply patches to dependencies, which needs to happen before the app can be used, like we do in the `balena-cli` repo). The `prepare.sh` script in turn calls `fetch-external.sh` which then downloads the markdown files from more than 10 other repos. Doing this in `postinstall` catches developers by surprise, and makes it inconvenient to test with new dependencies and update them -- one would not want to trigger `fetch-external.sh` every time they run `npm install`.

So instead of deleting `tools/prepare.sh` from the Dockerfile, how about deleting the `postinstall` script from `package.json`? Then I think this Dockerfile could have:

```
COPY . .
RUN npm install
RUN tools/prepare.sh
```

"Explicit is better than implicit" - most of the time. :-)

Re `npm ci` vs `npm install`, it works too. Considerations:
* `npm ci` starts by deleting `node_modules`. Note that the `node_modules` folder is listed in the `.dockerignore` file (as it should be), so this step of deleting `node_modules` would be a no-op.
* `npm ci` does not modify the `package-lock.json` file, and requires `package-lock.json` to be consistent with the `pakcage.json` file (no conflicts). This is probably a good thing, but it also requires developers to actually commit changes to `package-lock.json`, which from some experience with the docs repo, developers had not been doing in the past. (I had also found in the past that Node 10 vs 12 vs 14 produced different changes to the `package-lock.json` file in this repo, but this is not really a reason to avoid `npm ci`.)

vipulgupta2048: Thanks @pdcastro @xginn8 I back `npm CI` too. The only issue with it is that package-lock.json needs to be updated as well. With our mutual understanding as codeowners, we can make sure that if someone does update the `package.json`, then `package-lock.json` needs to get committed as well. Doesn't happen that much tbh. 

I made this change in the PR to keep in line with the current structure. But, improving the workflow would be quite good. I agree with Paulo on deleting the `postinstall` in `npm install`. Though this also affects our contributor experience as folks would now need to 

1. npm install
2. Run a script --> That also triggers npm run build
3. npm start

Here's my recommendation, modify the `build` script to also run prepare.sh instead of the other way around. I will remove `npm run build` from `prepare.sh` script

```
"build": "./tools/prepare.sh && ./tools/build.sh && webpack -p",
```
Checkout the commit now.

vipulgupta2048: Would need to edit the README too, so don't merge this yet.

vipulgupta2048:
@balena-ci retest

Dockerfile
```diff
@@ -1,13 +1,12 @@
...
 WORKDIR /src
 
-COPY package.json .
-RUN npm install
+COPY package.json package-lock.json ./ 
+RUN npm ci
 
 COPY . .
-RUN tools/prepare.sh
+RUN npm build
```
pdcastro, line 9: `npm build` is an error:
```
Step 6/9 : RUN npm build
 ---> Running in 2d57f00f9ad7
Top hits for "build"
————————————————————————————————————————————————————————————————————————————————
npm help install                                                         build:6
npm help config                                                          build:5
npm help explore                                                         build:5
npm help package-json                                                    build:4
npm help registry                                                        build:4
npm help rebuild                                                         build:4
npm help version                                                         build:4
npm help developers                                                      build:2
npm help semver                                                          build:2
npm help ci                                                              build:1
————————————————————————————————————————————————————————————————————————————————
(run with -l or --long to see more context)
```
It would be `npm run build`.

Weird thing that that be build succeeds (`docker build` succeeds) even though this RUN instruction fails.

Dockerfile
```diff
@@ -1,13 +1,12 @@
...
 
-RUN mkdir /src
 WORKDIR /src
 
-COPY package.json .
-RUN npm install
+COPY package.json package-lock.json ./ 
+RUN npm ci
 
 COPY . .
```
pdcastro, line 8: Why COPY twice? (Not saying it is not needed, but I don't see the reason.) Wouldn't the following work?
```
COPY . .
RUN npm ci
RUN npm run build
```
or even shorter:
```
COPY . .
RUN npm ci && npm run build
```
(Single COPY)

vipulgupta2048: @pdcastro Thought we follow best practice here of incremental builds. 

> Using this separation, if there are no changes in package.json or package-lock.json then the cache will be used for the layer generated by the RUN npm ci instruction. This means that when you edit your application source and rebuild, the dependencies won’t be redownloaded which saves time

Reference: https://www.docker.com/blog/speed-up-your-development-flow-with-these-dockerfile-best-practices/

Dockerfile
vipulgupta2048: Weird indeed, and I always miss this. Thanks for the catch @pdcastro

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
@pdcastro I think it looks good to go.

Dockerfile
pdcastro: That's a good one! 👍

pdcastro (approved):
Good work 👍 
I've taken the liberty of adding a commit that deletes the `Dockerfile.bindmount` file, which would also have needed updates re `npm run build`. In the past, I had added that file as a workaround for a `doxx` memory issue on macOS:
```
    balena-io/docs/node_modules/@resin.io/doxx/doxx.js:16	
        throw err	
        ^	
    RangeError: Maximum call stack size exceeded	
```
But I've tested today, and that memory issue no longer happens. Not sure when/how it got fixed, but it got fixed and thus the `Dockerfile.bindmount` is no longer useful as far as I can tell.

Dockerfile
pdcastro: Talking of Docker layer caching, I've noticed that `docker build .` for this repo will use the cache for all layers (will not rebuild the image) if the external markdown files changed (from repos like the supervisor, CLI, etc), but nothing else changed. I believe (not 100% sure) that this was a pre-existing issue before this PR, even when the `postinstall` script was used. I am not sure that this is a problem in practice (I suspect it is not), but if it was, stackoverflow has suggestions for forcing cache invalidation using ARG and RUN:
https://stackoverflow.com/questions/35134713/disable-cache-for-specific-run-commands

PR #1583, opened 2021-02-05
Add Offline Updates documentation

vipulgupta2048:
Change-type: minor

### DIFF
--- a/pages/learn/deploy/offline-updates.md
+++ b/pages/learn/deploy/offline-updates.md
@@ -1,0 +1,228 @@
+---
+title: Offline updates
+excerpt: An efficient approach for updating previously deployed IoT devices when they’re offline by design.
+---
+
+# Offline updates
+
+Offline updates is an efficient process to update your device without needing an internet connection. It involves preloading a image and reflashing it to on a device in a way that it retains its unique balena identity, apps, and configurations. This is helpful in scenarios where device's internet access is unavailable, limited, blocked or even when it's air-gapped. 
+
+## Performing an offline upgrade 
+
+To perform an offline upgrade, we will be using [balena-cli](https://www.balena.io/docs/reference/balena-cli/). All commands should work on Linux distributions running Docker on Linux Kernel with AUFS and overlay filesystem support. We will be extending support and more features to other platforms too where balena-cli is available such as Windows and MacOS soon. 
+
+> Note: In an offline upgrades process, all data on device is wiped at this point, making this different from a typical software update process in the balena ecosystem. In this process, any additional user data and system settings written to various device partitions would be lost.
+
+## Overview of the process
+
+<Add what will be happening over the next steps>
+
+Offline updates includes the following steps
+
+- [overview](#overview)
+- [setup](#setup)
+- [create application](#create-application)
+- [create offline device](#create-offline-device)
+- [download balenaOS image](#download-balenaos-image)
+- [configure balenaOS image](#configure-balenaos-image)
+- [create and preload release](#create-and-preload-release)
+- [create update media](#create-update-media)
+- [update device registration(s)](#update-device-registrations)
+
+### Setup
+
+> Download and install [balena-cli](https://www.balena.io/docs/reference/balena-cli/) on your linux distributions. The commands have been tested to work on Ubuntu 20.04.
+
+#### Logging in
+
+Several `balena-cli` commands require access to your balenaCloud account. Those commands require creating a CLI login session by running [`balena login`](https://www.balena.io/docs/reference/balena-cli/#logging-in) in your terminal. 
+
+```
+$ balena login
+```
+
+Store the filename of your OpenSSH private key in an environment variable or generate an SSH key using the commands below. (MENTION WHY)
+
+```
+# (optional) if you already have SSH key(s)
+$ ssh-keygen -o -a 100 -t ed25519 -f id_ed25519 -C 'offline-updates' -N ''
+
+# must contain the filename of your OpenSSH private key
+$ ssh_key=<NAME OF FILENAME>.pub
+```
+
+#### Create an application
+
+Initalise the `arch` and `device-type` environment variable with correct CPU architecture (`aarch64`, `armv7l`, etc.) and device type (`raspberrypi4-64`, `raspberrypi3`, etc.). The list of names for supported device types and their architectures can be found on the [hardware][supportedDevicesList] page.
+
+```
+$ arch=<CPU ARCHITECTURE TYPE>
+$ device_type=<DEVICE_TYPE>
+$ app_name=offline-${arch}
+```
+
+Create a new balenaCloud application by running [`balena app create`](https://www.balena.io/docs/reference/balena-cli/#app-create-name)
+
+```bash
+$ balena app create ${app_name} --type ${device_type}
+
+$ app_slug=$(balena app ${app_name} | grep SLUG | awk '{print $2}')
+```
+
+## create offline device
+
+multiple devices can be created using this approach
+
+```bash
+uuid=$(uuidgen | tr '[:upper:]' '[:lower:]' | sed 's/-//g')
+
+balena device register ${app_slug} --uuid ${uuid}
+```
+
+## download balenaOS image
+> download the latest version on the balenaOS
+
+```bash
+$ os_version=$(balena os versions ${device_type} | grep prod | head -n 1 | awk '{print $1}')
+
+$ tmpimg=$(mktemp).img
+
+$ balena os download ${device_type} \
+  --output ${tmpimg} \
+  --version ${os_version}
+```
+
+## configure balenaOS image
+> preserving the identify of the registered device `uuid`
+
+```bash
+$ tmpconfig=$(mktemp)
+
+$ config=$(mktemp)
+
+$ balena config generate \
+    --device ${uuid} \
+    --version ${os_version} \
+    --network ethernet \
+    --appUpdatePollInterval 10 \
+    --output ${tmpconfig} \
+    && cat ${tmpconfig} | jq .
+
+$ cat < ${tmpconfig} \
+    | jq --arg keys "$(cat ${ssh_key})" '. + {os: {sshKeys: [$keys]}}' > "${config}" \
+    && cat < ${config} | jq .
+
+$ balena os configure ${tmpimg} \
+    --app ${app_slug} \
+    --device-type ${device_type} \
+    --version ${os_version} \
+    --config-network ethernet \
+    --config ${config}
+
+$ rm ${tmpconfig}
+$ rm ${config}
+```
+
+## create and preload release
+> [preload](https://github.com/balena-io/balena-cli/blob/master/INSTALL-MAC.md#balena-preload) functionality requires Docker with AUFS support
+
+```bash
+$ date +%s > app/.epoch
+
+$ balena deploy ${app_slug} --build --source .
+
+$ commit=$(balena app ${app_slug} | grep COMMIT | awk '{print $2}')
+
+$ balena preload ${tmpimg} \
+    --app ${app_slug} \
+    --commit ${commit} \
+    --pin-device-to-release
+```
+
+## create update media
+> use this media to boot or flash your offline device(s)
+
+```bash
+# list the devices available to flash
+$ balena util available-drives
+/dev/...
+
+# ⚠️ please make sure `--drive` is really the disk you want to write to
+$ sudo balena local flash ${tmpimg} \
+    --drive /dev/{{flash-usb-disk}}
+
+$ rm ${tmpimg}
+```
+
+## update device registration(s)
+>  manually reflect the new state of the device in balenaCloud for offline devices or devices on private networks without Internet access
+
+### patch state endpoint
+> strictly not required for devices with low bandwidth Internet connections as they will update the state endpoint automatically
+
+```bash
+$ os_version_tag=$(echo ${os_version} | awk -F'+' '{print $1}')
+$ os_version_semver=$(echo ${os_version_tag} | sed 's/v//g')
+$ os_revision=$(echo ${os_version} | awk -F'+' '{print $2}' | awk -F'.' '{print $1}')
+$ os_variant=$(echo ${os_version} | awk -F'+' '{print $2}' | awk -F'.' '{print $2}')
+
+$ supervisor_version=$(curl --silent https://raw.githubusercontent.com/balena-os/meta-balena/${os_version_tag}/meta-balena-common/recipes-containers/resin-supervisor/resin-supervisor.inc \
+    | grep SUPERVISOR_TAG \
+    | awk '{print $3}' \
+    | sed 's/"//g' \
+    | sed 's/v//g')
+
+  # (e.g.) staging
+$ release_id=$(curl --silent \
+	  "https://api.balena-staging.com/v6/release?\$filter=startswith(commit,'${commit}')&\$select=id" \
+	  -H "Authorization: Bearer $(cat ~/.balena/token)" | jq -r .d[].id)
+
+$ curl --silent \
+	  -X 'PATCH' "https://api.balena-staging.com/v6/device(uuid='${uuid}')" \
+	  -H "Authorization: Bearer $(cat ~/.balena/token)" \
+	  -H 'Content-type: application/json' \
+	  --data-binary "{\"os_variant\":\"${os_variant}\",\"os_version\":\"balenaOS ${os_version_semver}+${os_revision}\",\"supervisor_version\":\"${supervisor_version}\",\"is_running__release\":${release_id}}"
+```
+
+### (optional) set tags
+
+```bash
+$ balena tag set 'offline:commit' "${commit}" \
+  --device ${uuid}
+
+$ balena tag set 'offline:hostOS' "${os_version}" \
+  --device ${uuid}
+```
+
+- generate a device-specific `config.json` file for the device
+- insert the custom `config.json` file into the downloaded balenaOS image
+- run `balena preload` on the image, pointing to whichever version of your code you want to deploy
+- flash the image to an SD card, a device’s eMMC or a separate (flasher) USB drive
+- update device state on balenaCloud and optionally tag the device
+
+
+# Reflash any device offline without losing its apps or identity
+Today, when you reflash a previously deployed device, you’re resetting it to a factory default state. Your device gets a new identity, a new API key, and updated `config.json` settings. All your apps, data and logs stored on the device are lost. 
+
+With the Offline Upgrades process, you’ll still be resetting the device to factory default state, while restoring its identity in `config.json`, and then using `balena preload` to load your updated applications. That way, your device will pick up right where it left off with the same name and UUID.
+
+
+### Remotely updating with a SD card or USB device
+
+If a device isn’t local, you can just ship the SD card or USB stick/drive to a remote location and let someone there run the update by simply inserting it into the device and booting.
+
+If the target device exists on an air-gapped or Internet restricted network, inserting [ssh](https://www.balena.io/docs/reference/OS/configuration/#sshkeys) keys during the configuration step will allow fleet managers at the remote site to connect into the device directly via OpenSSH and verify the update by examining container logs, etc.
+
+In the event the target device is connected via a low-bandwidth connection, it should eventually establish a connection to balenaCloud, and depending on the quality of the connection, it may respond to Web Terminal commands and output container logs to the dashboard.
+
+## Settings and data recovery
+
+Most common system settings can be (re)specified via [config.json](https://www.balena.io/docs/reference/OS/configuration/#valid-fields).
+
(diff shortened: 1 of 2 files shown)


pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,269 @@
...
+  --device ${uuid}
+```
+
+Read more about it: blog link goes here
+Here's an example project to test things out https://github.com/balena-io-playground/offline-updates/
+
+
+
+# Reflash any device offline without losing its apps or identity
+Today, when you reflash a previously deployed device, you’re resetting it to a factory default state. Your device gets a new identity, a new API key, and updated `config.json` settings. All your apps, data and logs stored on the device are lost.
```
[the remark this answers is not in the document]
vipulgupta2048: I would remove the statement since this doesn't effect anything user side tbh. Their device work need the key to talk to the API as long as that works, I think we should be good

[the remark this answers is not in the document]
vipulgupta2048: @ab77 Check it out now, how's this for wording I have moved this section to the top as the overview of the entire process. (2nd para)

vipulgupta2048:
@ab77 I need the WHY part of device registration. Why are we doing the 3 steps in te device registration part.

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,273 @@
+---
+title: Offline updates
+excerpt: An efficient approach for updating previously deployed IoT devices when they’re offline by design.
+---
+
+# Offline updates
```
ab77, line 6: @vipulgupta2048 we need to standardise on the terminology. In the blog we are calling these "offline upgrades".

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Store the filename of your OpenSSH private key in an environment variable or generate an SSH key using the commands below. The SSH keys will be added to the device and will be used to SSH into the device after provisioning that will help 
+
+```bash
+# (optional) if you already have SSH key(s)
+$ ssh-keygen -o -a 100 -t ed25519 -f id_ed25519 -C 'offline-updates' -N ''
+
+# must contain the filename of the OpenSSH private key
+$ ssh_key= id_ed25519.pub
```
ab77, line 70: @vipulgupta2048 there should be no space, this command is invalid otherwise.

vipulgupta2048: Thanks @ab77 Good catch

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Update Device Registration(s)
+
+This section deals with device registration.
+
+>  manually reflect the new state of the device in balenaCloud for offline devices or devices on private networks without Internet access
+
+#### [Optional] patch state endpoint
+
+This step is strictly not required for devices with low bandwidth internet connections as those devices would be able to update the state endpoint automatically.
```
thgreasi, line 223: Let me also add that for devices with low bandwidth internet connections, the user has probably already pinned the device to the release it was running before, so that the device doesn't try to start downloading any new release they create.
Given that, it might actually make sense to also set the `should_be_running__release` field in the script below.
What do you think @ab77 ?

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,273 @@
...
+
+The following steps will flash the latest release onto the balenaOS image downloaded in the previous steps along with pinning the device to said release. 
+
+```bash
+$ commit=$(balena app ${app_slug} | grep COMMIT | awk '{print $2}')
+
+$ balena preload ${tmpimg} \
+    --app ${app_slug} \
+    --commit ${commit} \
+    --pin-device-to-release
```
thgreasi, line 177: For devices with low bandwidth internet connections, the user has probably already pinned the device to the release it is running, so that the device doesn't try to start downloading any new release they create.
Given that, it might actually make sense to change this to `--no-pin-device-to-release` and do the pinning in the `patch state endpoint` script by adding
```
,\"should_be_running__release\":${release_id}
```
Such users will have to run that script then before they boot their device.
What do you think @ab77 ?

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,273 @@
...
+
+# Offline updates
+
+Offline updates is an efficient process to update devices without needing an internet connection. It involves preloading a image and reflashing it to on a device in a way that it retains its unique balena identity, apps, and configurations. This is helpful in scenarios where device's internet access is unavailable, limited, blocked or even when it's air-gapped. 
+
+## Overview of the process
+
+When a device is reflashed, it defaults back to a factory state. The device gets a new identity, a new API key, and updated `config.json` settings. All services, data and logs stored on the device is erased permanently. 
+
+With the offline upgrades process, the device still resets to a factory default state, all the preserving its identity in `config.json` with a new API key and then using `balena preload` to load an updated application release. That way, the device will pick up right where it left off with the same name and UUID but with an updated application release or/and balenaOS upgrade. Broad steps in the process include:
```
thgreasi, line 14: > all the preserving its identity in `config.json` with a new API key

I don't get that. Do you mean v?

```suggestion
With the offline upgrades process, the device still resets to a factory default state and a new API key, while preserving its identity in `config.json` and then using `balena preload` to load an updated application release. That way, the device will pick up right where it left off with the same name and UUID but with an updated application release or/and balenaOS upgrade. Broad steps in the process include:
```

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,273 @@
...
+
+In the event the target device is connected via a low-bandwidth connection, it should eventually establish a connection to balenaCloud, and depending on the quality of the connection, it may respond to Web Terminal commands and output container logs to the dashboard.
+
+### Update Device Registration(s)
+
+This section deals with device registration.
+
+>  manually reflect the new state of the device in balenaCloud for offline devices or devices on private networks without Internet access
+
+#### [Optional] patch state endpoint
```
thgreasi, line 221: Reminder mark this as non optional if we also need to cover the "online but low bandwidth" use case, which imo add even more value to offline updated than just airgapped devices.
I mean, if it's a permanently offline device, why are you using balenaCloud? You can just keep flashing and shipping new SD cards.

pages/learn/deploy/offline-updates.md
vipulgupta2048: Oh yeah, that doesn't look good. Here's the improved snippet. 

```
With the offline updates process, the device still resets to a factory default state and adds a new API key, while preserving its identity in `config.json`. It uses [`balena preload`][balena-preload] to load an updated application release. That way, the device will pick up right where it left off with the same name and UUID but with an updated application release or/and balenaOS upgrade. Broad steps in the process include:
```

pages/learn/deploy/offline-updates.md
vipulgupta2048: Removed optional tag.

vipulgupta2048:
> @ab77 I need the WHY part of device registration. Why are we doing the 3 steps in the device registration part? Got any good description we can add?

Still need, would be quite helpful to understand what we mean to do in device registration @ab77

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,271 @@
...
+Offline update revolve around the concept of [balena preload][balena-preload]. Preload is used to flash the balenaOS image and your application release in a single step, so the device starts running your application containers as soon as it boots. Preloading removes the need for your devices to download the initial application images directly from balena's build servers, making it an ideal base for the offline update process. Read more about [preloading a device image][preloading-device-image].
+
+> Important note: [balena preload](https://github.com/balena-io/balena-cli/blob/master/INSTALL-MAC.md#balena-preload) functionality requires Docker with AUFS support.
+
+Since preload involves flashing an application release with the balenaOS image, if the pre-existing application doesn't have any releases, then a release needs to be created. If present already, then the next step can be skipped. 
+
+```bash
+$ date +%s > app/.epoch
+
+$ balena deploy ${app_slug} --build --source .
```
ab77, line 163: @vipulgupta2048 @thgreasi we do need to add `--emulated` flag is these instructions are running under Linux (Ubuntu 20.04 LTS). I tested today and without emulated, I get an exec error during the build/deploy.

pages/learn/deploy/offline-updates.md
ab77: @thgreasi I really don't like that wall of bash text to patch state endpoints, so I don't really have an opinion. Ideally, we should go with whatever reduced the number of steps or even more ideally, have the CLI perform these as part of the preload flow (if a flag is specified presumably) cc @pdcastro

vipulgupta2048:
@balena-ci rebase

ab77:
> @ab77 I need the WHY part of device registration. Why are we doing the 3 steps in the device registration part? Got any good description we can add?

If you are referring to the part `## create offline device`, then the WHY is because all the the subsequent steps require it. If the user already has a device, they just need to set `uuid={{their-uuid}}`.

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Update Device Registration(s)
+
+This section deals with device registration.
+
+> manually reflect the new state of the device in balenaCloud for offline devices or devices on private networks without Internet access
+
+#### Patch State Endpoint
+
+This step is strictly not required for devices with low bandwidth internet connections. Those devices would be able to update the state endpoint automatically.
```
pdcastro, line 218: The word "strictly" confuses me in the sentence. Would it still mean the same if the word "strictly" was deleted?

Even then, whether or not this step is required would depend on the device having "low bandwidth", which is a loosely defined condition. As a user, how can I determine, with confidence, whether or not this step applies to me? _"Just tell me whether or not I should run this"_ kind of thing. :-)

pages/learn/deploy/offline-updates.md
pdcastro: > [...] even more ideally, have the CLI perform these as part of the preload flow (if a flag is specified presumably) 

Yes, sounds good to me. The wall of bash text could be replaced with a balena-cli PR. To decide on the name of the flag, what is a short description of the _purpose_ of patching the state endpoint? Why do we need / want to patch the state endpoint? Things like `--patch-state` or `--fix-state` sound rather generic...

pdcastro: But I am also not saying that offline updates should be delayed until more work is done on preload - up to you.

pages/learn/deploy/offline-updates.md
ab77: I suppose if users are blindly copying and pasting instructions without understanding what they do, having something like this forces some sort of consideration. Could be reworded to say that if a device never ever phones home, then this step is required (unless the user doesn't care how the device appears in the dash board at all).

pdcastro: > Could be reworded to say that if a device never ever phones home, then this step is required (unless the user doesn't care how the device appears in the dash board at all).

This would definitely be clearer. 👍  For example, _assuming it is correct:_
```
#### Update the device state in the cloud

This step updates the device state in the cloud (balenaCloud or openBalena)
on behalf of the device, as if the device was online. It is only required for
devices that do not have an internet connection. If the device has an internet
connection, even if low bandwidth, this step may be skipped because the device
itself will contact the cloud to update its state.
```
Assuming it is correct, this also answers my other question as to what the `balena preload` flag might be called:
```
$ balena help preload
OPTIONS
  --update-cloud-state  Update the device state in the cloud (for offline updates)
```

vipulgupta2048:
> @vipulgupta2048 If you are referring to the part ## create offline device, then the WHY is because all the the subsequent steps require it. If the user already has a device, they just need to set uuid={{their-uuid}}.

No, I don't mean this part. I meant, the actual device registration part with this heading, `### Update Device Registration(s)`. A description for this part would be nice to understand what this means to the user + what the next steps would entail.

pages/learn/deploy/offline-updates.md
vipulgupta2048: @ab77 Cool, Adding an emulated flag until someone says otherwise.

pages/learn/deploy/offline-updates.md
vipulgupta2048: That's a really good description. I am adding it, until someone says otherwise. 
cc: @ab77 @pdcastro

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,276 @@
...
+$ os_revision=$(echo ${os_version} | awk -F'+' '{print $2}' | awk -F'.' '{print $1}')
+$ os_variant=$(echo ${os_version} | awk -F'+' '{print $2}' | awk -F'.' '{print $2}')
+
+$ supervisor_version=$(curl --silent https://raw.githubusercontent.com/balena-os/meta-balena/${os_version_tag}/meta-balena-common/recipes-containers/resin-supervisor/resin-supervisor.inc \
+    | grep SUPERVISOR_TAG \
+    | awk '{print $3}' \
+    | sed 's/"//g' \
+    | sed 's/v//g')
+
+# (e.g.) staging
```
pdcastro, line 238: Instead of this `# (e.g.) staging` comment, how about replacing the two occurrences of `balena-staging.com` with an environment variable? The CLI uses the `BALENARC_BALENA_URL` environment variable to select a different cloud domain name (openBalena, staging, ...) than the `balena-cloud.com` default, so I suggest:
```
if [ -z "$BALENARC_BALENA_URL" ]; then
    BALENARC_BALENA_URL=balena-cloud.com
fi

$ release_id=$(curl --silent \
    "https://api.${BALENARC_BALENA_URL}/v6/release?
...

$ curl --silent \
    -X 'PATCH' "https://api.${BALENARC_BALENA_URL}/v6/device(uuid='${uuid}')" \
...
```
(Yes, ideally this bash code would be replaced with a new flag for the `balena preload` command.)

pages/learn/deploy/offline-updates.md
pdcastro: > so that the device doesn't try to start downloading any new release they create [as soon as the `preload` command is executed]

This is a good point. But even moving the pinning to the patch state endpoint step (which is currently described as an optional step for low bandwidth devices) would still require coordination with the physical updating of each device in a strict order, for example:

1. Flash the media with the image created by `preload` (without release pinning).
2. Take the device offline for the update (rebooting would do).
3. While the device is offline / before it boots, patch the state endpoint to pin the release.

Now, picture devices in multiple remote sites, where the operators doing the flashing do not have a laptop with them to patch the state endpoint then and there for that device...

There is a conflict here between "really offline" and _"online with low bandwitdh, could download images but don't want to, it would likely fail anyway while wasting bandwidth and money"._ For the latter, perhaps we needed a new supervisor feature like "low bandwith mode - don't ever try to download images, await for offline updates".

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Create and Preload Release
+
+Offline update revolve around the concept of [balena preload][balena-preload]. Preload is used to flash the balenaOS image and your application release in a single step, so the device starts running your application containers as soon as it boots. Preloading removes the need for your devices to download the initial application images directly from balena's build servers, making it an ideal base for the offline update process. Read more about [preloading a device image][preloading-device-image].
+
+> Important note: [balena preload](https://github.com/balena-io/balena-cli/blob/master/INSTALL-MAC.md#balena-preload) functionality requires Docker with AUFS support.
+
+Since preload involves flashing an application release with the balenaOS image, if the pre-existing application doesn't have any releases, then a release needs to be created. If present already, then the next step can be skipped.
+
+```bash
```
ab77, line 163: @vipulgupta2048 we should mention that this deploys a quick [demonstration project](https://github.com/balena-io-playground/offline-updates#sample-application). I don't believe this is mentioned anywhere.

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,276 @@
...
+$ balena tag set 'offline:commit' "${commit}" \
+  --device ${uuid}
+
+$ balena tag set 'offline:hostOS' "${os_version}" \
+  --device ${uuid}
+```
+
+Read more about how the process can work better for your usecase in the [offline updates](https://balena.io/blog/offline-updates-make-it-easier-to-update-balena-devices-without-the-internet) blog.
+Here's an [example project](https://github.com/balena-io-example/offline-updates/) to demonstrate offline updates for offline devices with user data mounted on external storage media persisting between hostOS and application updates.
```
ab77, line 263: @vipulgupta2048 this "example project" doesn't exist. Is the intention here to point somewhere? Just to note that we can't point to https://github.com/balena-io-playground/offline-updates directly, so it would have to be moved to some other relevant GH org, where all of our docs refer to examples. If we are going to move this, we would need to refactor the README, since the content would already be in this doc.

pages/learn/deploy/offline-updates.md
vipulgupta2048: The only problem with this @pdcastro is that it doesn't fit into the overall command structure we have going on in the document. Everything is a shell command and here comes a bash script IF statement. I can make this into a one-liner though.

`if [ -z "$BALENARC_BALENA_URL" ]; then  BALENARC_BALENA_URL=balena-cloud.com; fi`

vipulgupta2048: Replacing stuff with this.

vipulgupta2048: @pdcastro Do you think its a good idea to add somewhere that this all is experimental as the team will greatly smoothen out the entire process as things come along. I think its worth it to have it somewhere as the blog does it too. 

cc: @ab77

pages/learn/deploy/offline-updates.md
vipulgupta2048: @ab77  Totally upto you, folks regarding the same. As Alex M and Andrew has mentioned before. Any example or project being referenced into the docs or blogs have to be in the `github.com/balena-io-examples` repo. Atleast that's the plan what I read in the blog as well. I do agree on keeping one source of truth and the documentation could be that. Easier to maintain for everyone.

pages/learn/deploy/offline-updates.md
vipulgupta2048: @ab77 as far as the docs are concerned we aren't deploying any demonstration project because the docs are meant to be general. So I have purposely removed the part about deploying a demonstration project. 

Rather I have put an example project in the end that takes folks through the entire process with a demonstration project being deployed and everything. That could the core difference between docs (general) and example project (opinionated and reference)

pages/learn/deploy/offline-updates.md
```diff
@@ -0,0 +1,276 @@
...
+### Create and Preload Release
+
+Offline update revolve around the concept of [balena preload][balena-preload]. Preload is used to flash the balenaOS image and your application release in a single step, so the device starts running your application containers as soon as it boots. Preloading removes the need for your devices to download the initial application images directly from balena's build servers, making it an ideal base for the offline update process. Read more about [preloading a device image][preloading-device-image].
+
+> Important note: [balena preload](https://github.com/balena-io/balena-cli/blob/master/INSTALL-MAC.md#balena-preload) functionality requires Docker with AUFS support.
+
+Since preload involves flashing an application release with the balenaOS image, if the pre-existing application doesn't have any releases, then a release needs to be created. If present already, then the next step can be skipped.
+
+```bash
+$ date +%s > app/.epoch
```
ab77, line 164: @vipulgupta2048 this line needs to go, since it's tightly coupled to the example/demo: https://github.com/balena-io-playground/offline-updates

It needs to be replaced with something to tell the user to switch to their app context (or, probably less desirably, a demo project we host). Less desirably, because we add more code to maintain as a result.

vipulgupta2048: I was actually wondering to ask you about this, this line made no sense to the process as a whole. 

But, what does this mean `tell the user to switch to their app context`
Are you referring to instructions or commands that the user should run on basis of their app and how they got things set up?

ab77:
@vipulgupta2048 LGTM

ab77 (approved):

vipulgupta2048:
@balena-ci I self-certify!

pages/learn/deploy/offline-updates.md
thgreasi: @ab77 

If the device is really always offline, then it doesn't make any difference if they `--pin-device-to-release` or pin it in the `Patch State Endpoint` or don't pin at all (but in that case there is also the question of why are they paying for balenaCloud at the first place).

On the other hand, as @pdcastro said,
If the device has partial/slow/expensive connectivity, the `--pin-device-to-release` would issue the device to try to fetch the new release. I realize that having to run a curl before booting the new preloaded image doesn't sound great, but:
* it should be do-able for an operator to do that, given that the device already has limited connectivity
* It unlocks an extra use case, which imo sounds like what customers probably need

It would of course be great to have a way to instruct a device to not try to download any new release as Paulo said.

PR #1585, opened 2021-02-08
runtime: drop references to SUPERVISOR_VERSION env var

xginn8:
clean up some stylizations of 3ps

Connects-to: https://github.com/balena-io/balena-supervisor/pull/1469
Change-type: patch


xginn8:
@balena-ci I self-certify!

Issue #1587, opened 2021-02-11
API: document device_tag PATCH method


pdcastro:
There are docs for GET and POST:
https://www.balena.io/docs/reference/api/resources/device_tag/#create-tag

But it's missing a PATCH example:

```
curl -X PATCH \
  "https://api.balena-cloud.com/v6/device_tag?\$filter=device/uuid%20eq%20'$BALENA_DEVICE_UUID'" \
  -H "Authorization: Bearer $BALENA_API_KEY" \
  -H "Content-Type: application/json" \
  --data '{ "tag_key": "c", "value": "e" }'
```

vipulgupta2048:
@pdcastro The POST example is demonstrating the same functionality, is it necessary to have a PATCH example too?

pdcastro:
> @pdcastro The POST example is demonstrating the same functionality, is it necessary to have a PATCH example too?

@vipulgupta2048, I think so, because:
* The POST example does not quite demonstrate the same functionality or syntax. Using POST for an existing device tag results in an error: ""device" and "tag_key" must be unique". The POST request uses a numeric device ID in the request body, while the PATCH request uses a device UUID in a $filter query string.
* The user in the support thread was looking for an example of how to update a device tag using the API and couldn't find it or figure it out by themselves, then two support agents (one of them being me) also had to spend time trying to come up with a working example to help the user. If the user or the support agents were more knowledgeable about the API requests and resources, the extra examples would not be needed, but this wasn't the case.
* As Alexandros usually says, a user who comes to support represents many other users who struggled in silence or gave up.

I have created PR #1588 to add the examples - pending review.

PR #1578, opened 2021-01-30
Patch v6 fields, ID and examples

vipulgupta2048:
Updated all fields directly from the API, with minor corrections to formatting as needed. 

Also resolves: #1570  #1565  
Change-type: patch
Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>

PR #1588, opened 2021-02-11
Add PATCH and DELETE examples to the API documentation of device tags

pdcastro:
Resolves: #1587
Change-type: patch

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -529,6 +529,23 @@
         "endpoint": "/v6/device_tag",
         "filters": "",
         "data": "{\n    \"device\": \"<DEVICE ID>\",\n    \"tag_key\": \"<KEY>\",\n    \"value\": \"<VALUE>\"\n}"
+      },
+      {
+        "id": "update-tag",
+        "summary": "Update a device tag",
+        "description": "",
+        "method": "PATCH",
+        "endpoint": "/v6/device_tag",
+        "filters": "?\\$filter=device/uuid%20eq%20'<UUID>'",
+        "data": "{\n    \"tag_key\": \"<EXISTING-KEY>\",\n    \"value\": \"<NEW-VALUE>\"\n}"
+      },
+      {
+        "id": "delete-tag",
+        "summary": "Delete a device tag given its database ID",
+        "description": "",
+        "method": "DELETE",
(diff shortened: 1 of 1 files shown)


vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
Thanks @pdcastro this looks good to me. Though I do need an assist, in the recent PR of patching the v6 endpoints #1578, I have missed out on some backlashes here before `$`

https://github.com/balena-io/docs/blob/8beabd1334e004a5d738b726a346eabb41638a1e/config/dictionaries/resource.json#L39

Can you add that into your commit, thanks!

vipulgupta2048 (approved):

config/dictionaries/resource.json
```diff
@@ -36,8 +36,8 @@
         "summary": "Get all applications associated with authenticated user",
         "description": "",
         "method": "GET",
-        "endpoint": "/v6/application?$filter=is_directly_accessible_by__user/any(dau:1%20eq%201)",
-        "filters": ""
+        "endpoint": "/v6/application",
+        "filters": "?\\$filter=is_directly_accessible_by__user/any(dau:1%20eq%201)"
```
pdcastro, line 40: Thanks for fixing this @vipulgupta2048. 👍 
>  I have missed out on some backlashes here

Btw, if one had the time to do these improvements... (But I don't think it's a priority.) Instead of escaped strings, the `filters` and `data` properties _everywhere_ could be given as JSON objects, which could then get automatically translated into appropriately escaped strings depending on targets:
* Target: curl on bash
* Target: curl on PowerShell
* Target: curl on Windows cmd.exe
* Target: web browser URL bar
* Target: JSON
* Target: Javascript

I think that each of the targets above may need different escaping characters (e.g. single backslash, double backslash, caret instead of backslash, single quotes, double quotes, % escaping, etc), at least in some corner cases.

Currently:
```
        "filters": "?\\$filter=device/uuid%20eq%20'<UUID>'",
        "data": "{\n    \"tag_key\": \"<EXISTING-KEY>\",\n    \"value\": \"<NEW-VALUE>\"\n}"
```
What it could be instead:
```
        "filters": ["device/uuid eq '<UUID>'"],
        "data": {
                "tag_key": "<EXISTING-KEY>",
                "value":  "<NEW-VALUE>"
        }
```

PR #1591, opened 2021-02-16
Remove sections to configure delta updates

20k-ultra:
We no longer want to mention how to disable deltas because it will be hidden in the UI so that it is always used. The environment variable is still available so modifying the Supervisor response data that includes the variable is not correct.

Change-type: patch
Signed-off-by: Miguel Casqueira <miguel@balena.io>


pages/learn/manage/configuration.md
```diff
@@ -17,7 +17,6 @@ This list contains configuration variables that can be used with {{ $names.compa
 Name | Default | Description | Valid from
 --- | --- | --- | ---
 RESIN_SUPERVISOR_CONNECTIVITY_CHECK | true | Enable / Disable VPN connectivity check | v1.3.0
-RESIN_SUPERVISOR_DELTA | false | Enable / Disable [delta updates][deltas] | v1.7.0
```
xginn8, line 20: be sure to drop the footnote link if it's unused too

20k-ultra: godo catch, updated. I also missed some image files but those are removed now too.

xginn8 (approved):

PR #1600, opened 2021-03-02
Add instructions for changing billing plans

alisondavis17:
Connects-to: balena-io/docs#1564
Connects-to: balena-io/docs#1566
Connects-to: balena-io/docs#1592
Change-type: minor
Signed-off-by: Alison Davis Riddell alison@balena.io

PR #1606, opened 2021-03-02
Add images for updated billing docs

alisondavis17:
Change-type: minor
Signed-off-by: Alison Davis Riddell alison@balena.io


vipulgupta2048 (commented):
Image names have to be consistent with what we use in the markdown syntax, as we use image names to link the image to the document. Please use some identifiable image names which would be easier for you to link to the document in your other PR. 

Edit: You have added a good commit message. But, this time around the `Change-type` footer is missing in your commit which makes versionbot very unhappy (Tests are not passing below) but no worries about that. Most new contributors do miss that, I let them focus on the issue and then go in and add the footer myself using git rebase. Much faster that way. 

Reference: For this scenario, we can use `git commit --amend` if you like to edit and fix the commit message yourself https://docs.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

alisondavis17:
Closing in favor of https://github.com/balena-io/docs/pull/1600

PR #1660, opened 2021-03-19
Make it clearer that "balena ssh" requires adding an SSH key to balenaCloud.

cmfcruz:
Also, link to the procedure on how to add an SSH key to the account.

Change-type: patch
Signed-off-by: Carlo Miguel Cruz <carloc@balena.io>


cmfcruz:
@balena-ci I self-certify!

PR #1670, opened 2021-03-19
Removed Supervisor Upgrade Policy link

rmorillo24:
Change-type: patch
Signed-off-by: Rafael Morillo <rafael@balena.io>


pdcastro (approved):
I've noticed that this PR is not linked to any issue. Mentioning this just in case an issue exists.

20k-ultra (changes requested):
I (Miguel) said we can remove this but @xginn8 thinks we should leave it so more discussion is needed.

cmfcruz:
@balena-ci rebase

cmfcruz:
@balena-ci rebase

vipulgupta2048:
@20k-ultra Should we close this PR or what should we do?

20k-ultra:
@vipulgupta2048 I'm going to make a PR where this upgrade policy remains but it will be in this repository rather then the Supervisor so no need to download and I think upgrade policy is not something the Supervisor repository cares about. It's just saying we can do it if there's a bug.

PR #1671, opened 2021-03-19
Update proxy documentation section

cywang117:
Section should now reflect that the preferred method for enabling
redsocks is through the related Supervisor endpoints. Links have
been added. If the user prefers to not use the endpoints, the docs
for adding redsocks without touching the Supervisor API have also
been clarified.

Change-type: patch
Signed-off-by: Christina Wang <christina@balena.io>

### DIFF
--- a/shared/network/2.x.md
+++ b/shared/network/2.x.md
@@ -377,31 +377,36 @@ __Note:__ {{>"general/country-firewall"}}
 
 ## Connecting Behind a Proxy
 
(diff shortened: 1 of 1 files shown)


cywang117:
@balena-ci rebase

cywang117:
@balena-ci I self-certify!

20k-ultra:
How did you find out all the info for this PR @cywang117. It contains a lot more than I had expected, nice!!

cywang117:
@20k-ultra Thanks! Most of it was already there (meaning the suggested format when manually adding a .conf file), I added the parts that said we prefer to edit proxy configs through the Supervisor endpoint, and updated the recommended .conf fields to reflect what we set as unchangeable fields [here](https://github.com/balena-os/balena-supervisor/blob/master/src/host-config.ts#L19)

Issue #1674, opened 2021-03-19
Remove note that the user can disable delta updates


rmorillo24:
https://www.balena.io/docs/learn/deploy/delta/#enabling-delta-updates

Modify the note, or remove it completely, since the user can no longer disable delta updates.

PR #1681, opened 2021-03-19
Disable the note about disabling deltas

pranasziaukas:
User can no longer disable delta updates.

Change-type: patch
Connects-to: #1674
Signed-off-by: Pranas Ziaukas <pranas@balena.io>

### DIFF
--- a/pages/learn/deploy/delta.md
+++ b/pages/learn/deploy/delta.md
@@ -15,8 +15,6 @@ These binary deltas save on the amount of data needed to be downloaded, reduce t
 
 ## Enabling delta updates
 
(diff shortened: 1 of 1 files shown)


vipulgupta2048 (approved):

hgaiser:
The text below the removed text would indicate users can disable this behaviour.

I'm referring to "`the delta update behavior is enabled by default`". I would then guess that, given that it is enabled by default, I can disable it too.

vipulgupta2048:
Indeed, re-reading that seems to be true @hgaiser Here's a PR to rectify that https://github.com/balena-io/docs/pull/2140

PR #1700, opened 2021-03-19
Parallelize external docs fetching, dropping from 8s to 3s locally

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #1701, opened 2021-03-19
Create jetson-xavier-nx-devkit.md

dtischler:
Added Jetson Xavier NX SD-Card Device Type for Getting Started Guide, re-using the Jetson Nano as a starting point.

Change-type: patch


vipulgupta2048 (commented):
@dtischler Are you planning to add more to this PR? Though I would recommend closing this PR as we move towards automating the Getitng Guides instead.

dtischler:
Let's close this, I agree, it sounds like the process is going to change / improve.

Issue #1619, opened 2021-03-15
Incorrect edit source link for balena-cli docs


vipulgupta2048:
The "Improve this Document" button on https://www.balena.io/docs/reference/balena-cli/ links to the wrong file as the CLI documentation is auto-generated through files. 

@pdcastro How about we link it to this heading instead - https://github.com/balena-io/balena-cli/blob/master/CONTRIBUTING.md#editing-documentation-files-readme-install-reference-website

I mean, if we can't give them the direct file to edit. The next best thing is instructions to work with CLI docs, right?

![Screenshot from 2021-03-16 12-44-30](https://user-images.githubusercontent.com/22801822/111270064-6634a380-8655-11eb-83cd-a9cd9c1223b9.png)

pdcastro:
> The next best thing is instructions to work with CLI docs, right?

Right. 👍  I guess that most people will give up updating the CLI docs when see those instructions, 😞   but that's how it is at the moment.

vipulgupta2048:
I mean, it's way better than folks opening wrong PR's and leading to time wasted in fixing them both by contributor and maintainer.

PR #1702, opened 2021-03-19
Fix broken link for contributing to CLI docs

pipex:
Change-type: patch
Connects-to: #1619


pipex:
@balena-ci I self certify!

PR #1705, opened 2021-03-19
Update base images docs

nghiant2710:
Update stack versions to match with latest changes in contracts

Change-type: patch
Signed-off-by: Trong Nghia Nguyen <nghiant2710@gmail.com>

PR #1706, opened 2021-03-19
Etcher Pro SVG logo

mpous:
Change-type: patch


dfunckt:
@mpous where is that image used?

dfunckt:
#1705 is related.

vipulgupta2048 (commented):
@mpous As per the discussion we had on the release party, we will be looking to automate these Getting Started guides instead. Please do close this if there is nothing more to add to this. Considering you also have other PR's open for adding the etcher pro getting started.

PR #1710, opened 2021-03-19
Correct Etcher Pro svg logo

mpous:


vipulgupta2048 (commented):
@mpous As per the discussion we had on the release party, we will be looking to automate these Getting Started guides instead. Please do close this if there is nothing more to add to this. Considering you have other PR's open as well.

PR #1712, opened 2021-03-19
Adding Etcher Pro on devices.json

mpous:
Change-type: patch

### DIFF
--- a/config/dictionaries/device.json
+++ b/config/dictionaries/device.json
@@ -1318,6 +1318,47 @@
       }
     ]
   },
+ {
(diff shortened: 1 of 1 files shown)


vipulgupta2048 (commented):
@mpous As per the discussion we had on the release party, we will be looking to automate these Getting Started guides instead. Please do close this if there is nothing more to add to this. Considering you also have #1710 #1706 open as well. If you plan to go on, please finish the pull request with all the required images and partials required for a getting started guide.

PR #1715, opened 2021-03-19
Changing app types when upgrading plan

alisondavis17:
Change-type: minor
Signed-off-by: Alison Davis Riddell alison@balena.io


dfunckt (approved):
Common gotcha, good stuff 👍

Issue #758, opened 2018-03-20
Emphasize that for production base images should be version-pinned


brownjohnf:
I think we should make it clear in the docs that it's best practice to version-pin your base image, to ensure you don't accidentally pull in breaking changes.

At a minimum we should explain the concepts and the risks/benefits of pinning vs. not pinning.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_a2h7)

garethtdavies:
This is covered in #1344

vipulgupta2048:
Could be part of #1617

vipulgupta2048:
Closing in favor of #1720

PR #1720, opened 2021-03-19
Emphasized recommendation to pin base images on production.

ntzovanis:
Previously the recommendation was hidden in an example and wasn't very visible.

Connects-to: #758
Change-type: patch
Signed-off-by: Nicolas Tzovanis <nicolas@balena.io>


dfunckt (approved):
Nice one!

Issue #1543, opened 2020-12-11
Update device statuses


xginn8:
https://www.balena.io/docs/learn/manage/device-statuses/ is out of date and not yet updated to the combined statuses we now use.

karaxuna:
@thgreasi I see that combined statuses are also explained in docs. Shall we close this?

thgreasi:
I don't think the docs have been updated yet @karaxuna . I still see the separate vpn-heartbeat status icons.
![image](https://user-images.githubusercontent.com/1295829/111784910-9cc42580-88c4-11eb-9e72-ca839f5b51ac.png)

vipulgupta2048:
This issue required more investigation and context, hence moved it to intermediate.

PR #1721, opened 2021-03-19
Update the device statuses

thgreasi:
Resolves: #1543
Change-type: patch


pages/learn/manage/device-statuses.md
```diff
@@ -11,20 +11,21 @@ The application overview page shows the status of each device in the application
...
+| Configuring       | The device is applying OS configuration.                                                                                                                        |
+| Updating          | The device is updating to a new application release.                                                                                                            |
+| Offline           | The device is offline and is not connected to VPN and has not any recent API communications.                                                                    |
+| Post Provisioning | The device has been [provisioned][device-provisioning] but has not yet come online.                                                                             |
+| Inactive          | The device has been [deactivated][deactivated] or has been [preregistered][preregistered] but has not yet connected to the {{ $names.cloud.lower }} API.        |
+| Frozen            | The device has been frozen because it's outside the organization's allowance, or is in a paid [application type][application type] on a free tier organization. |
```
dfunckt, line 22: It would be great to link to a pagee that explains more about "frozen" devices, if one exists.

pages/learn/manage/device-statuses.md
```diff
@@ -11,20 +11,21 @@ The application overview page shows the status of each device in the application
...
 
 ## Device connectivity
 
 ![Device connectivity indicators](/img/common/main_dashboard/device_status.png)
 
-In addition to the device status, there are indicators to show if the device is currently connected to the VPN (checkmark icon) and has recent API communications (heartbeat icon). These indicators can be used to identify common issues, such as a firewall blocking VPN traffic or an inability of the device to communicate with the {{ $names.cloud.lower }} API.
+In addition to the device status, there are indicators to show if the device is having any partial connectivity issue. These indicators can be used to identify common issues, such as a firewall blocking VPN traffic or an inability of the device to communicate with the {{ $names.cloud.lower }} API.
```
dfunckt, line 28: I think the previous version that explained each condition separately is more informative. What do I miss?

thgreasi: I essentially just dropped the references to the two separate vpn/heartbeat status icons that we no longer have.
Now the secondary status indicator explains what connectivity issue there is (as seen in the new image).

Perhaps this could be further improved by changing the example below to enumerate all possible cases, but I think doesn't need to be part of this PR.

pages/learn/manage/device-statuses.md
thgreasi: Seems that atm we don't have such a page about device freezing. We definitely should have one, but it's also weird what it should say, given the nature of the current implementation.

pages/learn/manage/device-statuses.md
dfunckt: I didn’t realise we changed that and also missed the image! Thanks for clarifying.

dfunckt (approved):

vipulgupta2048:
@balena-ci rebase

PR #1730, opened 2021-03-19
Remove old resin dashboard screenshot

dtischler:
Remove old resin dashboard screenshot from Getting Started Guide

Change-type: patch


dtischler:
@balena-ci I self-certify!

PR #1742, opened 2021-03-19
Stop generating lunr index as we no longer use it

Page-:
This should significantly speed up builds

Change-type: patch


Page-:
@balena-ci rebase

Page-:
@balena-ci retest

Page-:
@balena-ci I self-certify!

PR #1759, opened 2021-03-19
Fix project demo URL

andrewnhem:
Fixed URL to user's project. They had changed domain and didn't update us.

Change-type: patch
Signed-off-by: Andrew Nhem <andrew@balena.io>


andrewnhem:
@balena-ci I self-certify!

PR #1772, opened 2021-03-25
Clarify section about device env vars and bandwidth usage

nazrhom:
Change-type: patch


nazrhom:
@balena-ci rebase

nazrhom:
@balena-ci I self-certify!

Issue #1734, opened 2021-03-19
Typos and corrections needed on the "Deploy to your fleet" page


LucianBuzzo:
- "will be executed on a x86_64 CPU" -> "will be executed on **an** x86_64 CPU"
- "not be visible in release logs page." -> "not be visible **on the** release logs page."
- "Docker layer cache and build from scratch" -> "Docker layer cache and build**s** from scratch"
- "the build system will endeavor to build" -> "the build system will **try** to build"
- "application which has its default device type selected as Raspberry Pi 3" -> "application**that** has its default device type **set to** Raspberry Pi 3"
- "define build variables which will be added" -> "define build variables **that** will be added"
- "flag is used, the build is built on an x86_64 machines with qemu" -> "flag is used, the build is built on an x86_64 **machine** with qemu"

PR #1780, opened 2021-03-29
Fix Typos on multiple pages

apexider:
Change-type: minor
Signed-off-by: alexandra <alexandra@balena.io>

### DIFF
--- a/pages/learn/deploy/deployment.md
+++ b/pages/learn/deploy/deployment.md
@@ -33,7 +33,7 @@ The `--source` flag allows you do define a path to your source code folder. This
 
 #### `--emulated, -e`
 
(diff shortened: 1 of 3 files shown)


apexider:
In [#1734](https://github.com/balena-io/docs/issues/1734) was one typo, that was part of this file which is why I fixed it here:

"define build variables which will be added" -> "define build variables that will be added"

vipulgupta2048:
@balena-ci rebase

vipulgupta2048 (approved):
LGTM, Thanks for opening the PR @apexider I have transferred your commits from other PR's to create one single PR to fix all typos. Renamed your commits to signify which typo you fixed and assigned the `Change-type` as patch.

Issue #591, opened 2017-09-15
Warn users do not set DBUS_SYSTEM_BUS_ADDRESS containerwide


imrehg:
Users should use `DBUS_SYSTEM_BUS_ADDRESS` only with the commands they want to direct onto the hostOS. If set this value for the whole container, things might/likely break.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_2ech)

PR #1782, opened 2021-03-29
Add warning on DBUS command usage

gelbal:
Connects to https://github.com/balena-io/docs/issues/591
Change-type: minor
Signed-off-by: Fırat Gelbal firat@balena.io

### DIFF
--- a/pages/learn/develop/runtime.md
+++ b/pages/learn/develop/runtime.md
@@ -80,6 +80,8 @@ services:
       io.balena.features.dbus: '1'
 ```
 
+__Note:__ Please be aware that setting `DBUS_SYSTEM_BUS_ADDRESS` as a service or environment variable and enabling systemd at the same time might introduce unexpected side effects. Systemd might start to interact with the host system instead of the container. These interactions can potentially cause balenaOS devices to disconnect from balenaCloud or even fall offline. Hence, users are advised to prefer prepending the command with the variable definition.
+
 __Note:__ To use the `dbus-send` command in the example you will need to install the `dbus` package in your Dockerfile if you are using the Debian image, or check under what name does your chosen operating system supply the `dbus-send` executable.
 
 #### Change the Device hostname


pages/learn/develop/runtime.md
```diff
@@ -80,6 +80,8 @@ services:
       io.balena.features.dbus: '1'
 ```
 
+__Note:__ Please be aware that setting `DBUS_SYSTEM_BUS_ADDRESS` as a service or environment variable might introduce unexpected side effects. So we advise our users to prefer the prepending method.
```
vipulgupta2048, line 83: @gelbal can you expand on the side-effects part? Just a bit. Maybe the support thread linked to the issue can tell us more regarding the same. I am saying because sometimes we use it and I haven't seen any ill-effects yet.

gelbal: Good call @vipulgupta2048. I elaborated based on what I read in the linked support thread. Please take another look.

pages/learn/develop/runtime.md
```diff
@@ -80,6 +80,8 @@ services:
       io.balena.features.dbus: '1'
 ```
 
+__Note:__ Please be aware that setting `DBUS_SYSTEM_BUS_ADDRESS` as a service or environment variable and enabling systemd at the same time might introduce unexpected side effects. systemd might interact with the host system instead of the container. These interactions can potentially cause balenaOS devices to disconnect from balenaCloud or even fall offline. So we advise our users to prefer the prepending method.
```
vipulgupta2048, line 83:
```suggestion
__Note:__ Please be aware that setting `DBUS_SYSTEM_BUS_ADDRESS` as a service or environment variable and enabling `systemd` at the same time might introduce unexpected side effects. `Systemd` might start to interact with the host system instead of the container. These interactions can potentially cause balenaOS devices to disconnect from balenaCloud or even go offline. Hence, users are advised to use the `io.balena.features.dbus` label instead.
```
Always being 3rd person and keeping things specific helps a ton in the docs.

pages/learn/develop/runtime.md
vipulgupta2048: Thanks for the changes @gelbal TIL for me with these

pages/learn/develop/runtime.md
gelbal: Funny that you suggest these @vipulgupta2048 as I also thought about these working on this PR. I didn't commit as you suggested initially for consistency reasons. Right now the docs always refer it as systemd and not `systemd`. Similarly, I saw mixed usage on how we refer to the users. 

I'm fine with committing your suggestion but I wanted to highlight the potential inconsistency we create, especially with `systemd` reference.

gelbal: On another note
> users are advised to use the `io.balena.features.dbus` label instead.

Not sure if I follow. How do you reason this? 

Admittedly, I don't know much about DBUS but in the linked support threads the agents kept re-iterating: "User was advised not to set DBUS_SYSTEM_BUS_ADDRESS as a container-wide setting and instead to put the variable definition before executing the process on the same line".

The way I see the `io.balena.features.dbus` label needs to be set regardless. It's more about warning the users on the potential side effects of exporting `DBUS_SYSTEM_BUS_ADDRESS` as an environment variable.

vipulgupta2048: > I'm fine with committing your suggestion but I wanted to highlight the potential inconsistency we create, especially with systemd reference.

I agree, consistency is important. If that's something we follow all over. We shouldn't change it.

vipulgupta2048: > The way I see the io.balena.features.dbus label needs to be set regardless. It's more about warning the users on the potential side effects of exporting DBUS_SYSTEM_BUS_ADDRESS as an environment variable.

Right on. I made this change in regards to this sentence here where I think being specific to what that pretending method is makes more sense. 

> So we advise our users to prefer the prepending method.

vipulgupta2048: We can replace what I suggested with what exactly the prepending method is that the customers should use.

gelbal: Cool! I pushed a new change that combines all we discussed so far Vipul. Please take another look.

vipulgupta2048 (commented):
LGTM, but I would still suggest changing it to Systemd (with a capital S) because it's grammatically correct, and since docs follow English conventions first. The small `s` on systemd just irks my brain tbh 😂

gelbal:
@vipulgupta2048, as I expected this uppercase question come up, I checked [systemd's website](https://www.freedesktop.org/wiki/Software/systemd/) and it says:
> The only situation where we find it OK to use an uppercase letter in the name (but don't like it either) is if you start a sentence with systemd.

I like you and your suggestions dude. Then I'd like to respect the choices of this free software's maintainers :smile: 

Having said these, I see that Wikipedia uses uppercase on the sentence starts. So it's fine to go with it. I wanted to highlight this for the chuckles : )

vipulgupta2048:
Yeah @gelbal it all comes down to mere semantics of the language, especially in docs. This is because when we are writing code the language has its own rules such as indentation, variable names, and how you write statements so it can be understood by the compiler. Same as when you write docs, one has to play by English rules first and then figure out code conventions alongside it. 

I mentioned the suggestion again in the end because I feel this will irk someone sometime later when they read it. And, a PR would then be needed to capitalize the S of systemd.

gelbal:
Sure. That's fine @vipulgupta2048. I think the best solution here is to type all the references across docs as `systemd` and not worry about uppercase. Such change goes beyond this PR's purpose. Anyway, I changed that one reference to start with uppercase now. Seems like we are good to merge this, right?

vipulgupta2048 (approved):

PR #1787, opened 2021-03-30
OS:meta-balena.md: Update device contribution guide

floion:
Change-type: patch
Signed-off-by: Florin Sarbu <florin@balena.io>


vipulgupta2048 (commented):
@floion Odd to see most things defaulting back to resin and not staying balena. Is there a reason behind that or just how the current process looks like and is bound to change later. Otherwise PR wise LGTM

floion:
> @floion Odd to see most things defaulting back to resin and not staying balena. Is there a reason behind that or just how the current process looks like and is bound to change later. Otherwise PR wise LGTM
> Do rebase and self-certify whenever you see this

Some of the renames should have not been done in the first place. I think they were done thinking all code was changed from resin to balena but that's not yet the case in meta-balena.

floion:
@balena-ci I self-certify!

Issue #1620, opened 2021-03-16
Link to Supervisor Upgrade Policy is broken


cmfcruz:
Affected page: https://www.balena.io/docs/reference/supervisor/supervisor-upgrades/

The offending link is located at the bottom of the page. I'm not sure what page this link should go to.

> For more information on specific limitations when upgrading the balena Supervisor, see the [upgrade policy](https://www.balena.io/docs/reference/supervisor/upgrade-policy).

vipulgupta2048:
Hey @20k-ultra in your commit here https://github.com/balena-io/docs/commit/0d04ec83e3e9817673c067d1ae576e42c26418a5 The `upgrade policy` download step was removed from the `fetch-external.sh` script. This is the reason why the link to file @cmfcruz mentioned is going 404.

Was there a specific reason for that or are we no longer providing a specific upgrade policy for users to follow? Or is it because the [self-service upgrades](https://www.balena.io/docs/reference/supervisor/supervisor-upgrades/) already cover that part and hence the policy is no longer needed. Let me know and I can get an appropriate fix in place, thanks!

Sub issue: Maybe we can check for broken links on each PR merge, 404 links are not all fun for anyone.

20k-ultra:
I see. If we can add that 404 linter it would have caught this but I definitely see that I failed to find all the places this file would have been used...

To fix this I would just remove https://github.com/balena-io/docs/blame/master/pages/reference/supervisor/supervisor-upgrades.md#L29-L31

toochevere:
this issue was closed by the PR.  Not sure why the issue didn't close automatically.

20k-ultra:
@toochevere I closed that PR. It wasn't merged.

I'll be making a PR to fix this today.

PR #1801, opened 2021-03-31
Move Supervisor upgrade policy to self-serve document

20k-ultra:
Change-type: patch
Closes: #1620
Signed-off-by: Miguel Casqueira <miguel@balena.io>


20k-ultra:
@balena-ci I self-certify!

Issue #1745, opened 2021-03-19
Broken link in table of supported machine names and archs


Jazzagi:
Docs page where broken link appears: https://www.balena.io/docs/learn/develop/dockerfile/

Sentence containing broken link:   Table content : 
Nitrogen8MM DWE | nitrogen8mm-dwe | aarch64 | Link

Broken url: https://github.com/balena-io-library/base-images/tree/master/balena-base-images/device-base/nitrogen8mm-dwe

mpous:
This image doesn't exist in the base images [repo](https://github.com/balena-io-library/base-images/tree/master/balena-base-images/device-base)

somombo:
Commit #[703e75f](https://github.com/balena-io-library/base-images/tree/703e75f7b3dc97af85aee690ac460c0469d42e39/balena-base-images/device-base/nitrogen8mm-dwe) is the last one that contained this image's directory. I cannot determine what happened to cause it to suddently disappear

vipulgupta2048:
Maybe @nghiant2710 can help here to understand what happened to that base image?

PR #1806, opened 2021-04-01
deviceTypeNames.md: Remove Nitrogen8MM DWE

AlidaOdendaal:
The Nitrogen8MM DWE device has been converted to a private
device type (https://www.flowdock.com/app/rulemotion/resin-devices/threads/UekZJBCxvruGdw5ceyVEwPU_z5d).
Thus it is better to remove it from the public docs.

Change-type: patch
Signed-off-by: AlidaOdendaal <alida@balena.io>
Connects-to: #1745


vipulgupta2048:
Right on @AlidaOdendaal When devices go private, we don't need to remove the Getting Started guides I suppose, or is there any specific process at play?

AlidaOdendaal:
@acostach See Vipul's question above. What do you think? I guess the device type can remain in the "Getting Started" guide, for the sake of the private device owners? Or do they have access to other documentation as well?

AlidaOdendaal:
@balena-ci I self-certify!

PR #1815, opened 2021-04-05
Add an example getting IPv4 addresses from NetworkManager via dbus

mtoman:


shared/network/2.x.md
```diff
@@ -362,6 +362,55 @@ settings.AddConnection(con)
...
+        addresses.append("%s/%s" % (addr.compressed, address[1]))
+
+    # Get device name
+    device_name = device_props.Get("org.freedesktop.NetworkManager.Device", "Interface")
+
+    # Print all the information for a single device
+    print("%s: %s" % (device_name, ", ".join(addresses)))
+```
+
+The above example will print device name and all associated IPv4 addresses for each network device with an IPv4 connection.
```
vipulgupta2048, line 412: The above line might be better off as the one being before the code snippet to clearly explain what the code snippet is all about. This way folks know what they are looking at before scrolling further down and looking for explanations.

vipulgupta2048: I would say the same for the example above it, as having a description before the example provides much needed context to the code in the example.

vipulgupta2048 (commented):
@mtoman reviews are inline, otherwise LGTM. Merge when ready.

mtoman:
@balena-ci rebase

shared/network/2.x.md
mtoman: Moved both descriptions above

mtoman:
I've been encouraged to merge, therefore

mtoman:
@balena-ci I self-certify!

PR #1824, opened 2021-04-08
patch: update text in Creating teams

apostolism:


apostolism:
@thgreasi would you like to have a look at the proposed changes?

thgreasi (approved):
Nice catch @apostolism

PR #1857, opened 2021-04-21
patch: Updated the list of domains that need to be whitelisted.

pranavpeshwe:
In order for any device to be reachable via web-terminal or `balena ssh`, being able to reach balenaCloud VPN servers is indispensable.
Our VPN servers run on EC2 instances on AWS. It is safer & scalable to whitelist the entire amazonaws.com domain rather than whitelist individual EC2 instances or regions.


lurch:
I guess it's also easy for am "attacker" to get a domain in amazonaws.com ?  I have zero context for this change though, so what I've just said may be 100% irrelevant :laughing:

Issue #1807, opened 2021-04-01
Remove Zynq zc702 from supported devices in docs, since it is now a private device type.


AlidaOdendaal:
Since it is currently only supported under private device types (https://www.flowdock.com/app/rulemotion/resin-devices/threads/UekZJBCxvruGdw5ceyVEwPU_z5d), it should not appear in the public-facing table [here](https://www.balena.io/docs/learn/develop/dockerfile/#dockerfile-templates) with supported devices.

PR #1580, opened 2021-02-04
Update supported SBCs page

rahul-thakoor:
- use /v6/device_type to query supported devices
- use /v6/application to find archived OS

Change-type: minor
Signed-off-by: Rahul Thakoor <rahul@balena.io>

PR #1875, opened 2021-04-28
Remove "Zynq zc702" from publicly supported devices

pranasziaukas:
It is currently only supported under private device types.

Change-type: patch
Connects-to: #1807
Signed-off-by: Pranas Ziaukas <pranas@balena.io>

### DIFF
--- a/shared/general/deviceTypeNames.md
+++ b/shared/general/deviceTypeNames.md
@@ -73,4 +73,3 @@
 | Variscite DART-MX8M Mini        | imx8mm-var-dart                           | aarch64                           | [Link]({{ $links.githubBaseImages }}/imx8mm-var-dart)       |
 | Variscite VAR-SOM-MX6           | var-som-mx6                               | armv7hf                           | [Link]({{ $links.githubBaseImages }}/var-som-mx6)           |
 | Variscite VAR-SOM-MX7           | imx7-var-som                              | armv7hf                           | [Link]({{ $links.githubBaseImages }}/imx7-var-som)          |
-| Zynq zc702                      | zc702-zynq7                               | armv7hf                           | [Link]({{ $links.githubBaseImages }}/zc702-zynq7)           |


vipulgupta2048:
Hey @pranasziaukas Are there more instances of this device type in the docs that you were able to find? 
Also, this might not be the entire solution to this page. We need to automate and generate this page using the /v6/device_type endpoint. Let me know if you are up for a bit more work to keep this page automated. 

Reference: [Supported device types](https://www.balena.io/docs/reference/hardware/devices) page --> https://github.com/balena-io/docs/pull/1580/files

vipulgupta2048:
@pranasziaukas Let me know your thoughts or we can merge this PR. Keep things moving?

pranasziaukas:
Thanks for the feedback @vipulgupta2048, and sorry for the delayed response.

Not sure I'm 100% following the flow of thoughts here. If I understand correctly, relying on the `/v6/device_type` endpoint would be a significant change and pretty much proper frontend work that requires wider understanding than a casual docs fix carried out by an ordinary balenista.

I followed the footsteps of #1806. Has anything changed since then? Or are we just looking forward to making a leap in the level of automation?

vipulgupta2048:
@pranasziaukas True that, automating a page does require a bit of work. We are looking forward to making that leap, yes. I am not sure of the complete process of removing devices from the docs and we haven't yet heard from Alexandru. Maybe @AlidaOdendaal did and then merged the PR. I think we should confirm again, maybe on FD.

pranasziaukas:
https://www.flowdock.com/app/rulemotion/resin-devices/threads/UekZJBCxvruGdw5ceyVEwPU_z5d <- for context

pranasziaukas:
@balena-ci rebase

vipulgupta2048 (approved):

pranasziaukas:
@balena-ci rebase

PR #1893, opened 2021-05-03
patch: move cpus to unsupported fields

frutti93:
The `cpus` command is not supported, because it was only introduced by docker compose v2.2. Using this leads to a ValidationError `<container-name> should NOT have additional properties`.

Also discussed here: [Link](https://forums.balena.io/t/limiting-cpu-usage-in-multi-container-environment-via-docker-compose/3469)


vipulgupta2048:
Thanks for contributing to the docs @frutti93 I have added folks from the supervisor team to confirm this change. Till then we can work towards prepping the PR. If you can squash your commits and reword the squashed commit to have the `patch:` prefix, that would be the most helpful. Let us know if you have any issues doing the same, we are happy to help. Thanks!

cywang117:
@vipulgupta2048 This isn't a Supervisor issue, but rather just a docs change. [Here](https://github.com/balena-io-modules/resin-compose-parse/blob/master/src/schemas/config_schema_v2.1.json#L107-L109) is the code that validates and parses `docker-compose.yml`. The latest compose schema is indeed v2.1, so `cpus` wouldn't be supported, since as @frutti93 mentions, `cpus` was added in v2.2. For reference the ValidatorError is being thrown from [ajv validator package](https://github.com/ajv-validator/ajv).

@frutti93 Is this blocking you in any way? Have you tried the suggestion in the forum link you mentioned?

frutti93:
@vipulgupta2048 Hi, thanks for the quick response, will do so!

@cywang117 It is not blocking me, I just switched to use `cpu_quota` instead. I just found it confusing, because `cpus` was listed as supported field in the docs and I had made several modifications to my compose file. It took me some time to realize it was in fact the `cpus` option and I thought it should not be listed as supported if it is in fact not. ;)

vipulgupta2048 (approved):
Thanks @cywang117 for the assist and @frutti93 for coming through with this change 🚀

vipulgupta2048:
@balena-ci rebase

PR #1925, opened 2021-06-03
Added dependabot config to format commit messages

20k-ultra:
This is so we can easily merge PRs from dependabot like https://github.com/balena-io/docs/pull/1914 without having to checkout the branch and fix versionbot.

dependabot config options https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10
+  commit-message:
+    prefix: patch
+  rebase-strategy: disabled


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "12:00"
+  open-pull-requests-limit: 10
+  commit-message:
+    prefix: patch
```
20k-ultra, line 10: we might have to modify this to contain a colon after patch but I'm not sure.

```suggestion
    prefix: patch:
```

vipulgupta2048: Ahh, you mean if the colon might mess up with YAML syntax? 
I think we can run with this and see what happens?

20k-ultra: I was thinking about versionbot if it wants the colon or not. But checking the source versionbot actually has prefix: patch without a colon in their dependabot config: https://github.com/product-os/versionbot/blob/e0bbcef1d7b1b06147519b90b51e7fe044c3c1bc/.github/dependabot.yml#L9

20k-ultra:
Never using the github merge button again...

20k-ultra:
@balena-ci retest

20k-ultra:
@balena-ci rebase

20k-ultra:
@balena-ci I self-certify

20k-ultra:
@balena-ci I self-certify!

PR #1928, opened 2021-06-07
add note about responsible disclosure

Hades32:
Before googleing for this topic and Balena didn't yield useful results


vipulgupta2048:
I like this +1 from my side. Ship it.

20k-ultra (approved):

Issue #1881, opened 2021-04-29
Clarify use of <DEVICE ID> in API Documentation


vipulgupta2048:
Folks often get confused with the use of DEVICE_ID in the API. It's a unique property that is visible nowhere on the UI to refer to a device apart from the UUID. Having a definition about DEVICE_ID on the resources page where folks are using it might prove to be helpful. 

For example, at the top of this page https://www.balena.io/docs/reference/api/resources/device/ we can add a definition for DEVICE_ID to mention what it is for folks to refer before executing any queries.

PR #1961, opened 2021-07-02
apiGeneral: calls: clarify internal IDs

jakogut:
Also add a header above the section detailing usage of internal IDs for
permalinking.

Signed-off-by: Joseph Kogut <joseph@balena.io>

A user of the API encountered some confusion over usage of internal IDs. Searching yielded the linked issue already describing the problem, and the need to clarify usage of internal IDs. Reading through the docs yielded a page that mostly already clarifies this, but it's part of a larger page on constructing API calls, and can't be linked to on the section for finding and using resource IDs.

Add an H2 header to enable linking, and add a sentence explaining which field for a returned resource is the internal ID.

Related-to: https://github.com/balena-io/docs/issues/1881

### DIFF
--- a/shared/apiGeneral/calls.md
+++ b/shared/apiGeneral/calls.md
@@ -46,6 +46,8 @@ curl -X GET "{{ $links.apiBase }}/v5/device(<ID>)" \
 -H "Authorization: Bearer <AUTH_TOKEN>"
 ```
 
+## Resource IDs
+
 Many times, however, you won't know the internal ID used by the API, and you'll want to use some other piece of information to find the appropriate resource. In these cases, you can use the `$filter` method to select resources based on any field. For example, if you are looking for a specific device, it's more likely that you'll have the device UUID than the device ID:
 
 ```shell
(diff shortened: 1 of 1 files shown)


vipulgupta2048:
Thanks for taking this up @jakogut Maybe an example would help?

I had it in mind where we can show the output of the /v6/device resource, and specifically pointing out the find DEVICE_ID value. Then, mentioning similar instructions for APPLICATION_ID as well. This clearly defined example would help folks out on support.

jakogut:
>  we can show the output of the /v6/device resource, and specifically pointing out the find DEVICE_ID value

@rahul-thakoor This is added starting at line 62. I truncated the output, as the serialized object is quite lengthy, but I could change this to the complete output using `select` to only retrieve the ID field.

jakogut:
@balena-ci rebase

vipulgupta2048:
@jakogut I would say providing at least some more fields would be nicer

```
{
a : c,
id : {ldfndlf}
a1 : c,
a2 : c,
a3 : c,
a4 : c,
a5 : c,
...
}
```

If having a generalised query or process that would help folks in finding their ID's faster then I am all for it. Do put this under a small heading like h3 or h4 as appropriate. So support agents can use that when they are looking.

vipulgupta2048 (approved):

LucianBuzzo (approved):

PR #1964, opened 2021-07-05
Fix bind9 URL and omit some URLs in link-checker

klutchell:
Resolves: #1963, #1962
Change-type: patch

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -27,9 +27,12 @@ jobs:
           # a result will always produce a failing link.
           # The myorg/myapp link is a dummy address used in an example contract,
           # so is omitted.
+          # The balena/resin API urls will not respond to unauthenticated requests
+          # Accept code 429 as returned by dockerhub rate limiting for our base images
           args: >
             --exclude-mail
             --exclude
+            --accept 429
             localhost
             127.0.0.1
             https://github.com/balena-io-library/base-images
@@ -37,6 +40,8 @@ jobs:
             https://raw.githubusercontent.com/balena-os/meta-balena
             https://github.com/myorg/myapp
             https://index.docker.io/v2/
+            http://api.resin.io/
+            https://api.balena-cloud.com/
             git://github.com/
             --verbose
             --no-progress
--- a/pages/learn/develop/runtime.md
+++ b/pages/learn/develop/runtime.md
@@ -267,7 +267,7 @@ var server = app.listen(80, function () {
 ```
 
 ### Using DNS resolvers in your container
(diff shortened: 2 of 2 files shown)


.github/workflows/link-checker.yml
```diff
@@ -37,6 +40,9 @@ jobs:
             https://raw.githubusercontent.com/balena-os/meta-balena
             https://github.com/myorg/myapp
             https://index.docker.io/v2/
+            http://api.resin.io/
+            https://api.balena-cloud.com/
+            https://hub.docker.com/r/balenalib/
```
vipulgupta2048, line 45: I was thinking to find an option where we can ignore any URL that goes 429 while checking. I don't think putting hub.docker.com might be the best solution. It's at least checking for several thousand URL's successfully and makes sure base images are actually present.

pages/learn/develop/runtime.md
```diff
@@ -267,7 +267,7 @@ var server = app.listen(80, function () {
 ```
 
 ### Using DNS resolvers in your container
-In the {{ $names.company.lower }} host OS [dnsmasq][dnsmasq-link] is used to manage DNS since {{ $names.os.lower }} 1.1.2. This means that if you have dnsmasq or other DNS resolvers such as [bind9](http://www.bind9.org/) running in your container, it can potentially cause problems because they usually try to bind to `0.0.0.0` which interferes with the host dnsmasq. To get around this you need to add `bind-interfaces` to your dnsmasq configuration in your container, or make sure your server only binds to external IPs, and there shouldn't be conflicts anymore.
+In the {{ $names.company.lower }} host OS [dnsmasq][dnsmasq-link] is used to manage DNS since {{ $names.os.lower }} 1.1.2. This means that if you have dnsmasq or other DNS resolvers such as [bind9](http://bind9.net/) running in your container, it can potentially cause problems because they usually try to bind to `0.0.0.0` which interferes with the host dnsmasq. To get around this you need to add `bind-interfaces` to your dnsmasq configuration in your container, or make sure your server only binds to external IPs, and there shouldn't be conflicts anymore.
```
vipulgupta2048, line 270:
```suggestion
In the {{ $names.company.lower }} host OS [dnsmasq][dnsmasq-link] is used to manage DNS since {{ $names.os.lower }} 1.1.2. This means that if you have dnsmasq or other DNS resolvers such as [bind9](https://bind9.net/) running in your container, it can potentially cause problems because they usually try to bind to `0.0.0.0`, which interferes with the host dnsmasq. To get around this, you need to add `bind-interfaces` to your dnsmasq configuration in your container or make sure your server only binds to external IPs, and there shouldn't be conflicts anymore.
```

We should get all URL's HTTPS as we move ahead.

vipulgupta2048 (commented):
Looking good.

vipulgupta2048 (approved):

vipulgupta2048:
@balena-ci rebase

PR #1974, opened 2021-07-08
Explain why balena's restart action is different than Docker restart

cywang117:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/r-product/threads/b0S9epziNSpbShwcrTI4zG3-Dwq
Signed-off-by: Christina Wang <christina@balena.io>

image

### DIFF
--- a/pages/learn/manage/actions.md
+++ b/pages/learn/manage/actions.md
@@ -32,7 +32,15 @@ You may also enable or disable public device URLs by clicking the _Public device
 
 The `Restart Application` action restarts the currently running **application containers**. Your application's running containers will be removed and recreated from scratch. This behavior is intended, and is different from running `balena restart [OPTIONS] CONTAINER [CONTAINER...]` in a host OS terminal instance from your dashboard, which will not remove your containers. If you are trying to persist data between container removals, see [persistent storage][persistent-storage] for strategies.
 
-When the containers are stopped, the application is politely asked to stop by sending a `SIGTERM` and after 10 seconds of wait time a `SIGKILL` is sent.
+By removing containers and recreating them from scratch, we see benefits like the following:
+
+- Containers are meant to be ephemeral, meaning that a new container should be a drop-in replacement for an old container with minimal to no impact. Removing and recreating containers adheres to this philosophy.
+
+- Because containers are removed and recreated with the restart action, you're encouraged to follow best practices in Docker data persistence. For more information, see our [persistent storage][persistent-storage] documentation or [Docker's data persistence strategies][docker-data-persistence-strategies]. These strategies also offer a performance boost over storing files in the container's writable layer.
+
+- Removing and recreating containers may allow recovery from application bugs where the application is stuck in an invalid state. For example, a process ID file that is no longer valid but is persisted to the container filesystem would be cleaned up when recreating the container.
+
+When the containers are stopped, the application is politely asked to stop by sending a `SIGTERM`. If the application hasn't stopped after 10 seconds, a `SIGKILL` is sent.
 
 __Note:__ During a restart any data that is not stored in `/data` will be lost.
 
@@ -180,3 +188,4 @@ __Warning:__ It is a good idea to [move your devices to another application][mov
 [support-access]:/learn/manage/support-access
 [inactive-devices]:/learn/manage/billing/#inactive-devices
 [application-members]:/learn/manage/account/#application-members
+[docker-data-persistence-strategies]:https://docs.docker.com/storage/


pages/learn/manage/actions.md
```diff
@@ -32,14 +32,21 @@ You may also enable or disable public device URLs by clicking the _Public device
 
 The `Restart Application` action restarts the currently running **application containers**. Your application's running containers will be removed and recreated from scratch. This behavior is intended, and is different from running `balena restart [OPTIONS] CONTAINER [CONTAINER...]` in a host OS terminal instance from your dashboard, which will not remove your containers. If you are trying to persist data between container removals, see [persistent storage][persistent-storage] for strategies.
 
-When the containers are stopped, the application is politely asked to stop by sending a `SIGTERM` and after 10 seconds of wait time a `SIGKILL` is sent.
+When the containers are stopped, the application is politely asked to stop by sending a `SIGTERM`. After 10 seconds of wait time, a `SIGKILL` is sent.
```
klutchell, line 35:
```suggestion
When the containers are stopped, the application is politely asked to stop by sending a `SIGTERM`. If the application hasn't stopped after 10 seconds a `SIGKILL` is sent.
```

pages/learn/manage/actions.md
```diff
@@ -32,14 +32,21 @@ You may also enable or disable public device URLs by clicking the _Public device
...
 
 __Note:__ During a restart any data that is not stored in `/data` will be lost.
 
 It should be noted that currently these action notifications are not queued up, so if a device is offline when the action is triggered, it will never be notified of it.
 
 __Warning:__ Restart device container is not equivalent to a reboot of the device!
 
+Here is {{ $names.company.lower }}'s rationale for the restart action being different than Docker's restart action:
```
klutchell, line 43:
```suggestion
By removing containers and recreating them from scratch we see benefits like the following.
```

klutchell: Can we move this entire block of reasons up to line 34 so it follows the paragraph about it being intended behaviour?

pages/learn/manage/actions.md
lurch: Does it actually try to detect if the application has stopped, or does it just send the SIGKILL regardless?

cywang117: Good point & nice catch!. I rephrased this bit without thinking about the functionality underneath.

cywang117: @lurch SIGKILL would only be sent if the container(s) is/are still running, since the command is much like Docker's `docker stop` command.

pages/learn/manage/actions.md
cywang117: Sounds good.

cywang117: I was trying to think of a better way to phrase the `Here is balena's rationale...` line but couldn't think of anything. Thanks!

pages/learn/manage/actions.md
```diff
@@ -32,7 +32,15 @@ You may also enable or disable public device URLs by clicking the _Public device
...
 
-When the containers are stopped, the application is politely asked to stop by sending a `SIGTERM` and after 10 seconds of wait time a `SIGKILL` is sent.
+By removing containers and recreating them from scratch, we see benefits like the following:
+
+- Containers are meant to be ephemeral, meaning that a new container should be a drop-in replacement for an old container with minimal to no impact. Removing and recreating containers adheres to this philosophy.
+
+- Because containers are removed and recreated with the restart action, you're encouraged to follow best practices in Docker data persistence. For more information, see our [persistent storage][persistent-storage] documentation or [Docker's data persistence strategies][docker-data-persistence-strategies]. These strategies also offer a performance boost over storing files in the container's writable layer.
+
+- Removing and recreating containers is useful in a scenario where recreating a container allows recovery from application bugs where the application got stuck in an invalid state that was persisted to the container's filesystem. In this scenario, data loss is a good thing, as the data consists of a bad state.
```
klutchell, line 41:
```suggestion
- Removing and recreating containers may allow recovery from application bugs where the application is stuck in an invalid state. For example, a process ID file that is no longer valid but is persisted to the container filesystem would be cleaned up when recreating the container.
```
I think mentioning that the "data loss is a good thing" isn't super helpful to the point. Please feel free to disagree with me!

cywang117: No disagreement here 👍🏼

klutchell (approved):
lgtm

Issue #1972, opened 2021-07-08
Prevent link checker from getting rate limited


20k-ultra:
Since the link checker is looking up a lot of URLs at once it is getting rate limited by docker hub but could also get rate limited by github.

Github 1000 request/hour: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting
Docker hub does not say their rate limit: https://docs.docker.com/docker-hub/download-rate-limit/#other-limits

A really simple solution is to have the checker run once every hour and before the link checker step we collect all the links in the docs repo and divide them into groups of 1000. Based on the UTC hour we select that hours group and store them into a links file. Now when lychee runs we specify to only check that links file. 

Example: if we have 10,000 links to check, 10,000 / 1000 (max_requests) = 10 groups. At 00:00 we check the first group of 1000. at 01:00 we check the second group...etc. At 11:00 we have no more groups so it will cycle back to the first group (00:00). This would allow us to safely check 24,000 links per github 1000 requests/hour limit. Docker hub might make it lower and we'll only know about experimenting (locally).

Inspired because of https://github.com/balena-io/docs/issues/1962

20k-ultra:
Finding all the links and grouping might cost a lot of time but this task is running every hour in the background...nothing is waiting for it to complete and I'm sure we can complete the operation within the hour.

klutchell:
I'm not sure about this, I would rather have all of the links checked in one run, even if it takes longer as a result.

What if we tried reducing `max-concurrency` and/or `threads` to super low values and see if we can avoid hitting the limits? 

Though I guess that would only work if we were polling one URL every 3.4s or more.

Why not batch the queries as you suggested, but keep them all in the same action with some artificial sleeps in between batches?

klutchell:
Related PR that we may want to partially revert: https://github.com/balena-io/docs/pull/1964

lurch:
> even if it takes longer as a result.

As long as it doesn't exceed the maximum job-time of whatever CI system you're using :wink:

20k-ultra:
Github actions are limited to 6 hours so the sleep strategy limits us to 6,000 links per execution and if we have 6,001 links how do we get that 6000-N remainder ? We'll have to batch anyways. 

[...]

PR #1977, opened 2021-07-12
Resume checking for rate-limit return codes

klutchell:
We don't want to exclude return code 429 as that may
mask issues with dead links.

Also, the way this flag was added broke the usage of
the --exclude flag by mistake.

See: https://github.com/balena-io/docs/issues/1972
Partially reverts: https://github.com/balena-io/docs/pull/1964

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #2055, opened 2021-08-23
patch: Format the device statuses document

vipulgupta2048:


vipulgupta2048:
@balena-ci I self-certify!

PR #2068, opened 2021-08-31
Remove "Upgrade Policy" link for device Supervisor

20k-ultra:
Removed this link because it's been superseded by https://github.com/balena-io/docs/blob/master/pages/reference/supervisor/supervisor-upgrades.md#upgrade-paths

PR #2071, opened 2021-09-01
Redirect Upgrade Policy links to self-service upgrades

20k-ultra:
Because of https://github.com/balena-io/docs/pull/2068 any links to the upgrade policy will 404 and I think that page is probably linked enough to warrant a redirect which I didn't think of until after.

Redirects https://balena.io/docs/reference/supervisor/upgrade-policy -> https://www.balena.io/docs/reference/supervisor/supervisor-upgrades/


20k-ultra:
@balena-ci I self-certify!

PR #2077, opened 2021-09-11
Fix linux cli install tab

codewithcheese:
Link id was misnamed, causing jquery to fail to show the tab on load


nucleardreamer (approved):

nucleardreamer:
Ah! Nice catch!

PR #2082, opened 2021-09-18
SSH access docs: Replace 'balena ssh' with 'ssh' when used with 'tunnel'

pdcastro:
See: https://jel.ly.fish/support-thread-1-0-0-front-cnv-bc3ywb1
Connects-to: balena-io/balena-cli/issues/2292
Change-type: patch

### DIFF
--- a/shared/general/container-ssh.md
+++ b/shared/general/container-ssh.md
@@ -83,16 +83,14 @@ command maps local port `4321` to remote port `22222` on the device:
 $ balena tunnel <device-uuid> -p 22222:4321
 ```
 
-The device can then be accessed on local port `4321` via the `balena ssh` command or
-standalone SSH client:
+The device can then be accessed on local port `4321` with a standalone SSH client:
 
 ```shell
-$ balena ssh 127.0.0.1 -p 4321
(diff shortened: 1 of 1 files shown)


shared/general/container-ssh.md
```diff
@@ -83,16 +83,14 @@ command maps local port `4321` to remote port `22222` on the device:
 $ balena tunnel <device-uuid> -p 22222:4321
 ```
 
-The device can then be accessed on local port `4321` via the `balena ssh` command or
-standalone SSH client:
+The device can then be accessed on local port `4321` with a standalone SSH client:
 
 ```shell
-$ balena ssh 127.0.0.1 -p 4321
```
pdcastro, line 90: The reason for avoiding `balena ssh` in conjunction with `balena tunnel` is that `balena ssh` currently uses `root` as the username (balena-io/balena-cli/issues/2292), which is fine with dev OS images but not so fine with production OS images as it then requires adding a ssh key to the `config.json` file.

pdcastro:
@balena-ci I self-certify!

thgreasi:
:clap:

PR #2084, opened 2021-09-20
Document 'version' field of balena.yml contract

pdcastro:
Change-type: patch

---

### DIFF
--- a/pages/learn/deploy/deploy-with-balena-button.md
+++ b/pages/learn/deploy/deploy-with-balena-button.md
@@ -87,5 +87,8 @@ version: 3.5.2
   - `applicationConfigVariables` - Configuration variables allow you to provide runtime configuration to the host OS and supervisor. These variables all begin with `BALENA_` or `RESIN_`.
   - `defaultDeviceType` - The device type that will be pre-selected in the "Create fleet" modal. It defaults to Raspberry Pi 4 if not provided. You can find a list of [device types](/reference/hardware/devices/) here.
   - `supportedDeviceType` - The device types that the fleet supports. You can find a list of [device types](/reference/hardware/devices/) here.
+- `version` - Provides a user-defined version value that is shown in the Version column when
+  listing Releases in the balenaCloud web dashboard, and which may also be queried through
+  the balena API.
 
 [balenahub]:{{ $links.balenaHubUrl }}


pages/learn/deploy/deploy-with-balena-button.md
```diff
@@ -87,5 +87,8 @@ version: 3.5.2
   - `applicationConfigVariables` - Configuration variables allow you to provide runtime configuration to the host OS and supervisor. These variables all begin with `BALENA_` or `RESIN_`.
   - `defaultDeviceType` - The device type that will be pre-selected in the "Create fleet" modal. It defaults to Raspberry Pi 4 if not provided. You can find a list of [device types](/reference/hardware/devices/) here.
   - `supportedDeviceType` - The device types that the fleet supports. You can find a list of [device types](/reference/hardware/devices/) here.
+- `version` - Provides a user-defined version value that is shown in the Version column when
+  listing Releases in the balenaCloud web dashboard, and which may also be queried through
+  the balena API.
```
pdcastro, line 92: @thgreasi, users have been asking how to set release versions and support agents share a link to this page. Besides what I've written above, I wanted to add some more details regarding the consequences of following the format `1.2.3` or setting a version value like `foo`. I remember seeing a PR where you (or maybe @toochevere or @pipex ?) added a semver matching regex and  logic such that "if the version matches a semver then set these API fields, otherwise set those other API fields". Do you remember what PR it was, or otherwise what we could document in this regard? I understand it would be technically incorrect to state that the semver _must_ be a value in format `1.2.3` if in fact the platform accepts different values. We could _recommend_ a certain format, but the recommendation would carry extra weight if it was justified, e.g. "we recommend the format `1.2.3` because this allows these extra fields to be set by the balena API and queried in your applications".

toochevere: @thgreasi ... I think you might be the best one to comment on this ^ for the latest about how this works?

thgreasi: Sorry for the delay & thanks for the ping @toochevere.
@pdcastro I would prefer to avoid documenting/mentioning the deprecated `release.release_version` field, since we want to eventually drop it.
The semver spec mentions the `X.Y.Z` part as `version core`, but I think that "a 3 digit X.Y.Z semver" is a more understandable way to mention it.
Maybe instead of that we can say that:
* We plan to add support for all valid semver formats in the future.
* It has to follow the three digit semver format so that balenaCloud can surface the version to the dashboard & handle revisions in case you cut an second release with the same semver.
* Otherwise the version has to be unique across the releases of the whole fleet and will not show in the dashboard and using such formats is deprecated & we plan to stop accepting those in the future.

thgreasi: Here are the PRs that added the ~"semver" regex
See: https://github.com/balena-io/balena-builder/pull/868/files
See: https://github.com/balena-io/balena-cli/pull/2274/files

pages/learn/deploy/deploy-with-balena-button.md
```diff
@@ -87,5 +87,8 @@ version: 3.5.2
   - `applicationConfigVariables` - Configuration variables allow you to provide runtime configuration to the host OS and supervisor. These variables all begin with `BALENA_` or `RESIN_`.
   - `defaultDeviceType` - The device type that will be pre-selected in the "Create fleet" modal. It defaults to Raspberry Pi 4 if not provided. You can find a list of [device types](/reference/hardware/devices/) here.
   - `supportedDeviceType` - The device types that the fleet supports. You can find a list of [device types](/reference/hardware/devices/) here.
+- `version` - Provides a user-defined version value that is shown in the Version column when
+  listing Releases in the balenaCloud web dashboard, and which may also be queried through
```
thgreasi, line 91: Maybe a bit simpler like v ?
```suggestion
- `version` - Provides a user-defined version value that is shown in the Version column of the Releases page of the balenaCloud web dashboard, and which may also be queried through
```

pdcastro: Done

pages/learn/deploy/deploy-with-balena-button.md
pdcastro: Thanks for the links and comments @thgreasi 👍  I have expanded the documentation with the information you have provided. Have a look.

pages/learn/deploy/deploy-with-balena-button.md
```diff
@@ -87,6 +87,19 @@ version: 3.5.2
...
+  the 3 core _major.minor.patch_ semver digits may be defined by the user (i.e. no support
+  for user-defined pre-release or metadata [semver syntax](https://semver.org/)). We plan
+  to support the full semver syntax in user-defined version values in the future. If the
+  version value follows the 3-digit format "1.2.3" and multiple deployments are made using
+  the same version value, the balena API will automatically increase the revision value in
+  the format "1.2.3+rev1", "1.2.3+rev2", "1.2.3+rev3" and so on. If the version value does
+  **not** follow the 3-digit format, currently deployment may still succeed as long as the
+  version value is unique across all releases in the fleet, however the value will not
+  surface in the web dashboard (which will then typically display version `0.0.0`) and
+  further deployments may not succeed in the future as non-conformant usage.
```
thgreasi, line 102: How about separating the paragraph with a line break for the deprecated part and simplify it a bit:
```suggestion
  the format "1.2.3+rev1", "1.2.3+rev2", "1.2.3+rev3" and so on.  
  Using version values that do **not** follow the 3-digit format is deprecated
  and may be blocked in the future. Even though currently such deployments
  may still succeed as long as the version value is unique across all
  releases in the fleet, the web dashboard will display version `0.0.0`
```

thgreasi (commented):
LGTM feel free to self-certify :+1:

pages/learn/deploy/deploy-with-balena-button.md
pdcastro: I wasn't convinced about the line break (because the description appears in indented bullet points), but I incorporated the wording simplification 👍  with minor changes.

pdcastro:
@balena-ci I self-certify!

PR #2092, opened 2021-09-30
Update compose file docs

robertgzr:
Occasionally users try to use fields that are supported by v2 but have
different properties in version 3 (like long syntax for ports) and
are surprised to get an error in response.
Clarify that our support is based on version 2.4

Change-type: patch
Signed-off-by: Robert Günzler <robertg@balena.io>

### DIFF
--- a/pages/reference/supervisor/docker-compose.md
+++ b/pages/reference/supervisor/docker-compose.md
@@ -5,13 +5,21 @@ excerpt: docker-compose.yml fields supported by {{ $names.company.lower }}
 
 # docker-compose.yml fields
 
-Our compose-file support is currently based on [version 2.2](https://docs.docker.com/compose/compose-file/compose-versioning/#version-22), as such any fields that depend on version >=2.3 are not supported.
+Our compose-file support is currently based on [version 2.4](https://docs.docker.com/compose/compose-file/compose-versioning/#version-24), as such any fields that were introduced in version 3 are not supported.
 
 ## Supported fields
 
 Field | Details
 --- | ---
-[build](https://docs.docker.com/compose/compose-file/compose-file-v2/#build) | `build` or `context` paths must point to project/app subfolders. URLs to git repositories are not currently supported.
+[build](https://docs.docker.com/compose/compose-file/compose-file-v2/#build) | when using a path it must point to project/app subfolders. URLs to git repositories are not currently supported.
+[build.context](https://docs.docker.com/compose/compose-file/compose-file-v2/#context) | must point to project/app subfolders. URLs to git repositories are not currently supported.
+[build.dockerfile](https://docs.docker.com/compose/compose-file/compose-file-v2/#dockerfile) |
+[build.args](https://docs.docker.com/compose/compose-file/compose-file-v2/#args) |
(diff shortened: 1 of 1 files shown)


klutchell:
Do we actually support v2.4 fully? Why can't I put `version: "2.4"` at the top of my compose file, is it just a CLI bug?

robertgzr:
@klutchell hmm how did you try? `balena push`?

klutchell:
@robertgzr yeah, I get `Unsupported composition version` so it might need updating in the CLI

robertgzr:
@klutchell oohh yes definitely... see the hq issue for the list of todos

pages/reference/supervisor/docker-compose.md
```diff
@@ -5,13 +5,22 @@ excerpt: docker-compose.yml fields supported by {{ $names.company.lower }}
 
 # docker-compose.yml fields
 
-Our compose-file support is currently based on [version 2.2](https://docs.docker.com/compose/compose-file/compose-versioning/#version-22), as such any fields that depend on version >=2.3 are not supported.
+Our compose-file support is currently based on [version 2.4](https://docs.docker.com/compose/compose-file/compose-versioning/#version-22), as such any fields that were introduced in version 3 are not supported.
```
klutchell, line 8:
```suggestion
Our compose-file support is currently based on [version 2.4](https://docs.docker.com/compose/compose-file/compose-versioning/#version-24), as such any fields that were introduced in version 3 are not supported.
```

klutchell:
Should we wait to merge this until all of the related 2.4 support PRs have merged in the CLI, etc?

vipulgupta2048:
@robertgzr Hey, what if we can automate this page to generate documentation as per the version we support? Can we do that if it is possible?

robertgzr:
@vipulgupta2048 I'm not sure about that. Right now the properties we do and don't support result from our code in [compose-parse](https://github.com/balena-io-modules/balena-compose-parse), with the actual implementation scattered between [multibuild](https://github.com/balena-io-modules/balena-multibuild) (for buildtime) and supervisor (for runtime)

and sometimes there is partial support (like with `depends_on` introduced here)

maybe we can express the information that is needed here in a json schema form though? we already [kind of use one](https://github.com/balena-io-modules/balena-compose-parse/blob/master/src/schemas/config_schema_v2.4.json)...

robertgzr:
another thing that came to mind is where this documentation actually appears. It's not obvious to me why it's part of the supervisor reference, since it matters at build time as well (which is more cloud/cli doc)

vipulgupta2048:
Oh wow, this got left behind quite a bit @robertgzr 
Can you add https://github.com/balena-io/docs/issues/2189 to this PR and we can finish this. 

My automation request was a more of a suggestion, although maybe the engine should have a contract of what properties it supports to help list things out in a better way.

robertgzr:
@vipulgupta2048 you can check https://github.com/balena-io/balena-io/issues/2311#issuecomment-931564390, this work is stalled until I or somebody finds the time to investigate the breakage we get from using livepush with `target` directive

we should keep this PR open until those things are all completed I think...

dfunckt:
@balena-ci rebase

dfunckt (approved):

PR #2132, opened 2021-11-23
Refactor `SUPERVISOR_POLL_INTERVAL` desc to mention back off

20k-ultra:
old description was wrong and adding confusion if you can use this variable and instant updates


20k-ultra:
@balena-ci I self-certify!

PR #2133, opened 2021-11-24
Clarify instant update usage requires it to be disabled

20k-ultra:
previous wording did not make it clear that you need to disable instant updates AND spread out the poll interval across a fleet of devices to create a deployment that prevents all the devices from downloading updates at once.


20k-ultra:
@balena-ci I self-certify!

20k-ultra:
@balena-ci retest

PR #2134, opened 2021-11-24
Update CLI docs source URL inline with landr related changes

srlowe:
Should not be merged until https://github.com/balena-io/balena-cli/pull/2388 merged.

Change-type: patch
Signed-off-by: Scott Lowe <scott@balena.io>


pdcastro (commented):
For the record, :-) I think this PR is good for self-certification as per plan in comment:
https://github.com/balena-io/balena-cli/pull/2388#issuecomment-978145126

srlowe:
@balena-ci I self-certify!

PR #2150, opened 2021-12-04
Rework reference section on how to select a base image

kb2ma:
Adds hint that use of an image with a name that starts with `%%BALENA_ARCH%%` includes fewer tools. I felt it was useful to include this in the _How to Pick a Base Image_ section. Alternatively, we could just add a note a little lower in the _Image Names_ sub-section.

### DIFF
--- a/pages/reference/base-images/base-images.md
+++ b/pages/reference/base-images/base-images.md
@@ -34,7 +34,11 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
 ## How to Pick a Base Image
 
 <!-- TODO: Add image on how image hierarchy and discuss -->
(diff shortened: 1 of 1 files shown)


pages/reference/base-images/base-images.md
```diff
@@ -34,7 +34,11 @@ excerpt: Docker images maintained by {{ $names.company.lower }}
...
+When starting out a project, it's generally easier to have a "fatter" image, which contains a lot of prebuilt dependencies and tools. These images help you get setup faster and work out the requirements for your project. For this reason, it's recommended to start with `-build` variants. Also, images whose name begins with `%%BALENA_MACHINE_NAME%%` generally include more tools, while `%%BALENA_ARCH%%` named images are kept as small as possible.
```
klutchell, line 37: Should we include some examples of the differences? I assume you referring to the extra packages installed in the device-type base images like this?
https://github.com/balena-io-library/base-images/blob/master/balena-base-images/device-base/raspberrypi3/debian/bullseye/build/Dockerfile

vipulgupta2048: How about some stats as well. Everyone likes stats of how big/small are the images.

kb2ma:
Replaced by #2160

PR #2153, opened 2021-12-10
patch: Remove TODO & docs_env

vipulgupta2048:
Not sure why we have this env file, let's see what breaks. 

---


docs_env
```diff
@@ -1,14 +0,0 @@
...
-export MAIN_SITE='https://balena.io'
-export DASHBOARD_SITE='https://dashboard.balena-cloud.com'
-export BASE_URL='https://balena.io/docs'
-export COMPANY_LOWER='balena'
-export COMPANY_ALL_CAPS='BALENA'
-export COMPANY_SHORT='balena'
-export COMPANY_DASH='balena-io'
-export OS_LOWER='balenaOS'
-export ENGINE_LOWER='balenaEngine'
-export DOMAIN='balena.io'
```
klutchell, line 14: These seem to be used by `config/names.coffee` if it were to be sourced in the environment, but that doesn't seem to be the case. Maybe it was left there for local development?

klutchell: Added via https://github.com/balena-io/docs/pull/823 but it seems it's never sourced so the defaults are used and not this file.

vipulgupta2048: Indeed, I try checking any reference for this file but it seems this files was left and deprecated. As the deployment moved to use names.coffee and links.coffee

klutchell: This file would still work though, as long as you sourced the file manually before running. I think the expected workflow was the following:
```
# manually change docs_env for testing stuff
vi docs_env

# source docs_env to set env vars that would overwrite names.coffee and links.coffee
. docs_env
npm start
```

If that workflow still works, cool. But does anyone still use it? Who is it useful for? Do we know for sure that it's not being sourced by our master deploy to prod steps?

vipulgupta2048: We deploy on Heroku using the Dockerfile and AFAIK it seems to me all variables have been migrated to `names.coffee` or `index.coffee`. You are right where this file can be still be used if we wanted to. 

How about we commit a docs_env.example file instead without the values. That way we get to keep the workflow if someone has a use for it. I don't know who uses it thought atm.

vipulgupta2048: Okay I checked the Heroku dashboard, it seems we do configure some variables and even more for production but not using this file.

klutchell: In that case I think we are good to go, no need to create a replacement workflow file. There are many ways to set env vars without having a helper file in version control.

klutchell:
@balena-ci rebase

klutchell (approved):

PR #2157, opened 2021-12-13
Network requirements: Add disclaimer about ipv6 only networks

pipex:
Change-type: patch


pipex:
IPv6 support all around is imminent, is not clear we need this change but it is a nice to have.

vipulgupta2048:
Hey @pipex any update on this.

pipex:
@vipulgupta2048 I had forgot about this. I think this PR as it contains stale information. The registry is already ipv6 afaik

PR #2224, opened 2022-03-07
Update SSH access docs ('balena ssh' usage with production OS variants)

pdcastro:
Connects-to: balena-io/balena-cli/pull/2457
Connects-to: balena-io/balena-cli/pull/2461
Change-type: patch


pdcastro:
@balena-ci I self-certify!

PR #2257, opened 2022-04-21
Patch: Correct update-tag syntax

rmorillo24:
Update-tag requires the tag filter in the URL instead of in the data

---

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -535,8 +535,8 @@
         "description": "",
         "method": "PATCH",
         "endpoint": "/v6/device_tag",
-        "filters": "?\\$filter=device/uuid%20eq%20'<UUID>'",
-        "data": "{\n    \"tag_key\": \"<EXISTING-KEY>\",\n    \"value\": \"<NEW-VALUE>\"\n}"
+        "filters": "?\\$filter=(tag_key%20eq%20'<EXISTING-KEY>')%20and%20(device%20eq%20<DEVICE_ID>)",
+        "data": "{\n    \"value\": \"<NEW-VALUE>\"\n}"
       },
       {
         "id": "delete-tag",


nelsonpina (approved):
👍🏼

config/dictionaries/resource.json
```diff
@@ -535,8 +535,8 @@
         "description": "",
         "method": "PATCH",
         "endpoint": "/v6/device_tag",
-        "filters": "?\\$filter=device/uuid%20eq%20'<UUID>'",
-        "data": "{\n    \"tag_key\": \"<EXISTING-KEY>\",\n    \"value\": \"<NEW-VALUE>\"\n}"
+        "filters": "?\\$filter=(tag_key%20eq%20'<EXISTING-KEY>')%20eq%20(device/uuid%20eq%20'<DEVICE_ID>')",
```
thgreasi, line 538: Do you mean v ?
```suggestion
        "filters": "?\\$filter=(tag_key%20eq%20'<EXISTING-KEY>')%20eq%20(device%20eq%20<DEVICE_ID>)",
```

Which made me spot the `create-tag` right above, which afaik shouldn't use quotes around the `<DEVICE_ID>` there either...

rmorillo24: In fact, it's one of these two options:

- DEVICE_ID without quotes
- DEVICE_UUID with quotes

I'll submit the change with the DEVICE_ID option as it seems to be the preferred one

vipulgupta2048 (changes requested):
According this, https://jel.ly.fish/support-thread-1-0-0-front-cnv-cehahxp it seem this change changes the value for all devices in the fleet as @nelsonpina obsvered on the support thread. We need to fix that. I see @rmorillo24 might fix this PR or create a new one but adding a `Changes requested` tag on this PR so it doesn't get merged without prior testing or changes. 

I think this could be our intended solution: https://jel.ly.fish/239228d9-367b-44e0-8504-8af3ace0c7da?event=e20b98d5-80fc-4b72-8ff9-0d4d204d3ad1

thgreasi:
@vipulgupta2048 assuming there are no unexpected character escaping issues, the current query is what the SDK does as well.
Since it filters on the tag key & device id, it should only even be able to affect a single tag.

vipulgupta2048:
@thgreasi Thanks for throwing in some light. It's becoming a little confusing for the customer it applied the tag to all devices but when @pipex tested it was fine. I am looking for a resolution. Can we use for @pipex mentioned here instead as the docs fix, since this is what actually helped the customer

https://jel.ly.fish/239228d9-367b-44e0-8504-8af3ace0c7da?event=e20b98d5-80fc-4b72-8ff9-0d4d204d3ad1

thgreasi:
@vipulgupta2048 for consistency I would offer the device id based one.
If we can now add more than one examples, then Felipe's response that uses the device uuid is a nice to have as well.

thgreasi (approved):

vipulgupta2048 (approved):

PR #2282, opened 2022-05-19
patch: Fix link reference for balena deploy

vipulgupta2048:


vipulgupta2048:
@balena-ci I self-certify!

PR #2304, opened 2022-05-30
patch: Fail build script when doxx command errors out

vipulgupta2048:
The doxx command in the build script used to error out with an exit code 0. This PR fixes that behavior with converting the unhandled-rejections in node to strict which makes sure `unhandledpromiserejection` errors always exit the program with exit code 1. Also, added `set -e` along with other bash environment variables to improve script's behavior overall. 

This primarily makes sure our CI/CD builds won't be successfull if the docs build is failing. Hopefully preventing us from merging broken documentation changes. 

Context: https://www.flowdock.com/app/rulemotion/resin-devops/threads/lRgvlRyPy6MLl4c5dsgI-vVbnEw

---

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -1,3 +1,13 @@
-dirname=$(dirname $0)
-cd "$dirname/.."
-./node_modules/.bin/doxx "$(pwd)/config/doxx"
+#!/usr/bin/env bash
+
+# exit on error, unassigned vars, or pipe fails
+set -euo pipefail
+
+# get the absolute path to the script in case it is called from elsewhere
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# run cd in a subshell so we don't end up in another directory on failure
+(
+    cd "$SCRIPT_DIR/.."
+    node --unhandled-rejections=strict ./node_modules/.bin/doxx "$(pwd)/config/doxx"
+)


tools/build.sh
```diff
@@ -1,3 +1,12 @@
-dirname=$(dirname $0)
-cd "$dirname/.."
-./node_modules/.bin/doxx "$(pwd)/config/doxx"
+#!/usr/bin/env bash
+
+# exit on error, unassigned vars, or pipe fails
+set -euo pipefail
+
+# get the absolute path to the script in case it is called from elsewhere
+dirname="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
pranavpeshwe, line 7: > dirname

_Generally_ not a good idea to have a variable named the same as a command. I know you've carried it over from earlier code. Please rename if you get a chance.

klutchell: +1 to this and I should have caught it, a good term for this path is `SCRIPT_DIR` or something like that

vipulgupta2048: Roger that.

klutchell: > @vipulgupta2048 Now that I have your attention stuck_out_tongue, please do share what the whitespaces in the command `$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` are for. Althought they shouldn't make any difference at runtime, it is not customary to have such spaces before/after commands. They made me wonder whether they are there to address some weird shell corner case (are they!?).

That was me, I recommended that change to the script to safely determine the script path and I copy-pasted the solution from SO since I had used it previously and I vaguely remembered what it looked like.

The spaces can be removed and it only adds to the readability in this case.

pranavpeshwe: Got it 👍🏻 
Thanks @klutchell .

vipulgupta2048: All suggestions added, thanks for the review. Resolving.

vipulgupta2048:
@balena-ci I self-certify!

PR #2343, opened 2022-07-15
Fix reference to missing dictionary key in Cloud IoT Provisioning

kb2ma:
Simple fix for an oversight in the final Cloud IoT Provisioning dynamic page. I also verified that there were not any other missing keys.


klutchell (approved):
Lgtm

kb2ma:
Thanks, @klutchell !

PR #2356, opened 2022-07-27
Fix location of persistent logs

### LINKED ISSUE #2355: Documented location of persistent logs is not accurate
Our docs on persistent logs say that

> Logs are located on-device at /var/log/journal/<uuid> where the UUID is device UUID.

but in fact that `<uuid>` is the one found in `/etc/machine-id`. This is a random UUID generated during the first boot, and not related to the device UUID.

lmbarros:
The docs previously said persistent logs were stored at
"`/var/log/journal/<uuid>` where the UUID is device UUID". In fact, the
UUID used in this path is not related with the device UUID: it comes
from the "machine ID" (found at `/etc/machine-id`) that is generated by
Systemd during the first boot.

Resolves #2355


klutchell (approved):
lgtm

PR #2362, opened 2022-08-02
patch: Deprecate Dynamic Docs config option

vipulgupta2048:
Had a thought about this. Our overall goal is remove as many static paritals as we could for our dynamic docs. The newest edition to Dyanmic docs is the Cloud Provisioning docs that doesn't use partials at all. Hence, I am putting together this patch to deprecate the dynamic docs proprety completely. This benefits two-folds

1. No more confusion on where the "Improve this doc" link will point to 
2. With no partials to point to, we can point to the main blueprint doc as the improve this doc link

Additionally fixes issues with "Improve this Doc" links for Cloud Provisioning docs since they don't have any partials. Their improve link was going 404

---

### DIFF
--- a/config/doxx.coffee
+++ b/config/doxx.coffee
@@ -14,9 +14,8 @@ module.exports = {
   templatesDir: config.templatesDir
   partialsDir: config.partialsDir
   metaExtra: (fileName) ->
-    fileMatch = fileName.match(config.dynamicDocs)
     externalFileMatch = fileName.match(config.externalDocs)
-    improveDocsLink: if fileMatch then "#{config.editPageLink}/#{config.partialsDir}/#{fileMatch[1]}" else "#{config.editPageLink}/#{config.docsSourceDir}/#{fileName}"
+    improveDocsLink: "#{config.editPageLink}/#{config.docsSourceDir}/#{fileName}"
     externalImproveDocsLink: if externalFileMatch then links.externalDocs[externalFileMatch[1]]
     $links: links
     $names: names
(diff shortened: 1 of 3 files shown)


vipulgupta2048:
@kb2ma Let me know what you feel about this. Since we have talked at length about dynamic documentation and their place for "Improve this Doc" links. I think this could clear up things a bit

vipulgupta2048:
@balena-ci rebase

kb2ma:
Apologies for the delay, @vipulgupta2048 . I'm working on a deadline but plan to review later this week.

kb2ma:
I think this PR is fine. The existing link to the `shared` subdirectory requires some understanding from the potential improver, and misses the root document. On the other hand, this PR only links to the root document, so an improver needs to know to look in the `shared` directory tree for the details. I don't see how to keep the interface simple (a single button) and accomplish both goals.

I would guess the Improve this Doc functionality is most useful on simpler pages without partials. So the new functionality of taking the user to the equivalent root page is at the same level of integration, and should be easy to understand. The improver will see the import directives in the markdown and understand they can't access those details from this editor.

As you say, it's a nice benefit that this change automatically fixes the Improve this Doc link for Cloud Provisioning.

vipulgupta2048:
Exactly and you understood it perfectly well. It does't solve the problem for the improver but it's way better than what we have before. Thanks for pitching in. I wanted your thoughts exactly for this reason since you know the system as a contributor.

vipulgupta2048:
@balena-ci I self-certify!

Issue #2395, opened 2022-10-12
Update recommended wireless adapters


alexgg:
The list of recommended wireless adapters in https://www.balena.io/docs/reference/hardware/wifi-dongles/ is outdated. This information should be generated from the OS documentation instead https://github.com/balena-os/meta-balena#recommended-wifi-usb-dongle  

---

## About this issue tracker

*Before creating a issue, check for duplicates. If there is another issue describing
the same bug or problem, please add comments to pre-existing issues or link them to add more context.*

While creating a new issue, always make sure to provide as much context as possible with an
expected solution or change that you would like to see in the documentation. This helps
contributors understand the issue and execute a better solution. 

If applicable, please add additional links to GitHub projects, forums.balena.io threads,
gist.github.com, Google Drive attachments, screenshots, JF/FD/Front threads for reference, etc.

*Thank you for creating an issue and helping us improve balenaCloud's documentation!*

lurch:
This probably supersedes #2193 ?

vipulgupta2048:
Not sure @lurch I am still waiting for folks concerned to update that PR. I will ping again

PR #2399, opened 2022-10-13
patch: Update Wifi USB Dongle to network adapters list

vipulgupta2048:
Solves #2395, add USB dongles and modems to the list

---


vipulgupta2048:
@alexgg @majorz This PR can be used as a reference to do future changes related to bringing information from meta-balena. 
This information will stay updated as long as the headings on meta-balena readme aren't moved or changed. This is critical for it to work. 

Please review the information already on the docs about wifi adapters and let me know if I should remove it or keep it there. It lists more wifi adapters that are not listed on the meta-balena readme.

alexgg:
@vipulgupta2048 some comments.

I think we should only list devices that we actually verify, otherwise they can silently break or become unsupported.

To this extend, the source of truth should be the lists in the meta-balena README. We also have to make sure all the devices in there are actually reflected in at least one device type in the rack.

There is a section that says:
```
Here are some notable devices known to work:

[Pi Hut USB WiFi Adapter](http://thepihut.com/products/usb-wifi-adapter-for-the-raspberry-pi) - Small form-factor and works right out of the box!
[TP-Link WN725N](https://www.amazon.com/TP-Link-wireless-network-Adapter-SoftAP/dp/B008IFXQFU) - Small, reliable and cheap TP Link device.
[TP-Link AC600](https://www.amazon.com/TP-Link-Mini-Wireless-Supports-10-9-10-14/dp/B07PB1X4CN) - Dual band alternative to WN725N.
[TP-Link Nano Router](http://www.amazon.com/TP-LINK-TL-WR702N-Wireless-Repeater-150Mpbs/dp/B007PTCFFW) - Though this isn't strictly a WiFi adapter, it does enable you to connect to WiFi network using the ethernet port of the Pi and is known to work correctly with Balena. As a result no further configuration is required.
```

That list does not come from meta-balena. We should add those devices to the rack, to the meta-balena README and then to the documentation. Otherwise move this information to a pattern instead, not to the documentation as `have been known to work` is not really useful.

Also, there is also a list of tested modem in `pages/reference/OS/network.md`. It also fetches from meta-balena but I am wondering if we should have just one place in the documentation to list these and then cross reference that, instead of sourcing twice from meta-balena.

vipulgupta2048:
You got it, I agree on everything. I will add all of them in one place and will only have verified ones in the list.

vipulgupta2048:
@alexgg Took a decision and pulled that entire section out from `network.md`, let me know how you feel about this now? Much clearer imho

alexgg:
> @alexgg Took a decision and pulled that entire section out from `network.md`, let me know how you feel about this now? Much clearer imho

I agree. I wouldn't mind a link though but I don't have a strong opinion.

vipulgupta2048:
@alexgg The link back to the page was added so all good. I will get this merged.

vipulgupta2048:
@balena-ci I self-certify!

PR #2417, opened 2022-10-25
Add cspell for spellcheck test

zoobot:
Changelog-entry: Add cspell for spellcheck test
Change-type: patch
Signed-off-by: zoobot <rommims@gmail.com>

---


package.json
```diff
@@ -9,8 +9,8 @@
   "main": "index.js",
   "scripts": {
     "start": "node index.js",
-    "test": "npm run test-spelling",
-    "test-spelling": "find . -iname \"*.*md\" -not -path \"./CHANGELOG*\" -not -path \"./node_modules/*\" | ./tools/test-spelling.sh",
+    "test": "npm run test:spelling",
+    "test:spelling": "cspell lint --no-summary --no-progress --gitignore --exclude '**/*.svg' '**/*'",
```
vipulgupta2048, line 13: I think narrowing down cspell to just docs directories might be a better choice.

cspell.json
```diff
@@ -0,0 +1,25 @@
...
+            "addWords": true
+        },
+        {
+            "name": "computerTerms",
+            "path": "./.cspell/computer-terms.txt",
+            "addWords": true
+        }
+    ],
+    "exclude": ["*/*.svg"],
+    "ignorePaths": ["node_modules","CHANGELOG.md"]
```
vipulgupta2048, line 24: I wonder then where are the names coming from if changelog is already ignored.

.cspell/computer-terms.txt
```diff
@@ -0,0 +1,354 @@
+Standardise
+xginn
```
vipulgupta2048, line 2: those are usernames and not computer terms. Can we add comments in custom dictionaries to set what each custom dictionary is for?

.cspell/computer-terms.txt
```diff
@@ -0,0 +1,354 @@
...
+popperjs
+rechoir
+bunzip
+uglifyjs
+esbuild
+watchpack
+webidl
+cloudiotcore
+gcloud
+FORMATED
```
vipulgupta2048, line 354: Wow that's a big list, where did we get this from or did you make this @zoobot? 
If it's the latter, would it be a good idea to contribute this back into Cspell?

vipulgupta2048 (commented):
Couple of questions @zoobot 
- Did you manually have to write each and every name in the custom dictionary? 
- If cspell doesn't like people's names, can we identify which file is the problem (probably changelog.md) and ignore that file from spellcheck? Better than having the names of all balenistas in a file that has to be constantly updated because the spellchecker doesn't like it. 


More comments below, good job on your continuing to push this forward

cspell.json
zoobot: Good question, some came from changelog before I ignored that. Sometimes though, there's also people's names in the files themselves or listed as authors in posts.

.cspell/computer-terms.txt
zoobot: I created it. Contributing back would be the nice thing to do. We'll still need to include the dictionary in cspell.json

cspell.json
vipulgupta2048: @zoobot Can you point to those examples?

.cspell/computer-terms.txt
zoobot: I think our linter complains about having comments in the json file. I haven't tried to push something despite the complaint. We should name the dictionaries something obvious so it doesn't require comments? Like balena-git-users or balena-users?

package.json
vipulgupta2048: Sure at the moment, we are running cspell on the entire docs using the wildcard `**/*`
Instead of doing that, I would say let's run the spellcheck on only the directories containing docs. 
Namely, `config, pages, shared, templates, README.md, CONTRIBUTING.md`

vipulgupta2048: Let me know if I forgot something

zoobot: content?

vipulgupta2048: There is a directory called content?

cspell.json
vipulgupta2048: If we don't have an example then I recommend removing the names from the dictionary, also if you can answer the questions in my primary review comment https://github.com/balena-io/docs/pull/2417#pullrequestreview-1155904613

.cspell/computer-terms.txt
vipulgupta2048: While I agree with better naming, I am not very sure about maintaining a dictionary of all balena-users. Can you point to example docs where cspell is finding usernames and names of people.

zoobot: I reran this after removing the balena users from the docs and removing changelog and it looks like it was indeed from before I removed changelog. I created a balena-user.txt dictionary but I don't think we need it after ignoring changelog. Some people have usernames that look like computerTerms or balena dictionary terms so let me know if you see people in there who need to be removed.

package.json
zoobot: hmm don't see one on master

vipulgupta2048:
Hey @zoobot Checked out your PR locally and found many issues. So I thought I commit a fix to them myself. Here are the changes:
1. Lot of words were repeated in both lower and upper case, some were spelled incorrectly in the custom dictionary. some weren't needed.
2. Streamlined the balena dictionary by quite a lot 
3. Sorted the dictionary in ascending order 
4. Added couple of more dictionaries to cspell and remove misc. dictionary
5. Removed the flags for no progress as that hid the report which is something we would like to see in CI/CD
6. Critically, the wildcards you set were only checking 4 files. I have fixed them to check all 189 files. Do give them a look and see if we can improve even more. 
7. Fixed couple of more spelling errors and removed wrong words. 

We actually need to remove the flag that ignores files mentioned in gitignore but its too much work right now to fix the hundred or so spelling errors in the masterclass + with me moving the content of masterclasses. We can safely ignore it. This is good to go now, thanks for your efforts on this.

vipulgupta2048 (commented):
Approved, you can go ahead and self-certify when you checked the changes.

.cspell/balena-words.txt
```diff
@@ -0,0 +1,27 @@
+balena
+balena-api
+balena-base-ui
+balena-build
+balena-builder
+balenahup
+balenalib
+balenistas
+bindmount
+boostrap
```
lurch, line 10: Is this really a "balena word", or is it a typo for `bootstrap` ?

.cspell/balena-words.txt
```diff
@@ -0,0 +1,27 @@
...
+capitano
+contracts
+docsearch
+doxx
+Fincm
+generic
```
lurch, line 16: I wouldn't have thought `generic` or `model` are balena-specific words? :shrug:

.cspell/balena-words.txt
zoobot: @lurch This got merged as-is so if you see something that needs to change, feel free to change it in another PR!! Making the dictionaries perfect will be a group effort. It looks like possibly boostrap is misspelled somewhere and we should just fix the spelling instead of including it in the dictionary at all.

.cspell/balena-words.txt
vipulgupta2048: Cspell detects spellings of the word as wrong but having a subset of the word can be helpful for multiple permutations. So here the error case was probably `genericx86-64-ext`

.cspell/balena-words.txt
lurch: > Making the dictionaries perfect will be a group effort.

I'm not gonna submit PRs, I don't work for Balena. I'm just trying to be helpful by pointing out things that my eye-for-detail spotted :eyes: :slightly_smiling_face:

.cspell/balena-words.txt
lurch: Ahhh. I've not used `cspell` myself, but I have used `codespell` elsewhere, which works in a slightly different way - instead of insisting that _every_ word appears in a dictionary, codespell instead looks for "commonly misspelled words" and suggests a suitable replacement. So please forgive my misunderstanding.

.cspell/balena-words.txt
zoobot: HA! I like your eye for detail!! 🙂 I'll do a hunt for that word and remove it from the dictionary. Thanks!!

.cspell/balena-words.txt
lurch: (but even then, we only used codespell as a "source of suggestions", and didn't insist that it passed 100% with no errors)

zoobot: Just getting started with cspell on here, will see how it goes. Who knows how many people will misspell thingrs on purposes. 

Anything in the dictionaries gets a pass. The rest gets handled by cspell and fails tests if it's misspelled.

PR #2421, opened 2022-10-31
patch: Add docs for download endpoint

vipulgupta2048:
Added 2 examples explaining how balenaOS can be downloaded using the /download endpoint for both public and private device types. This is how it looks. 

Screenshot from 2022-10-31 15-28-58

---

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -555,21 +555,21 @@
     "examples": [
       {
         "id": "download-private-balenaos",
-        "summary": "Download unmanaged releases of balenaOS for private device types",
-        "description": "Unmanaged releases of balenaOS for private device types are not released publically. If you have access to a private device type, then images can be downloaded by providing an API key. Add an `--output` flag to provide a path for the image to be downloaded in the curl command below.",
+        "summary": "Download private or ESR balenaOS images",
+        "description": "In order to download an ESR or private balenaOS image, you need to provide your [bearer token](https://www.balena.io/docs/learn/manage/account/#api-keys). The `deviceType` and `version` parameters are required. Add an `--output` flag to provide a path for the image to be downloaded.",
         "method": "GET",
         "endpoint": "/download",
-        "content-type": "application/zip",
+        "content-type": "application/octet-stream",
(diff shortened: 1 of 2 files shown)


vipulgupta2048:
We have no field for this I assume @thgreasi So I assume this is good to go as is. Definitely long due.

config/dictionaries/resource.json
```diff
@@ -548,6 +548,31 @@
...
   },
+  {
+    "id": "download",
+    "name": "Download balenaOS",
+    "fields": [],
+    "examples": [
+      {
+        "id": "download-private-balenaos",
+        "summary": "Download unmanaged releases of balenaOS for private device types",
+        "description": "Unmanaged releases of balenaOS for private device types are not released publically. If you have access to a private device type, then images can be downloaded by providing an API key. Add an `--output` flag to provide a path for the image to be downloaded in the curl command below.",
```
thgreasi, line 559: > Unmanaged releases of balenaOS for private device types are not released publically.

That's a bit confusing. Yes they aren't publically but you could provide a bearer header and then it should work. So I would rephrase this like "in order to access private or ESR images you need to provide your bearer token".

thgreasi (commented):
I'm a bit against documenting this tbh and offering only the UI & the CLI as the ways to download images :thinking: 
If you still prefer to add this, then let's explicitly mention that both the device type and the version parameters are required (b/c we do want to do a major about that soon).

thgreasi:
> We have no field for this I assume

What kind of field @vipulgupta2048 ? The response is a stream so I guess there are no fields since it's not a json...

config/dictionaries/resource.json
vipulgupta2048: > I'm a bit against documenting this tbh and offering only the UI & the CLI as the ways to download images

I know, I know but we do provide a way and it's not like it's going anywhere. IMHO I feel it's better we document it. 
I will mention the requirement for sure, and will also mention that it's a stream. 
I like your version better, yeah. Coming up.

vipulgupta2048:
I was referring to fields like other resources that we have such as fleet, or device. I knew the download endpoint wouldn't have any but still wanted to confirm.

vipulgupta2048:
@balena-ci I self-certify!

PR #2459, opened 2022-12-13
Convert docs to full static & deploy on Cloudflare

vipulgupta2048:
In order for the migration from Heroku --> Cloudflare to be complete, the site needs to become fully static. Added multiple minor changes to make the migration smoother. The major issue I feel for this migration is regrettably the poor support Cloudflare Pages has for redirects.

Docs: https://developers.cloudflare.com/pages/platform/redirects/
Support ticket on CF: https://support.cloudflare.com/hc/en-us/requests/2644158
Zulip: Long discussion @ https://balena.zulipchat.com/#narrow/stream/350505-aspect.2Fcommunication/topic/Redirects.20in.20Cloudflare.20Pages

After spending a week on and off working on it. I don't want to block the migration PR for too long, I have at least added the best changes I have for the moment and with everything I have tried out. Everything else is and should be working quite well. Check out the preview URL for the site preview: https://4e77d057.balenacloud-docs.pages.dev/

---


.github/workflows/flowzone.yml
```diff
@@ -10,5 +10,8 @@ on:
 jobs:
   flowzone:
     name: Flowzone
-    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
-    secrets: inherit
\ No newline at end of file
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@vipul/custom-build
```
klutchell: This branch has been merged now

.github/workflows/flowzone.yml
```diff
@@ -6,12 +6,21 @@ on:
...
+  # allow external contributions to use secrets within trusted code
+  pull_request_target:
+    types: [opened, synchronize, closed]
+    branches: [main, master]
 
 jobs:
   flowzone:
     name: Flowzone
-    uses: product-os/flowzone/.github/workflows/flowzone.yml@vipul/custom-build
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
```
klutchell, line 17: Could you rebase the 2nd commit to include the additional changes to this file?

.github/workflows/link-checker.yml
```diff
@@ -4,9 +4,6 @@ on:
   # Check links once a week on Monday
   schedule:
     - cron: "0 0 * * 1"
-  pull_request:
-    branches:
-      - "master"
```
klutchell, line 9: We could move this to a flowzone custom test action, but why disable it now?

package.json
```diff
@@ -9,6 +9,7 @@
   "main": "index.js",
   "scripts": {
     "start": "node index.js",
+    "start:cloudflare": "npm run build:fast && npm install wrangler && npx wrangler pages dev build/",
```
klutchell, line 12: Who should be using this script vs `deploy-docs` used by Flowzone? Is this one for local development or manual deployment or something?

.github/workflows/link-checker.yml
vipulgupta2048: Not being disabled. I just felt it was not helping a lot when it was running as part of a PR. Let me do that in another PR

package.json
vipulgupta2048: Indeed, this command is for local development 
For deployment we are using deploy-docs.

.github/workflows/flowzone.yml
vipulgupta2048: Have never tried to do it but on the interactive rebase I used the e option to edit the commit and it end up swallowing the commit entirely. I only have 5 commits locally for some reason.

vipulgupta2048: Will pull and try again.

vipulgupta2048: Edited the 2nd commit as mentioned.

.github/workflows/link-checker.yml
klutchell: Yeah, I don't think we need to disable that as part of this PR.

.github/workflows/flowzone.yml
klutchell: You sure? The second commit still shows the old values. I would prefer this commit not contain any changes to the workflow and there be a separate commit for the Flowzone changes only.

.github/workflows/link-checker.yml
```diff
@@ -4,6 +4,9 @@ on:
   # Check links once a week on Monday
   schedule:
     - cron: "0 0 * * 1"
+  pull_request:
+    branches:
+      - "master"
```
klutchell, line 9: You now have a commit where this is removed, and another commit where you add it back. I think a bunch of these commits need to be reworked and squashed, or have changes moved between them.

klutchell: You can probably get away with one commit for the links and reference changes, and another commit with just the build script and workflow changes (if that makes things easier).

klutchell (changes requested):

.github/workflows/link-checker.yml
vipulgupta2048: @klutchell Already had restored the changes to this file as per my last comment. Forgot to resolve this. Please resolve this comment when you see it.

.github/workflows/flowzone.yml
vipulgupta2048: The second commit has changes that make sense for the commit. It's commit for Deploying to CF. I added configuration changes needed to deploy to CF? Not sure why there should be a separate commit for Flowzone changes. Here's what it shows me.

Screenshot from 2022-12-29 17-10-41

.github/workflows/link-checker.yml
vipulgupta2048: Forget about this, check the commits. Tell me in a new review how I did. I have checked all commits I think they should all make sense. 

TIL `git reset @~`

.github/workflows/flowzone.yml
```diff
@@ -11,4 +11,7 @@ jobs:
   flowzone:
     name: Flowzone
     uses: product-os/flowzone/.github/workflows/flowzone.yml@master
-    secrets: inherit
\ No newline at end of file
+    secrets: inherit
+    with:
+      cloudflare_website: "balenaCloud-docs"
+      custom_website_build: "npm run build"
```
klutchell: My concern with this commit is that the option `custom_website_build` was never merged in Flowzone master, and this line is changed in a future commit. So this commit without the future commit is invalid and difficult to follow the blame.

klutchell (changes requested):

.github/workflows/flowzone.yml
vipulgupta2048: I agree on that. Let me go back and restore this change as well. My other commit includes the correct change.

vipulgupta2048: Resolved, how about now. Fixed that change in the second commit.

vipulgupta2048: also wrote a blog post: https://mixstersite.wordpress.com/2023/01/02/edit-git-commits/

tools/deploy-static.sh
```diff
@@ -1,7 +1,18 @@
...
+
+# Extract only the Cloudflare Redirects from redirect.txt
+DIVISION_LINE=`awk '/####*/ {print NR}' config/redirects.txt | sed -n 3p`
+head -n $DIVISION_LINE config/redirects.txt > build/_redirects
+
+# Command to create redirects that don't have trailing slash since Cloudflare has limited regex support
+# awk '/^\/[[:alnum:]]+/ {print $1}' build/_redirects | sed -e 's!.$!!'
+
+# Copy the 404 page to root of the build folder for CF pages to find it
+cp build/404/index.html build/404.html
```
klutchell, line 20: Suggest we re-write this file to verify that the redirects exists with both / and no slash. Assuming we can figure out why CF is treating the redirects as dynamic that is.

```bash
#!/usr/bin/env bash

# Script to prepare Docs for Cloudflare Pages deployment

set -e

dirname=$(dirname $0)
cd "${dirname}"/.. || exit 1

# Build folder is uploaded to Cloudflare
cp -r static/* build/

# normalize all Cloudflare Style Redirects to include both trailing slash and no trailing slash variants
generated="$(sed -r 's/^(\/[^ ]+\/[^\/ ]+) /\1\/ /g' config/redirects.txt | sed -rn 'p; s/^(\/[^ ]+)\/ /\1 /p' | sort)"
current="$(sort config/redirects.txt)"

if [ "${current}" != "${generated}" ]
then
    echo "Error: one or more redirects is missing the trailing-slash or non-trailing slash variant, both must exist"
    diff  <(echo "${current}" ) <(echo "${generated}")
    exit 1
fi

# copy the Cloudflare Style Redirects
sed -rn '/^(\/[^ ]+\/[^\/ ]+)/p' config/redirects.txt build/_redirects

# Copy the 404 page to root of the build folder for CF pages to find it
cp build/404/index.html build/404.html
```
The assumption here is that we commit the redirects.txt with both variations so there are no surprises on deploy.

vipulgupta2048: Right on, I agreed with your suggestion and actually added the changes as well locally along with couple of more commits but haven't pushed. I was hoping for Cloudflare to respond sooner but they are taking ample time with it. Fingers crossed for us to get unblocked.

vipulgupta2048: Update: Let's do the validation in a separate PR. Pushing without the checks.

Dockerfile
```diff
@@ -1,14 +1,17 @@
...
-EXPOSE 3000
+# COPY . .
+# RUN npm run build
 
-ENTRYPOINT ["/usr/local/bin/npm"]
-CMD ["start"]
+# EXPOSE 3000
+
+# ENTRYPOINT ["/usr/local/bin/npm"]
+# CMD ["start"]
```
klutchell, line 17: I would rather delete this file, or uncomment the lines. It serves no purpose in this state.

> Kept around if someone comes looking to deploy Docs using a docker image

In that case the lines should be uncommented.

vipulgupta2048:
@balena-ci I self-certify!

klutchell (approved):

PR #2483, opened 2023-01-10
deploy-static: Exit with error if any redirects are missing

klutchell:
Cloudflare Pages treats redirects with and without trailing slashes separately as per their documentation. So for every redirect we need to check that we have included both variations.

Change-type: patch

---

### DIFF
--- a/config/redirects.txt
+++ b/config/redirects.txt
@@ -2,30 +2,56 @@
 # Cloudflare Style Redirects
 ###############################################
 
+# When adding new redirects in here, make sure you include both
+# one with a trailing slash, and one without, as Cloudflare treats
+# these separately.
+# If you miss adding one of the variants the deploy-static step
+# should fail with an error to avoid publishing to production.
+
 # Legacy URLs
 /pages/ /
+/pages /
 /introduction/introduction/ /learn/welcome/introduction/
+/introduction/introduction /learn/welcome/introduction/
 / /learn/welcome/introduction/ 301
 /learn/ /learn/welcome/introduction/ 301
+/learn /learn/welcome/introduction/ 301
 /installing/gettingStarted/ /learn/getting-started/raspberrypi3/nodejs/
+/installing/gettingStarted /learn/getting-started/raspberrypi3/nodejs/
 /installing/gettingStarted-BBB/ /learn/getting-started/beaglebone-black/nodejs/
+/installing/gettingStarted-BBB /learn/getting-started/beaglebone-black/nodejs/
 /installing/gettingStarted-Humming/ /learn/getting-started/hummingboard/nodejs/
+/installing/gettingStarted-Humming /learn/getting-started/hummingboard/nodejs/
 /installing/gettingStarted-NUC/ /learn/getting-started/intel-nuc/nodejs/
+/installing/gettingStarted-NUC /learn/getting-started/intel-nuc/nodejs/
 /installing/gettingStarted-Nitrogen6x/ /learn/getting-started/nitrogen6x/nodejs/
+/installing/gettingStarted-Nitrogen6x /learn/getting-started/nitrogen6x/nodejs/
 /installing/gettingStarted-Edison/ /learn/getting-started/intel-edison/nodejs/
+/installing/gettingStarted-Edison /learn/getting-started/intel-edison/nodejs/
 /installing/gettingStarted-Odroid-C1/ /learn/getting-started/odroid-c1/nodejs/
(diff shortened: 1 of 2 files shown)


config/redirects.txt
```diff
@@ -4,128 +4,242 @@
 
 # Legacy URLs
 /pages/ /
+/pages /
```
vipulgupta2048, line 7: IMHO I don't think we should commit these in in. It makes a pattern where folks unassuming will add redirects with or without slash then duplicate them without knowing why. On having these done build time we can create a predictable file that we know is taking care of all the cases. It should also be clear here why this is even happening because this is not an ideal behavior by Cloudflare Pages. This should have been handled quite easily if Cloudflare provided support for regex which they told me could be coming soon™️

tools/deploy-static.sh
```diff
@@ -1,23 +1,37 @@
 #!/usr/bin/env bash
-# Script to prepare Docs for Cloudflare Pages deployment 
+
+# Script to prepare Docs for Cloudflare Pages deployment
 
 set -euo pipefail
 
-dirname=$(dirname $0)
+# resolve absolute path and account for spaces, etc
+dirname="$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)"
```
vipulgupta2048, line 8: We can actually change this command in all the scripts, since many of them use this.

tools/deploy-static.sh
```diff
@@ -1,23 +1,37 @@
...
-head -n $DIVISION_LINE config/redirects.txt > build/_redirects
+# normalize all Cloudflare Style Redirects to include both trailing slash and no trailing slash variants
+generated="$(sed -rn 'p; s/^(\/([^ ]+\/)?[^\/ ]+) \//\1\/ \//p' config/redirects.txt |
+    sed -rn 'p; s/^(\/([^ ]+\/)?[^\/ ]+)\/ \//\1 \//p' |
+    awk '!/^(\/([^ ]+\/)?[^\/ ]+)/ || !seen[$0]++' | sort)"
+current="$(sort config/redirects.txt)"
+
+if [ "${current}" != "${generated}" ]
+then
+    echo "Error: one or more entries config/redirects.txt is missing the trailing-slash or non-trailing slash variant, both must exist"
```
vipulgupta2048, line 22: This is making things harder than they should be I feel. I think we should have this validation occurring for the file that is being created on build time which is build/_redirects (output)

Which is actually the file being sent to Cloudflare.

config/redirects.txt
klutchell: As discussed previously, I don't like the idea of redirects being generated behind the scenes that we may not understand, especially if the regex is wrong or breaks. I would much rather have the redirects be in source control and put the burden on the developer to add the required variants.

I want to avoid a case where redirects start being weird in production and no one knows why because a regular expression is doing the heavy lifting.

tools/deploy-static.sh
klutchell: True, but not something I wanted to pull into this PR. If they aren't broke...

config/redirects.txt
klutchell: Another way to look at it is, assume the regex is bad, or goes bad, I would rather block a developer trying to merge a PR, then break something in production.

vipulgupta2048: Okay that last part I agree, but then I recommend we add some instruction into how exactly folks will need to add a new redirect. Should they add only the trailing slash redirect and generate the one without trailing slash?

klutchell: I don't think developers should have to generate anything (ideally). If adding a redirect they should add both variations or they will get met with an error. I think re-generating the file with the provided commands should be a last resort.

vipulgupta2048: > Should we document the requirement at the top of the redirects file?

Yeah, that's my point. I think a couple of sentences might help in the redirects file and error.

vipulgupta2048: You can make these changes and should be good to go. Do merge it so we can see the changes on Monday.

tools/deploy-static.sh
klutchell: Made the change to validate the build output rather than the input. The impact should be the same.

config/redirects.txt
klutchell: Added a couple sentences to the top of the redirects file. Let me know if it makes sense to you.

vipulgupta2048 (approved):

PR #2694, opened 2023-05-04
Add `device history` docs

fisehara:
Change-type: minor


vipulgupta2048 (commented):
How's this going @fisehara

fisehara:
@vipulgupta2048 As the feature currently undertakes a little restructuring of the data model, the device history itself won't be surfaced in the UI but a real device audit log. May be similar but not same. Will keep it updated.

PR #2722, opened 2023-05-29
Add the `overall_status` field for `device`

lmbarros:
As [discussed internally](https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/device.2Eoverall_status/near/362024508), the `overall_status` field was missing for the `device` resource.

First time touching the API resources part of the docs. [Looks correct](https://38e5a981.balenacloud-docs.pages.dev/reference/api/resources/device/), but let me know if I did something wrong here!

### DIFF
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -228,6 +228,7 @@
       "note",
       "local_id",
       "status",
+      "overall_status",
       "is_online",
       "last_connectivity_event",
       "is_connected_to_vpn",
--- a/config/dictionaries/resource.json
+++ b/config/dictionaries/resource.json
@@ -300,6 +300,14 @@
         "filters": "?\\$filter=uuid%20eq%20'<UUID>'"
       },
       {
+        "id": "device-overall-status",
+        "summary": "Get the device overall_status field",
+        "description": "The overall_status field is returned only when explicitly requested with $select.",
+        "method": "GET",
+        "endpoint": "/v6/device(<ID>)",
+        "filters": "?\\$select=overall_status"
+      },
+      {
         "id": "rename-device",
         "summary": "Rename device",
         "description": "",


config/dictionaries/resource.json
```diff
@@ -228,6 +228,7 @@
       "note",
       "local_id",
       "status",
+      "overall_status",
```
cywang117, line 231: Is there a way to specify that this field only appears if explicitly `$select`ed, perhaps in the examples?

vipulgupta2048: That could be a good one, I have been confused by the CLI's output of device status where it show is_online === true for even devices that are offline. @lmbarros Let's push to get this merged and add an example if you can.

lmbarros: @cywang117 , @vipulgupta2048 I added a commit with the example. We could maybe have a more inspiring `description` field, perhaps giving some info about it is and how `overall_status` differs from `status`. I think I don't grok this to the point of being able to come with a correct and concise description, though. Feel free to suggest!

vipulgupta2048: How about we point back to the docs to explain overall_status? 
I don't know what `status` is about @lmbarros

lmbarros: I don't think `overall_status` is documented, @vipulgupta2048 :-/ My understanding is that it is on the same spirit as the status on the dashboard (Thodoris described it as "the device state from the observer's/user's point of view"), but it is not the same logic as the one used on the dashboard, with different (and undocumented) possible values, IIUC.

lmbarros: [A bit more context](https://balena.zulipchat.com/#narrow/stream/345882-_help/topic/Computed.20connectivity.20state.20SDK.20method) on a thread from less than an hour ago, by coincidence.

vipulgupta2048 (approved):

PR #2723, opened 2023-05-31
develop/hardware: Update I2C example in the i2c-and-spi section

acostach:
The existing sample application is now old, let's use the balena-sense project as an example for I2C.

Change-type: patch

---

### DIFF
--- a/pages/learn/develop/hardware/i2c-and-spi.md
+++ b/pages/learn/develop/hardware/i2c-and-spi.md
@@ -30,26 +30,28 @@ Many sensors and peripherals use either the [I²C (Inter-Integrated Circuit)][i2
 
 ### I2C
 
-To enable I2C communication in your projects you will need to add the command `modprobe i2c-dev` to your package.json or Dockerfile.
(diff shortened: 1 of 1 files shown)


pages/learn/develop/hardware/i2c-and-spi.md
```diff
@@ -30,26 +30,28 @@ Many sensors and peripherals use either the [I²C (Inter-Integrated Circuit)][i2
...
+
+```
+services:
+  myservice:
+  ...
+    privileged:true
+```
 
-__Note:__ A few places will talk about adding the modules to the /etc/modules file so that they are there on boot. This will not work on the {{ $names.company.lower }} system because that file is not mapped to the host OS.
+Single-container fleets run in privileged modeby default. If security and sandboxing are required then a multicontainer release should be created, even if only a single service is required.
```
lurch, line 52: "modeby" -> "mode by" ?

acostach: Fixed, thanks for spotting that typo

lmbarros (changes requested):
This looks great to me Alexandru! The CI checks are complaining that `myservice` is an unknown word :roll_eyes: . I added comments with suggestions that will hopefully make this check pass.

lmbarros:
LGTM

PR #2786, opened 2023-08-27
Use fleet slug instead of fleet name for examples

otaviojacobi:
Change-type: patch

---


vipulgupta2048 (changes requested):
Thanks for suggesting this @otaviojacobi Would recommend that we replace `myuser` as the org name with slugs like `org` or `myorg` to make it explicitly clear what that the value here means the organization slug. It doesn't mean `username` here but on the first read, I understood it like that.

PR #2811, opened 2023-09-28
Update commands in offline updates

pipex:
The `balena fleet` commands in this guide no longer work with the latest version of the CLI which returns a capitalized table instead of all uppercase items. Not sure when this change took place on the CLI side. 

Change-type: patch

### DIFF
--- a/pages/learn/deploy/offline-updates.md
+++ b/pages/learn/deploy/offline-updates.md
@@ -93,7 +93,7 @@ $ balena fleet create ${fleet_name} --type ${device_type}
 Initialize the `fleet_slug` environment variable with the command below to store the slug of the fleet.
 
 ```bash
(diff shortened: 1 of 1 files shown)


volks73:
@pipex Hello! I think the `awk` command also needs to be changed from `$2` to `$3`. The `$2` value returns the `:` not the values.

As an example,

```sh
balena fleet ${fleet_name} | grep Slug | awk '{print $2}'
:
```

versus

```sh
❯ balena fleet ${fleet_name} | grep Slug | awk '{print $3}'
<org>/<fleet>
```

The `<org>` and `<fleet>` have been omitted.

pipex:
@volks73 nice catch. I've added the changes

vipulgupta2048 (approved):
Thanks @pipex @volks73 for bringing this log. Approved 🚀

Issue #2909, opened 2024-02-19
Add SRI to Docs JS resources being fetched from CDN's


vipulgupta2048:
Refer to https://github.com/balena-io/docs/security/code-scanning/1 for more information

PR #2910, opened 2024-02-19
patch: Add Subresource Integrity to scripts being fetched

vipulgupta2048:
Adding Subresource integrity to script tags in the docs. 

Hash generated from: https://www.srihash.org/
Report: https://github.com/balena-io/docs/security/code-scanning/1
Solves: https://github.com/balena-io/docs/issues/2909

### DIFF
--- a/templates/_devices.html
+++ b/templates/_devices.html
@@ -1,4 +1,4 @@
-<script src="//cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.15/lodash.min.js" type="text/javascript"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.15/lodash.min.js" integrity="sha512-3oappXMVVac3Ge3OndW0WqpGTWx9jjRJA8SXin8RxmPfc8rg87b31FGy14WHG/ZMRISo9pBjezW9y00RYAEONA==" crossorigin="anonymous"></script>
 <script type="text/javascript">
 
   (function () {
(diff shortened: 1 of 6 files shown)


templates/_devices.html
```diff
@@ -1,4 +1,4 @@
-<script src="//cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.15/lodash.min.js" type="text/javascript"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.15/lodash.min.js" integrity="sha512-3oappXMVVac3Ge3OndW0WqpGTWx9jjRJA8SXin8RxmPfc8rg87b31FGy14WHG/ZMRISo9pBjezW9y00RYAEONA==" crossorigin="anonymous"></script>
```
thgreasi, line 1: The docs say that you could just use `crossorigin` as a tag w/o a value, and that should be equivalent to `anonymous`
> Setting the attribute name to an empty value, like crossorigin or crossorigin="", is the same as anonymous.

But other than that lgtm :+1: 

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin#sect1

thgreasi (approved):

templates/_devices.html
vipulgupta2048: TIL.
A field without a value is just confusion and an extra google search waiting to happen. Leaving it explicit makes the code self-explanatory.

PR #2919, opened 2024-02-19
Update config.txt to reflect changes in config.txt parsing with Supervisor v16+

cywang117:
Supervisor v16+ adds each dtparam for a dtoverlay on its own line to comply with the 80-char line limit for config.txt.

Change-type: patch

### DIFF
--- a/pages/reference/OS/advanced.md
+++ b/pages/reference/OS/advanced.md
@@ -95,6 +95,16 @@ dtoverlay=i2c-rtc,ds1307
 dtoverlay=lirc-rpi
 ```
 
+As of Supervisor v16+, the above will translate to:
+
+```
+dtoverlay=i2c-rtc
+dtparam=dc1307
+dtoverlay=lirc-rpi
+```
+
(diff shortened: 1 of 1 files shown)


cywang117:
I self-certify!

thgreasi (approved):

pages/reference/OS/advanced.md
```diff
@@ -95,6 +95,16 @@ dtoverlay=i2c-rtc,ds1307
...
 
+As of Supervisor v16+, the above will translate to:
+
+```
+dtoverlay=i2c-rtc
+dtparam=dc1307
+dtoverlay=lirc-rpi
+```
+
+This modifies each parameter to be on its own line in order to avoid the 80 character line limit imposed by config.txt. It is the recommended method of setting one or more overlays with their own parameters.
```
lurch, line 106: Only a minor point, but https://www.raspberrypi.com/documentation/computers/config_txt.html#dtparam says that the limit is actually 98 characters.

PR #2962, opened 2024-04-23
patch: Add node v20 as supported version

Iznogohul:
Add node v20 as supported version and change node versions to reflect what nodejs contract currently builds

### DIFF
--- a/pages/reference/base-images/base-images.md
+++ b/pages/reference/base-images/base-images.md
@@ -22,7 +22,7 @@ excerpt: Docker images maintained by balena
   - [Ubuntu](https://www.ubuntu.com/): xenial (16.04), bionic (18.04), focal (20.04), impish (21.10), kinetic (22.10) and jammy (22.04)
   - [Fedora](https://getfedora.org/): 36, 37 and 38
 - Multiple language stacks:
-  - [Node.js](https://nodejs.org/en/): 14.21.2, 16.19.0, 18.14.0 and 19.6.0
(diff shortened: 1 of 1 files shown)


vipulgupta2048 (commented):
Thanks for making a PR to update the documentation. There is an ongoing effort to update node version for our base images. I am checking if it has been completed and we are clear to merge this PR

thgreasi (approved):

Iznogohul:
@vipulgupta2048 @thgreasi Just noticed that node 14 and 16 has been removed from the node contract but they exist in the balenalib hub.I think it would be helpful to mark them as deprecated/EOL in the docs and reflect that these versions of sw.stack will not get rebuild for example with newer sw.os releases.

thgreasi:
That indeed sounds correct @Iznogohul .
I recently actually removed a number of EOL `sw`/`os`es from the contracts, which means that base images will no longer be built, but back then thought that those changes would propagate to the docs automatically
https://github.com/balena-io/contracts/pull/421
https://github.com/balena-io/contracts/pull/422
https://github.com/balena-io/contracts/pull/423

We will probably try to make that automation work, but I wouldn't block any PR doing those changes manually in the mean time.

vipulgupta2048:
Indeed docs, at the moment don't generate this page automatically and we are figuring out an strategy to automate all balena-base images docs to reflect the right information out there.

Iznogohul:
@thgreasi Will try to open these PRs if I find some spare time.

thgreasi:
Thank you @Iznogohul

PR #3053, opened 2024-08-08
Generate base images stack versions

vipulgupta2048:
This change generates versions for base images stacks

- **patch: Generate base images stack versions automatically**
- **Add base images generation script to build process**

After:

Screenshot 2024-08-08 at 12 55 18 PM

Fibery: https://balena.fibery.io/Work/Project/Automate-base-images-public-docs-202

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules/
 server/nav.json
 npm-debug.log
 static/dist
+contracts/
 
 .#*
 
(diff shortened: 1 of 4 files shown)


vipulgupta2048:
@klutchell Thanks for reviewing Kyle, very much indeed. This is one of two PRs. Talking with Thodoris, I realized this could be a viable separation of work where this part of the documentation can be generated right here by adding three or four seconds extra to the build just because of cloning the contracts. 

If you agree, please approve the PR to get this going and yes, I am making another PR for the base-images list table page that would originate from the base-images folder.

klutchell (approved):

PR #3062, opened 2024-08-22
Reduce repetitive paths for FAQ section

vipulgupta2048:
URLs that have repetitive paths or subfolders within the string (example.com/page1/page/page1/page2). In some cases this can be legitimate and logical, however it also often points to poor URL structure and potential improvements. It can also help identify issues with incorrect relative linking, causing infinite URLs.

Part of SEO sprint: https://balena.fibery.io/Work/Task/URL-Repetitive-Path-2079

- **patch: Move pages from /faq/troubleshooting to /faq/**
- **Add redirects & navigation fix**


shaunmulligan (approved):
All looks great!

PR #3081, opened 2024-10-08
API/overview: Replace the application.device_type property with is_for__device_type since it's not in the v6 model

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Task/Update-the-doc-examples-to-use-the-v7-model-2208

### DIFF
--- a/pages/reference/api/overview.md
+++ b/pages/reference/api/overview.md
@@ -49,11 +49,10 @@ Depending on the number of fleets you have access to, this could return much mor
 
 `$select` specifies which fields to return for each resource. By default, every field comes back as part of the response, but most use cases require only one or two of these pieces of information.
 
-The following API call uses `$select` to only return the name and device type for each application:
+The following API call uses `$select` to only return the name and the device type id for each application:
 
 ```shell
-curl -X GET
-"{{ $links.apiBase }}/v6/application?\$select=app_name,device_type" \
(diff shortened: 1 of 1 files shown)


pages/reference/api/overview.md
```diff
@@ -103,6 +102,15 @@ curl -X GET \
 -H "Authorization: Bearer <AUTH_TOKEN>"
 ```
 
+Similarly we can extend our earlier API call that retrieves all applications to also include their device type slug by using a `$expand`:
```
thgreasi, line 105: Since it's clear from the above (v5 based) example that we intended to show how to retrieve an application w/ its 
device type slug, but we never updated that for v6, I've now added that as example with $expand.

pages/reference/api/overview.md
```diff
@@ -49,11 +49,10 @@ Depending on the number of fleets you have access to, this could return much mor
 
 `$select` specifies which fields to return for each resource. By default, every field comes back as part of the response, but most use cases require only one or two of these pieces of information.
 
-The following API call uses `$select` to only return the name and device type for each application:
+The following API call uses `$select` to only return the name and the device type id for each application:
 
 ```shell
-curl -X GET
-"{{ $links.apiBase }}/v6/application?\$select=app_name,device_type" \
```
thgreasi, line 56: There is no `device_type` property in v6!!!

joshbwlng (approved):

PR #3103, opened 2024-11-11
Device Statuses: Clarify the meaning of update status duration

kb2ma:
Change-type: patch

### DIFF
--- a/pages/learn/manage/device-statuses.md
+++ b/pages/learn/manage/device-statuses.md
@@ -53,7 +53,7 @@ __Note:__ Data for these states will only get populated for devices with [superv
 
 ## Update Status Duration
 
-The `Update Status Duration` field is calculated based on the timestamp of the most recent update event. For example in the screenshot above for Device Connectivity states,  device *bitter-wind* shows an update status duration of 17 minutes. This value means that the update status changed to `Downloaded` 17 minutes ago. When querying this data from the API, you will receive a timestamp instead of a duration.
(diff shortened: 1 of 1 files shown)


pages/learn/manage/device-statuses.md
```diff
@@ -53,7 +53,7 @@ __Note:__ Data for these states will only get populated for devices with [superv
 
 ## Update Status Duration
 
-The `Update Status Duration` field is calculated based on the timestamp of the most recent update event. When querying this data from the API, you will receive a timestamp instead of a duration.
+The `Update Status Duration` field is calculated based on the timestamp of the most recent update event. For example in the screenshot above for Device Connectivity states,  device *bitter-wind* shows an update status duration of 17 minutes. This value means that the update status changed to `Downloaded` 17 minutes ago. When querying this data from the API, you will receive a timestamp instead of a duration.
```
vipulgupta2048, line 56:
```suggestion
This field is calculated based on the timestamp of the most recent update event. For example: In the screenshot above for Device Connectivity states, the device `bitter-wind` shows an update status duration of 17 minutes. This signifies that the device update status changed to `Downloaded` 17 minutes ago, which was the most recent update event. This value resets again when the update status changes again for the device. When querying this data from the API, you will receive a timestamp instead of a duration.
```

I feel a line was missing. Added that.

kb2ma: Thanks @vipulgupta2048 . I found having "again" twice in the new sentence confusing, so I reworded it for clarity.

vipulgupta2048 (approved):

PR #3142, opened 2025-02-27
patch: Update query to fetch device types

vipulgupta2048:
Updated the query to fetch the right device types from the API


the-real-kenna:
LGTM

PR #3220, opened 2025-09-08
patch: Account for ratelimits on releases API

vipulgupta2048:
✅ With GitHub Token: Gets real version data (3 versions each)
✅ Without GitHub Token: Creates valid fallback (1 "latest" version each)
✅ Build Always Succeeds: No more "Unknown key $balenacli" errors

### DIFF
--- a/tools/versioning.js
+++ b/tools/versioning.js
@@ -38,6 +38,22 @@ function githubRequestOptions(endpoint) {
 }
 
 /**
+ * Creates a safe fallback version when GitHub API fails
+ * @returns {Array} Always returns a valid "latest" version entry
+ */
+function createSafeFallback() {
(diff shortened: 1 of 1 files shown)


tools/versioning.js
```diff
@@ -112,58 +133,146 @@ function findComplyingTags(tagsWithDates) {
...
-      req.on('error', (error) => {
-        reject(error);
-      });
+  try {
+    const releases = await fetchRecentReleases(owner, repo);
+    
+    // If we got a fallback response, return it directly
+    if (releases.length > 0 && releases[0].id === "latest") {
+      return releases;
+    }
```
myarmolinsky, line 148: Why have a fallback at all instead of just returning an empty array?

PR #3348, opened 2026-03-24
Add `generate-node-sdk-pages` script for making polished Node SDK pages

myarmolinsky:
Fibery: https://balena.fibery.io/Work/Project/Identify-the-next-steps-for-improving-growing-our-docs-2317
Improvement: https://balena.fibery.io/Work/Improvement/SDKs-(and-probably-other-generated-external-doc-pages)-Internal-linking-is-not-working-3947
Change-type: patch

---


tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+					// Stitch it back together using the captured hashes, the isolated last part, and the new bold string
+					return `${hashes} ${lastPart}\n**${newBoldContent}**\n`;
+				},
+			);
+
+			// Some auth model methods have a md links hardcoded as `#balena.models.auth.X`
+			// This should correct them to `#X`
+			sectionContent = sectionContent.replace(
+				/\[([^\]]+)\]\(#balena\.(?:.*\.)?([^.)]+)\)/g,
+				'[$1](#$2)',
```
klutchell, line 229: The content splitting relies on hardcoded string markers (`\n## Modules`, `\n## balena-sdk\n`, `<a name="balena"></a>`, `<a name="balena.errors"></a>`). If the upstream SDK docs change the order of these sections, rename them, or alter the anchor formatting, the script will silently produce garbage output or crash.

Consider adding guards after each split to make failures explicit, e.g.:

```js
if (firstSplitParts.length < 2) {
    console.error(`Expected "## Modules" marker not found in ${file}`);
    process.exit(1);
}
```

Same for `secondSplitParts`, `thirdSplitParts`, and `fourthSplitParts`.

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+const path = require('path');
+
+const nodeSDKDocsDir = path.join(
+	__dirname,
+	'../pages/external-docs/sdk/node-sdk',
+);
+const inputPath = path.resolve(nodeSDKDocsDir);
+const outputDir = path.join(__dirname, '../pages/reference/sdk/node-sdk');
+const outputPath = path.resolve(outputDir);
+const semver = require('semver'); // If you have semver installed, otherwise use a regex split
```
klutchell, line 11: Beyond removing that, `require('semver')` should be moved to the top of the file with the other requires rather than being buried after variable declarations.

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+	// Top-level (Version folders) sorting
+	if (level === 0) {
+		items.sort(sortVersions);
+	} else {
+		// Sub-folder/File sorting: Alphabetical
+		items.sort((a, b) => a.name.localeCompare(b.name));
+	}
+
+	for (const item of items) {
+		if (item.name === 'README.md') {
```
klutchell, line 47: At `level === 0`, `items.sort(sortVersions)` runs over all directory entries. If a non-version entry (stray `.DS_Store`, etc.) appears at this level, `semver.rcompare` will throw and fall through to the `localeCompare` fallback. Harmless today, but worth either filtering to directories-only before sorting, or adding a comment noting the assumption.

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+	// If topLevelFound is 5 (#####), shiftAmount is 3.
+	// If topLevelFound is 2 (##), shiftAmount is 0.
+	const shiftAmount = topLevelFound - 2;
+
+	let lines = content.split('\n');
+
+	return lines
+		.map((line) => {
+			const headingMatch = line.match(/^(#{2,6})\s(.*)/);
+			if (headingMatch) {
```
klutchell, line 102: `flattenHeadings` converts anything deeper than `###` into bold text. The comment says this is to stay within GitBook's hierarchy, but it means any `####` (or deeper) headings in the SDK docs -- commonly used for method parameters and return values -- lose their semantic heading status and become unlinkable bold paragraphs.

Is that an acceptable tradeoff? If GitBook truly only supports 3 levels, it's fine, but worth confirming and documenting here.

package.json
```diff
@@ -9,12 +9,13 @@
   "main": "index.js",
   "scripts": {
     "test": "npm run sync-external -- --dry-run && npm run renovate:validate",
-    "sync-external": "node tools/sync-external.js",
+    "sync-external": "node tools/sync-external.js && npm run generate-node-sdk-pages",
```
klutchell, line 12: Chaining `generate-node-sdk-pages` into `sync-external` unconditionally means every `sync-external` invocation regenerates all SDK pages even if nothing in the SDK source docs changed.

Worth considering either making it a separate CI step that only triggers if the target is the Node SDK.

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+		} else {
+			// The regex captures the 3 specific pieces of data we need to build the new string
+			const anchorHeadingsPattern =
+				/(?:\*\s\*\s\*\s*)?<a name="([^"]+)"><\/a>\s*(#+)\s*([^\n\r]+)/g;
+
+			// The replacement string references those captured groups using $1, $2, and $3
+			sectionContent = sectionContent.replace(
+				anchorHeadingsPattern,
+				'$2 $1\n**$3**',
+			);
```
klutchell, line 198: The content splitting here relies on hardcoded string markers (`\n## Modules`, `\n## balena-sdk\n`, `<a name="balena"></a>`, `<a name="balena.errors"></a>`). If the upstream SDK docs ever change the order, rename a section, or alter the anchor format, this will silently produce bad output or crash.

Would be good to add guards after each split, something like:

```js
if (firstSplitParts.length < 2) {
    console.error(`Expected "## Modules" marker not found in ${file}`);
    process.exit(1);
}
```

At least then failures are obvious instead of generating broken pages.

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+	if (startIndex !== -1) {
+		// Find where the Node SDK section ends.
+		// We look for the next line that is a top-level bullet (indented by 0 or 2 spaces)
+		// but is NOT a child of Node SDK (like Python SDK).
+		let endIndex = summaryLines.findIndex(
+			(line, idx) =>
+				idx > startIndex &&
+				line.trim().startsWith('* [') &&
+				!line.startsWith('    '), // Not deeply indented
+		);
```
klutchell, line 321: The `endIndex` logic for finding where the Node SDK section ends is fragile. It looks for the next line starting with `* [` that isn't indented by 4+ spaces, but if someone changes the indentation style in `SUMMARY.md` or adds a non-standard line, this could eat content from adjacent sections (like Python SDK).

Consider matching more specifically, e.g. looking for the next sibling-level entry at the exact same indent depth as the Node SDK line rather than relying on "not deeply indented."

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+			createFiles(
+				formatSectionContent(
+					sectionContent.split('\n').slice(1).join('\n').trim(),
+				),
+				modelsPath,
+			); // Recursive call to split models into subfiles
+		} else {
+			// The regex captures the 3 specific pieces of data we need to build the new string
+			const anchorHeadingsPattern =
+				/(?:\*\s\*\s\*\s*)?<a name="([^"]+)"><\/a>\s*(#+)\s*([^\n\r]+)/g;
```
klutchell, line 192: This `anchorHeadingsPattern` regex and the subsequent multi-capture-group `sectionContent.replace` calls are doing non-trivial transformations. Some inline comments showing the before/after for each transform would help future maintainers understand what these are doing without having to mentally execute the regex. For example:

```js
// Before: <a name="balena.auth.authenticate"></a>
// ##### balena.auth.authenticate(credentials) ⇒ <code>Promise</code>
//
// After: ##### balena.auth.authenticate
// **balena.auth.authenticate(credentials) ⇒ <code>Promise</code>**
```

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+
+			fs.writeFileSync(path.join(outputDir, `${fileName}.md`), fileBody);
+		}
+	}
+}
+
+function splitDocs(inputDir, outputDir) {
+	// Create base output directory if it doesn't exist
+	if (fs.existsSync(outputDir)) {
+		fs.rmSync(outputDir, { recursive: true, force: true });
```
klutchell, line 246: The script `rm -rf`s the output directory and regenerates from scratch every time. The third commit commits the generated output, but there's nothing preventing someone from editing these files by hand and having their changes silently blown away on the next run.

Two suggestions:
1. Add a header comment to each generated file, e.g. `<!-- This file is auto-generated by tools/generate-node-sdk-pages.js. Do not edit manually. -->`
2. Consider adding a note in a README or `.gitattributes` marking these as generated.

tools/generate-node-sdk-pages.js
```diff
@@ -0,0 +1,347 @@
...
+			// Recurse into subdirectories
+			lines.push(...generateSummaryLines(fullPath, relativePath, level + 1));
+		} else if (item.name.endsWith('.md')) {
+			lines.push(`${indent}* [${displayName}](${relativePath})`);
+		}
+	}
+	return lines;
+}
+
+function flattenHeadings(content) {
```
klutchell, line 82: `flattenHeadings` converts anything deeper than `###` into bold text to fit GitBook's hierarchy constraints. Just want to confirm that's acceptable for the SDK docs, since any `####` headings (e.g. for method parameters or return values) will lose their semantic heading status and won't be linkable.

myarmolinsky:
Closing in favor of: https://balena.fibery.io/Work/Project/Docs-Fix-SDK-linking-not-working-2488

PR #3434, opened 2026-04-29
Update link-checker workflow

klutchell:
- Removed 'deploy-docs' step
- Add missing input from workflow dispatch trigger
- Removed 'setup-node' step
- Fixed base path to ./pages
- Removed conditions from cache-save step
- Update lychee exclude paths
- Add pull request target

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -119,10 +119,12 @@ exclude = [
     "https://github.com/balena-io/docs/issues/new?*",
     "https://api/",
     "https://raw.githubusercontent.com/balena-os/meta-balena/$%7Bos_version_tag%7D/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor.inc",
+    ".gitbook/assets/",
+    "^[^http]"
     ]
 
 # Exclude these filesystem paths from getting checked.
-exclude_path = ["./build/img/", "./build/css/", "./build/favicon.png", "./build/dist/", "./node_modules"]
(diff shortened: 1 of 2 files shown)


lychee.toml
```diff
@@ -119,10 +119,11 @@ exclude = [
     "https://github.com/balena-io/docs/issues/new?*",
     "https://api/",
     "https://raw.githubusercontent.com/balena-os/meta-balena/$%7Bos_version_tag%7D/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor.inc",
+    ".gitbook/assets/"
     ]
 
 # Exclude these filesystem paths from getting checked.
-exclude_path = ["./build/img/", "./build/css/", "./build/favicon.png", "./build/dist/", "./node_modules"]
+exclude_path = ["./pages/.gitbook/"]
```
myarmolinsky: Why do we exclude these files? If there are broken links among the reusable content, we should be aware of those as well

klutchell: That was the first attempt to hide all the 404s returned by assets, but it wasn't actually required so I removed it now. The `".gitbook/assets/"` exclusion is the important one.

klutchell: Though, it still only scanned 8213 total, so either there are no links in there or it's still not being included.
https://github.com/balena-io/docs/actions/runs/25134213522/job/73668177742

myarmolinsky (approved):

PR #3618, opened 2026-07-31
Replace `Get all devices` with `Get all devices by organization`

Page-:
Getting all devices the user has access to via any path is both costly and liable to cause problems later if the user becomes part of more organizations or otherwise gains access to devices that they do not expect to be included

Change-type: patch

---


shaunmulligan (approved):
Looks good to me 👍

